PR #87, opened 2013-08-01
Added an option to create pretty output via the JSONWriter class

JoeArmstrong:
Added a new constructor, public JSONWriter(Writer w, boolean pretty), which enables the writer to add appropriate whitespace to make the output more human-reader friendly.

### DIFF
--- a/JSONWriter.java
+++ b/JSONWriter.java
@@ -64,6 +64,24 @@ public class JSONWriter {
      * value.
      */
     private boolean comma;
+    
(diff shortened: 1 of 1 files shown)


stleary:
Thanks for the pull request. In addition to Phillipp's comments, the comment block for JSONWriter includes this text: "No whitespace is added, so the results are ready for transmission or storage".  I think this change would go against the intent of the class design.

PR #92, opened 2014-02-07
JSONObject.numberToString() now outputs large Double integers as integer...

LeviticusMB:
...s.

For instance, 99999999 is now serialized as "99999999" instead of "9.9999999E7"

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -771,18 +771,12 @@ public class JSONObject {
         testValidity(number);
 
 // Shave off trailing zeros and decimal point, if possible.
-
-        String string = number.toString();
-        if (string.indexOf('.') > 0 && string.indexOf('e') < 0
-                && string.indexOf('E') < 0) {
-            while (string.endsWith("0")) {
-                string = string.substring(0, string.length() - 1);
-            }
-            if (string.endsWith(".")) {
-                string = string.substring(0, string.length() - 1);
-            }
+        if (number.doubleValue() == number.longValue()) {
+            return "" + number.longValue();
         }
-        return string;
+        else {
+            return number.toString();
+        }
     }
 
     /**
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -772,7 +772,7 @@ public class JSONObject {
 
 // Shave off trailing zeros and decimal point, if possible.
         if (number.doubleValue() == number.longValue()) {
-            return "" + number.longValue();
+            return Long.toString(number.longValue());
         }
         else {
             return number.toString();


stleary:
Hi, can you clarify what this change fixes? Using Java v1.8, when I checked the example mentioned in your comment (99999999), I get the same results for JSONObject.numberToString(), the proposed change, and number.toString().  Thanks, Sean

johnjaylward:
From what I could tell it was simply replacing string index searches with object to primitive conversions for some unknown amount of performance improvement.

LeviticusMB:
```java
class numberToString {
        Double number = Double.parseDouble(args[0]);

        System.out.println("Old: " + numberToString(number));
        System.out.println("New: " + numberToString2(number));
    }

    public static String numberToString(Number number) {
        String string = number.toString();
        if (string.indexOf('.') > 0 && string.indexOf('e') < 0
            && string.indexOf('E') < 0) {
            while (string.endsWith("0")) {
                string = string.substring(0, string.length() - 1);
            }
            if (string.endsWith(".")) {
                string = string.substring(0, string.length() - 1);
            }
        }
        return string;
    }
    
    public static String numberToString2(Number number) {
        if (number.doubleValue() == number.longValue()) {
            return Long.toString(number.longValue());
        }
        else {
            return number.toString();
        }
    }
}
```

```sh
iris:~ leviticus$ java -version
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8
java version "1.8.0_20"
Java(TM) SE Runtime Environment (build 1.8.0_20-b26)
Java HotSpot(TM) 64-Bit Server VM (build 25.20-b23, mixed mode)
iris:~ leviticus$ java numberToString 99999999
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8
Old: 9.9999999E7
New: 99999999
iris:~ leviticus$ 
```

stleary:
Thanks for the clarification, seeing the same results now. I guess the impact is restricted to JSONArray and JSONObject instances that are initialized from Objects. Still not clear on whether this is to improve performance (see comment by johnjaylward), a bug fix, or for some other purpose.

johnjaylward:
Personally, I'm not sure this pull request is a great solution. There is nothing inherently wrong with the E notation, and using Number.longValue() and Number.doubleValue() can lead to imprecise results.

From the javadoc: https://docs.oracle.com/javase/8/docs/api/java/lang/Number.html

public abstract long longValue()
Returns the value of the specified number as a long, which may involve rounding or truncation.

public abstract double doubleValue()
Returns the value of the specified number as a double, which may involve rounding.

Note the "rounding" on the doubleValue method. In the case of some BigDecimal like 9.0000000000000000000000000000000000001, you will get an output of 9 instead of the correct value.


```
import java.math.BigDecimal;

class numberToString {
        BigDecimal number = new BigDecimal(args[0]);

        System.out.println("Old: " + numberToString(number));
        System.out.println("New: " + numberToString2(number));
    }

    public static String numberToString(Number number) {
        String string = number.toString();
        if (string.indexOf('.') > 0 && string.indexOf('e') < 0
            && string.indexOf('E') < 0) {
            while (string.endsWith("0")) {
                string = string.substring(0, string.length() - 1);
            }
            if (string.endsWith(".")) {
                string = string.substring(0, string.length() - 1);
            }
        }
        return string;
    }

    public static String numberToString2(Number number) {
        if (number.doubleValue() == number.longValue()) {
            return Long.toString(number.longValue());
        }
        else {
            return number.toString();
        }
    }
}
```

```
$ java numberToString 9.0000000000000000000000000000000000001
Old: 9.0000000000000000000000000000000000001
New: 9
```

LeviticusMB:
@stleary: The fix was because [Zendesk](https://www.zendesk.com/) send out a letter to all customers stating that they would start rejecting all API calls that didn't write "integer fields" as plain integers, without E-notation or decimals. Now *I* know there is no such thing as an "integer" in JSON, and I tried to argue so, but in the end, we were depending on those API calls to work ... That's the motivation for this patch.

@johnjaylward: JSON-java only supports "Boolean, Double, Integer, JSONArray, JSONObject, Long, String, or the JSONObject.NULL object" so that is not a problem that can happen.

The patch is safe and produces nicer JSON, but we are already embedding the JSON code in our projects so it doesn't really matter if you merge it or not. Do what you think is right; I'm just sharing it.

stleary:
* Agree that the BigDecimal example does not appear to be a problem since it will be added as string (in the example, I think it becomes "9" when added to JSONArray or JSONObject). 
* The proposed change does not fix a bug or improve performance. It is an enhancement for a particular customer.
Thanks for providing the additional information and discussion. I don't think we want to merge this change, but will leave the pull request open for a few days in case anyone else wants to comment.

stleary:
Thanks for the pull request. At this time changes that are not bug fixes or performance enhancements aren't being merged.

PR #94, opened 2014-04-01
Update Cookie.java

hokiemike84:
Initialized capacity of a StringBuffer to avoid the need for re-sizing.  The returned string should be of equal or lesser capacity than the original string.


stleary:
Hi Mike,
Your recommended change was incorporated during the Java 1.8 date. In the case of Cookie.java, we now have:
        StringBuilder   sb = new StringBuilder(length);
Thanks for the suggestion, 
Sean

PR #103, opened 2014-09-03
Added a util method to wrap dates with colons with double quotes.

agur:


amirgur1:
Yes, sure it is not valid.  
The issue is that in practical real life the library users will face those and need ways to handle it properly and easily.  So maybe it's better to handle it centrally and have the whole community come with the best ways to address it - as the alternative is thjat each one will create his own way and re-invent the wheel.

PR #110, opened 2014-12-12
Removed compilation warnings

xhaleera:
I have updated the code to removed several compilation warnings.


stleary:
Hi Christophe, 
Sorry for not responding sooner. 
- One of your changed files, Zipper.java, is no longer part of the project and should not be included in this pull request. 
- In JSONObject.java and Property.java, I am not sure that the annotation @SuppressWarnings("unchecked") is advisable to add to the code. Can you explain why you think this is the best solution? 
Thanks, Sean

xhaleera:
Thanks for your answer @stleary.
I've updated my branch to reflex recent deletions in the repository.

Concerning the @SuppressWarnings part, I've added it to remove the warnings I get when Object are casted to something else, resulting in unchecked cast.

Eventually, I can provide another commit ignoring only the specific cast lines.

stleary:
I think SupressWarnings and unchecked casts are equally annoying. I note that the problem lines are all casts to generics, e.g. 
    return new JSONObject((Map<String, Object>) value).toString();
What would you think about something like, 
    return new JSONObject((Map<?, ?>) value).toString();
[Note: I am not going to be able to commit any changes until the unit test suite is completed, which will take a little while. But it would be good to get rid of the warnings in a safe way. Let me know what you think.]

xhaleera:
Thank you for the tip. I've pushed a new commit using <?> and no warning any more.
I think we're good now.

stleary:
Thanks! As mentioned earlier,  I need to hold off on commits until the unit tests are done. If you wish to contribute, the repository is: https://github.com/stleary/JSON-Java-unit-test.

stleary:
Thanks for the pull request. I think removing warnings is advisable if there is 100% certainty that the code is not affected. My bad for suggesting Map<?,?> and Collection<?>. That invokes the JSONObject(Object) and JSONArray(Object) ctors, and incidentally breaks the unit tests. To summarize:
* Properties.java change should be OK. properties.propertyNames() returns an Enumeration<? extends Object>. 
* JSONObject casts to Class<Object> should be OK. object.getClass() returns a Class<? extends Object>.
* All other casts need to have @SuppressWarnings restored, but not at function level. Instead, please try something like this to limit the scope to just the line where the cast occurs:
```
        } else if (value instanceof Map) {
            @SuppressWarnings("unchecked")
            Map<String, Object> map = (Map<String, Object>) value;
            new JSONObject(map).write(writer, indentFactor, indent);
        }
```

xhaleera:
I've restored "local" @SuppressWarnings for the specific lines. Now we should be OK.

stleary:
Looks good. Leaving it up for a day or so in case of additional comments. Will have to hand merge, looks like there is a formatting issue in JSONObject.

xhaleera:
The formatting issue should be fixed now, allowing you to automerge.

stleary:
Thanks, merged on a separate branch.

PR #111, opened 2015-01-05
Wildcard generics in Collection and Map arguments

domusofsail:
There is a problem with some generics in some function/constructor arguments: for example the constructor
~~~
public JSONArray(Collection<Object> collection)
~~~
won't be catched by:
~~~
List<String> myList = new ArrayList<String>();
JSONArray myJsonArray = new JSONArray(myList);
~~~
but it catches this constructor `public JSONArray(Object array)` and so it throws this exception:
~~~
 org.json.JSONException: JSONArray initial value should be a string or collection or array.
	at org.json.JSONArray.<init>(JSONArray.java:177)
~~~
The reason is that `Collection<Object>` doesn't match all kinds of collections.

The correct way to reference collections of all type of objects would be  `Collection<? extends Object>` or more concisely `Collection<?>`

Hope to see this patch applied because I have a lot of code that gets the above mentioned exception since constructors changed from `public JSONArray(Collection collection)` to `public JSONArray(Collection<Object> collection)`

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -150,10 +150,10 @@ public class JSONArray {
      * @param collection
      *            A Collection.
      */
-    public JSONArray(Collection<Object> collection) {
+    public JSONArray(Collection<?> collection) {
         this.myArrayList = new ArrayList<Object>();
         if (collection != null) {
-            Iterator<Object> iter = collection.iterator();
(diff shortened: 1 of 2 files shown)


domusofsail:
Closed because of duplicate #112

PR #114, opened 2015-04-01
Replace Collection<Object> to Collection<?> (wildcard).

TimReset:
Otherwise List<?> will call  JSONArray(Object array) but is not right.

Example:
List<?> list = new ArraysList<Object>();
JSONArray array = new JSONArray(list);

With old code this call JSONArray(Object array) and then throw exception ( throw new JSONException("JSONArray initial value should be a string or collection or array.");) But is not right! JSONArray(Collection<Object> collection) will be call! With wildcard that be work.

P.S. Sorry for my English.

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -150,10 +150,10 @@ public class JSONArray {
      * @param collection
      *            A Collection.
      */
-    public JSONArray(Collection<Object> collection) {
+    public JSONArray(Collection<?> collection) {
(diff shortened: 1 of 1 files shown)


stleary:
Evaluating, will get back to you soon. No worries on the English. Thanks, Sean

stleary:
Please see pull requests 111 and 112, which proposed a similar implementation.

PR #115, opened 2015-04-12
Faster and simpler JSONArray constructor and trimming of numeric strings.

sbandara:
1. I moved trimming of floating point numeric strings into its own function <code>trimNumericString</code>, as it is used in at least two places in <code>JSONObject</code>. The trimming procedure is simplified and avoids creation of multiple intermediate string objects.

2. The constructor of <code>JSONArray(Object)</code> used to employ reflection for initialization from a Java array. Here, I simply changed the method signature of the constructor to take an argument <code>Object[] array</code> instead of <code>Object array</code>. This then allows us to access and wrap the array elements directly. This does change the org.json library interface because all valid calls to this constructor use an argument that is already an <code>Object[]</code>.

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -27,7 +27,6 @@ package org.json;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
-import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Iterator;
@@ -166,16 +165,10 @@ public class JSONArray {
      * @throws JSONException
      *             If not an array.
      */
-    public JSONArray(Object array) throws JSONException {
+    public JSONArray(Object[] array) throws JSONException {
         this();
-        if (array.getClass().isArray()) {
-            int length = Array.getLength(array);
-            for (int i = 0; i < length; i += 1) {
-                this.put(JSONObject.wrap(Array.get(array, i)));
-            }
-        } else {
-            throw new JSONException(
-                    "JSONArray initial value should be a string or collection or array.");
+        for (int i = 0; i < array.length; i ++) {
+            put(JSONObject.wrap(array[i]));
         }
     }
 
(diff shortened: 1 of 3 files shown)


sbandara:
Correcting typo: the changes do **not** change the library interface.

stleary:
Hi Samuel, Thanks for the pull request, looking into the changes and will get back to you soon. FYI, the JSON-Java unit tests are not finished, but the work in progress can be found here: https://github.com/stleary/JSON-Java-unit-test.  Feel free to test your code against it if you are so inclined. Are you sure the library API is not changed? There seem to be 2 places where this is occurring. Thanks, Sean

sbandara:
Hi Sean, sounds great! I saw your test repository and will use it for future validation. And yes, there is a publicly visible change of the <code>JSONArray</code> constructor argument from <code>Object</code> to <code>Object[]</code> but calling this constructor with anything else than an array as argument was never allowed and used to cause a <code>JSONException</code> to be thrown. In other words, the change as a side-effect now also enforces the correct argument type at runtime. Thanks for getting back to me so quickly! Let me know if you see any other concerns.

sbandara:
... at compile-time, I mean.

stleary:
I think this code works with the current API but would cause a syntax error with the proposed change to JSONArray(Object[] object):
        int[] myInts = { 1, 2, 3, 4, 5, 6, 7 };
        Object myObject = myInts;
        JSONArray jsonArray = new JSONArray(myObject);
See http://stackoverflow.com/questions/219881/java-array-reflection-isarray-vs-instanceof for additional discussion about this.

Unless throwing a JSONException can be characterized as a bug, I don't think the exception thrown from numberToString() should be changed; existing code may be catching just that particular exception - e.g. it fails one of the unit test cases.

Not sure yet about  trimNumericString(String num). Refactoring in general is good, but the new code is not exactly the same as the deleted code.

sbandara:
You are right! I had forgotten about arrays of primitives. The switch from <code>JSONException</code> to <code>NullPointerException</code> was a minor change that isn't important to me either. Let's close this pull request, then.

I'd submit a few more focused commits, including the isolated addition of <code>trimNumericString</code> again. I was thinking some cleanup here and there could be beneficial, but are you saying we should not rewrite any existing code? Let me know how conservative we should be.

douglascrockford:
If the the code is buggy or unmaintainable, then it must be repaired. This code has been around for 15 years, so we must be cautious about making unnecessary changes. We want no breakage.

If we are making changes for performance reasons, then I want to see clear evidence of a significant performance improvement. A change that is motivated by performance that does not substantially improve performance is a bug.

sbandara:
Many thanks for clarifying, Doug. I re-submitted the isolated change to the trimming routine which stood out to me mostly as code-duplication.

PR #116, opened 2015-04-15
Extract trimming of numeric strings.

sbandara:
String representations of floating point numbers are trimmed of trailing zeros in two places using duplicated code. I refactored those uses into calls to a new static method <code>trimNumericString</code>. The body of that function looks different than the extracted code, but behaves identically for all input and operates on a single <code>String</code> object before invoking <code>substring</code> once.

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -27,7 +27,6 @@ package org.json;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
-import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


douglascrockford:
What is the evidence that it is faster? How much faster?

sbandara:
A third less object allocations when writing arrays of <code>double</code>. Profiling the encoding of an array of 1024 elements on the Oracle VM for JRE 8

    public static void main(String args[]) {
    	JSONArray array = new JSONArray();
    	for (double d = 0; d < 1024; d ++) {
    		array.put(d);
    	}
    	String str_array = array.toString();
    	System.out.println(str_array);
    }

with JProfiler shows me savings in <code>char[]</code> allocations of **25 kb** when tested with JProfiler. Reducing object churn on a loaded server or Android phone should be a good thing to do.

douglascrockford:
The justification for this change is performance. How much time is being saved? Is it a sufficient improvement to justify the change?

sbandara:
Memory allocations in a GC'ed environment affect performance in terms of speed, but it depends on the memory state of the JVM, i.e, the context in which your library is used and how. The above example will not show you any improvements in speed because there won't be any garbage collection going on. However, a server under heavy load that streams JSON arrays may benefit, and also tightly memory-constrained environments like smartphones. Anyways, thanks for your prompt responses! If you are not convinced, just close this...

PR #118, opened 2015-04-19
fix toString(JSONArray) to emit object.toString() values

stleary:
Please review and post any questions or concerns.
Might have misplaced the branch notes, which are included here:

JSONML converts XML to JSONObject or JSONArray objects. The corresponding toString() methods should generate similar XML.

**Does the change conflict with the JSON spec?**
No

**Any changes to the API?**
No

**Any changes to how the code behaves?**
No changes to exception behavior. JSONML.toString(JSONArray) will now include object.toString(), in addition to string, JSONObject, and JSONArray content output. E.G. boolean output. This will change the output for some XML inputs.

Yes, but see [https://github.com/stleary/JSON-Java-unit-test/tree/JSONMLTest-for-toString-issue], which will be committed if/when this change is merged.

No

### DIFF
--- a/JSONML.java
+++ b/JSONML.java
@@ -373,6 +373,8 @@ public class JSONML {
                         sb.append(toString((JSONObject)object));
                     } else if (object instanceof JSONArray) {
                         sb.append(toString((JSONArray)object));
+                    } else {
+                        sb.append(object.toString());
                     }
                 }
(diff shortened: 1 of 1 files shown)


stleary:
No one has posted comments or concerns, so this change will be merged.

PR #121, opened 2015-05-04
added chain of key access in all get methods

raseshshah:
Before Commit:
chain of method call to access literal or object
json.getJSONArray("person").getJSONObject(0).getJSONObject("name").getString("fname");

After Commit:
Now able to access literal or object in one call itself 
json.getString("person[0].name.fname");


johnjaylward:
I personally wouldn't want this in the core lib. It's asking for more problems than solving, and is the reason libraries like QueryDSL were written for SQL statements.

douglascrockford:
No thank you.

raseshshah:
ok.

PR #122, opened 2015-05-06
Merge branch for issue #110

stleary:
Replaced tab chars, updated versions


stleary:
These are the changes from pull request #110

**Has any code changed?**
Yes

**Verified via unit test?**
Yes. I had to add 2 unit tests to make sure each changed line was executed at least once. Will commit the updated unit tests after this commit.

PR #128, opened 2015-05-22
Handle precise and large numbers

FritzMock:
Use BigDecimal and BigInteger for numbers that will either be to precise for Double or to large for Long.
Since the conversion of Numbers changed, also some _invalid_ notations of JSON-numbers that got parsed before will no longer be supported. @see differences between Double.valueOf(String s) and new BigDecimal(String s)

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -30,6 +30,8 @@ import java.io.Writer;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.util.Collection;
 import java.util.Enumeration;
 import java.util.HashMap;
@@ -114,7 +116,7 @@ public class JSONObject {
 
         /**
          * A Null object is equal to the null value and to itself.
-         *
+         * 
          * @param object
          *            An object to test for nullness.
          * @return true if the object parameter is the JSONObject.NULL object or
@@ -122,7 +124,7 @@ public class JSONObject {
          */
         @Override
         public boolean equals(Object object) {
-            return object == null || object == this;
+            return object == null || object == this; 
         }
 
         /**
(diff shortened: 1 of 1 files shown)


stleary:
This looks like an enhancement, not a bug fix. No changes to the API, but see below. The internal behavior has changed. Some data inputs may no longer be allowed, which will probably be unacceptable and should be avoided, invalid or not. 

The RFC does not seem to define numbers sharply, and allows implementations to restrict the range of numbers processed. Interoperability is mentioned as a concern, since users may assume that all implementations support the same ranges as what they use locally.  JSON-Java advertises its numeric restrictions in the API contract: only double, long, int, and in places Number is supported. 

It would probably be advisable to add new API methods for BigDecimal and BigInteger in JSONObject, JSONArray, JSONWriter, etc. Other internal areas of the code may have to be touched as well.

It seems reasonable to enhance JSON-Java to take a larger numeric range. Merging should not be performed until all code changes are completed and concerns addressed. Will this work for you?

FritzMock:
A new API for BigDecimal/BigInteger will be impossible without changing parsing and internal representation of numbers and that will ultimately change behaviour.
To preserve current behaviour we would end up in two different "modes of operation", switching code/types at various places. Or we duplicate lots of code in parallel classes. Neither is an approach I generally do not like much, introduces unneeded complexity and causes more work than actually needed with a clean design (i.e. BigDecimal/BigInteger as only internal representation.)
But since JSON-java lacks efficient versioning, there is no choice but seamless integration or a complete fork ...
I will give it some try, with adding new API methods and switching modes of operation.

stleary:
We should shift discussion of support for BigDecimal and BigInteger to https://github.com/douglascrockford/JSON-java/issues/126. If I missed something important to this issue, please bring it up there. I don't think other changes to the numeric behavior of JSON-Java will be able to get traction, given the need to maintain backwards compatibility. Thanks for bringing up the issue, hopefully it will lead to improved big number support.

PR #135, opened 2015-06-21
Proposed BigInteger, BigDecimal support

### LINKED ISSUE #126: Investigate feasibility of enhancing JSON-Java to support BigDecimal and BigInteger

stleary:
This commit resolves #126

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -28,6 +28,7 @@ import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
 import java.lang.reflect.Array;
+import java.math.*;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


stleary:
Glad I waited, JSONArray has some missed get[type] methods, probably should be updated too.

PR #136, opened 2015-07-03
add toArrayList method

Spedwards:
Provided a method to return the JSONArray Object as an ArrayList<Object>.

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -878,6 +878,15 @@ public class JSONArray implements Iterable<Object> {
         }
         return jo;
     }
+    
+    /**
+     * Returns the JSONArray as an ArrayList.
+     * 
+     * @return a copy of the JSONArray as an ArrayList.
+     */
+    public ArrayList<Object> toArrayList() {
+        return this.myArrayList;
+    }
 
     /**
      * Make a JSON text of this JSONArray. For compactness, no unnecessary
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -885,7 +885,9 @@ public class JSONArray implements Iterable<Object> {
      * @return a copy of the JSONArray as an ArrayList.
      */
     public ArrayList<Object> toArrayList() {
-        return new ArrayList<Object>(this.myArrayList);
+        ArrayList<Object> copy = new ArrayList<Object>(this.myArrayList.size());
+        for (Object o : this.myArrayList) copy.add((Object) o.clone());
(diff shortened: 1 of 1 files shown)


johnjaylward:
It feel like it's bad form to expose the internal array list like this. It could lead to unexpected modifications. I'd either wrap it in a Collections.unmodifiableArray or do an array copy.

Spedwards:
@johnjaylward The internal ArrayList is private. How could it lead to unexpected modifications?

johnjaylward:
your new method is exposing it.

On Fri, Jul 3, 2015 at 12:39 PM, Liam <notifications@github.com> wrote:

> @johnjaylward <https://github.com/johnjaylward> The internal ArrayList is
> private. How could it lead to unexpected modifications?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/douglascrockford/JSON-java/pull/136#issuecomment-118390343>
> .
>



-- 


John.

stleary:
Thanks, but can't accept this code at the present time. Feel free to open a new pull request if you find an error in the API or have a use case for an enhancement.

johnjaylward:
Sean, I'd be happy about the shallow copy since JSON Java makes no guarantees about thread safety or even immutability. I don't think we'd need a deep copy for this. It would be similar to if someone did:

```java
ja.put(new Object());
Object o = ja.get(0);
// now do something with o that modifies its fields
```
or
```java
ja.put(new Object());
List<?> l = ja.toArrayList();
Object o = l.get(0);
// now do something with o that modifies its fields
```

both end with the same element in ja getting modified. My main concern would be that adding an element to ja should not modify the contents of l, or the other way as well.

stleary:
Good point, that would avoid the problem of using clone(). Returning a List<Object> instead of an ArrayList<Object> would be an improvement as well. However, I still don't see a use case for this method. As you mentioned, users already have a way to recreate the list externally to JSONArray.

Spedwards:
To recreate the ArrayList you have to iterate through the JSONArray. Why do
that when you have an ArrayList in the class that can be used?

stleary:
This is true. However, I can't come up with a compelling reason why this method needs to be added to the API.

Spedwards:
On the flip side, I can't think of a compelling reason why not to add it.

douglascrockford:
That is not a sufficient reason to risk changing code.

PR #138, opened 2015-07-16
Pull request for two enhancements: Node-name-escaping and right handling of NULL-values.

JaXt0r:
Hi there,

for one of my current Java-projects I used your tool for converting JSON --> XML.
The named two circumstances made me fork your project and now I want to give the fixes back to you.

If you have any questions, please let me know.

Best regards
JaXt0r

### DIFF
--- a/XML.java
+++ b/XML.java
@@ -404,7 +404,7 @@ public class XML {
             while (keys.hasNext()) {
                 key = keys.next();
                 value = jo.opt(key);
-                if (value == null) {
+                if (value == null || value == JSONObject.NULL) {
                     value = "";
                 }
                 string = value instanceof String ? (String)value : null;
--- a/XML.java
+++ b/XML.java
@@ -393,7 +393,7 @@ public class XML {
 
             if (tagName != null) {
                 sb.append('<');
-                sb.append(tagName);
+                sb.append(getEscapedNodeKey(tagName));
                 sb.append('>');
             }
 
@@ -434,11 +434,11 @@ public class XML {
                         value = ja.get(i);
                         if (value instanceof JSONArray) {
                             sb.append('<');
-                            sb.append(key);
+                            sb.append(getEscapedNodeKey(key));
                             sb.append('>');
                             sb.append(toString(value));
                             sb.append("</");
-                            sb.append(key);
+                            sb.append(getEscapedNodeKey(key));
                             sb.append('>');
                         } else {
                             sb.append(toString(value, key));
@@ -446,7 +446,7 @@ public class XML {
                     }
                 } else if ("".equals(value)) {
                     sb.append('<');
-                    sb.append(key);
+                    sb.append(getEscapedNodeKey(key));
                     sb.append("/>");
 
 // Emit a new tag <k>
@@ -460,7 +460,7 @@ public class XML {
 // Emit the </tagname> close tag
 
                 sb.append("</");
-                sb.append(tagName);
+                sb.append(getEscapedNodeKey(tagName));
                 sb.append('>');
             }
             return sb.toString();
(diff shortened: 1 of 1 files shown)


stleary:
Changes to functionality  have a high bar for acceptance, lest we break someone's working code. This looks like a bug fix; feel free to provide more information about what it is fixing and/or use cases demonstrating how the existing code is broken. You can also use or modify the JSON-Java unit tests located at https://github.com/stleary/JSON-Java-unit-test. They have a low bar for accepting changes. Otherwise, thanks for the pull request, will investigate.

johnjaylward:
the second commit for the NULL check looks like a bug fix. The first commit lookes related to #123

stleary:
Regarding the fix in XML.toString() for JSONObject.NULL. See JSONObjectTest  jsonObjectNullOperations(). It seems that Java null and JSONObject.NULL are intentionally handled differently. Unless someone can make a case that this is an error, I don't see anything to fix.

stleary:
@JaXt0r, Please provide some sample values that you are converting with *getEscapedNodeKey()*.

JaXt0r:
Hello @stleary ,

I created two commits, which shows the desired behaviour of my pull-request. You already merged them.
They are:
1. https://github.com/stleary/JSON-Java-unit-test/pull/18
1. https://github.com/stleary/JSON-Java-unit-test/pull/19

Thanks for your quick reply.
I'm interested in your answer about the usefullness.

PS: Can you please also look at my comment at https://github.com/douglascrockford/JSON-java/pull/123#issuecomment-122658685

Like it to help
JaXt0r

stleary:
Help me understand the problem. Why are all unicode name productions replaced? For example, is not this valid XML?
<水>water</水>

JaXt0r:
The need for replacing disallowed chars is the same argumentation already discussed at https://github.com/douglascrockford/JSON-java/pull/123:
Some applications export JSON so that the application - which have to transform the data into XML - has no choice either to use it or to replace its JSON content before (with another framework etc.) The easier approach for this might be to tell the transformator-tool _json.org.XML_ to give a valid XML with replacements.

So I think there are two options:
1. We don't want to alter any content at all, so we need to tell it at least in comment of XML.toString() method.
2. We give the additional possibility to always have valid XML.

If you think version _2._ has the need to go further, then I have more informations depending on your questions.

---

I found the XML-1.1-declaration for allowed element-chars: http://www.w3.org/TR/2006/REC-xml11-20060816/#sec-starttags (Older specs have similar values for allowed element-values).
--> You're right. If we want to replace element-chars standard-conform we need to use at least this spec-value for replacement:
```
STag            ::=     '<' Name (S Attribute)* S? '>'
Name	        ::=   	NameStartChar (NameChar)*
NameStartChar   ::=   	":" | [A-Z] | "_" | [a-z] | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x2FF] | [#x370-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D] | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF] | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]
NameChar	    ::=   	NameStartChar | "-" | "." | [0-9] | #xB7 | [#x0300-#x036F] | [#x203F-#x2040]
```
--> If you want to use it I can effort coding this regex-checker-method.

To give the user the opportunity to choose its behaviour in such a non-valid-name-situation my pragmatic solution is to give 3 API methods to the user (According your last comments _[Later update]_).
--> If we don't want to break existing code we can also override the existing toString method like the following example.
--> Example is another solution for https://github.com/douglascrockford/JSON-java/pull/123#issuecomment-122658685
```java
// Original implementation.
// Leaved unchanged for backwards compatibility.
XML.toString()
// New method which replaces invalid node-names with valid ones.
XML.toString(char replacement)
// New method which tells the user if value can't be transformed directly. If param=false the original implementation of XML.toString() is used. Otherwise there will be an additional validation for every node (tag).
// Parameter needs to be here so thats an overridden method in JAVA.
XML.toString(boolean informAboutNonValidXML) throws JSONException
```
--> If you're interested in this three possibilities, then I can create a first version of these API with implementation.

Let me know, which way you think is preferred.

stleary:
Does JSON-java support XML 1.0 or 1.1? My opinion is that only 1.0 should be supported. If anyone thinks otherwise, please let me know. 

It is better to use Name, NameStartChar, and NameChar productions, rather than regex, so please feel free to recode.

Overloading does not seem like the right solution for several reasons, although granted it does allow the user to explicitly choose the behavior they want. 
* Current behavior (in JSONML) is to throw an exception when a transformation would produce invalid XML.
* JSON-java is a reference app. I don't think it should be required to support all use cases
* Overloading  toString(), which overrides the Object method, seems like something to be avoided if possible.

Not sure that any of these reasons are sufficient not to implement a workaround, but being cautious with changes to widely used code is, I think, a good thing. For that reason I am still leaning towards throwing an exception when invalid chars are detected.

stleary:
No further actions planned at this time.

PR #142, opened 2015-07-28
fix #141 stackoverflow when creating JSONObject from an object which …

### LINKED ISSUE #141: Stackoverflow when creating JSONObject from an object with a getter which is returning itself
Calling 
`new JSONObject(new RecursiveObject());`

results in StackOverflow.

```
public static class RecursiveObject {
	public RecursiveObject getCurrent() {
		return this;
	}
}
```

saral:
…returns itself in a getter

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -1099,7 +1099,7 @@ public class JSONObject {
                         }
 
                         Object result = method.invoke(bean, (Object[]) null);
-                        if (result != null) {
+                        if (result != null && result != bean) {
                             this.map.put(key, wrap(result));
                         }
                     }


johnjaylward:
this only addresses a single use case if is not depth independent.

stleary:
Please see discussion in https://github.com/douglascrockford/JSON-java/issues/141. Thank you for the offer. The Javadocs already document the current design, which is that only acyclical data structures are supported.

saral:
Thank you for the reply. Javadocs of some other methods contain this information but the javadoc for the constructor method does not reflect this precondition. 
```
/**
     * Construct a JSONObject from an Object using bean getters. It reflects on
     * all of the public methods of the object. For each of the methods with no
     * parameters and a name starting with <code>"get"</code> or
     * <code>"is"</code> followed by an uppercase letter, the method is invoked,
     * and a key and the value returned from the getter method are put into the
     * new JSONObject.
     *
     * The key is formed by removing the <code>"get"</code> or <code>"is"</code>
     * prefix. If the second remaining character is not upper case, then the
     * first character is converted to lower case.
     *
     * For example, if an object has a method named <code>"getName"</code>, and
     * if the result of calling <code>object.getName()</code> is
     * <code>"Larry Fine"</code>, then the JSONObject will contain
     * <code>"name": "Larry Fine"</code>.
     *
     * @param bean
     *            An object that has getter methods that should be used to make
     *            a JSONObject.
     */
    public JSONObject(Object bean) {
        this();
        this.populateMap(bean);
    }
```

PR #143, opened 2015-07-29
change README to add gradle compile line for java 1.7/1.8

divyenduz:
A very common query on the Internet is the ability to add this repository using gradle. Adding a couple of lines to README to make it easier for people.

### DIFF
--- a/README
+++ b/README
@@ -1,6 +1,6 @@
 JSON in Java [package org.json]
 
-This package needs a new owner. I have not used it in over a decade, and I do 
+This package needs a new owner. I have not used it in over a decade, and I do
 not have time to maintain programs that I do not use.
 
 If you think you can give this package a good home, please contact me.
@@ -73,3 +73,8 @@ JSONML.java: JSONML provides support for converting between JSONML and XML.
 XMLTokener.java: XMLTokener extends JSONTokener for parsing XML text.
 
 Unit tests are maintained in a separate project. Contributing developers can test JSON-java pull requests with the code in this project: https://github.com/stleary/JSON-Java-unit-test
+
+##### To include using gradle :-
+
+* compile 'org.json:json:20141113' (Java 1.8 compatible release)
+* compile 'org.json:json:20140107' (Java 1.7 compatible release)


stleary:
Thanks for the suggestion. I think some kind of support for provisioning is a good idea. Not sure that Gradle-oriented text referencing specific tags or commits is the best solution. Open to other ideas.

divyenduz:
Maybe using git release tags with something like Jitpack.io would be a better way than my earlier suggestion ?

stleary:
Which commits, if any, should have a tag?
Could the needs of most users be met by simply providing a jar file containing the latest build?
What would be accomplished by providing additional support for ant, maven, or gradle builds?

divyenduz:
I agree with your point. My view as a user is that I found it only after including the wrong version that it is not 1.8 compatible. 

While a release system won't suit this repository much, as it is not changing that frequently. 
Just a mention of gradle compile code will make it easier I believe for people to include it.

And since it is not changing much frequently, it won't be a pain to maintain that as well.
Just a opinion where we can weigh that the profit is less but still it is a good practice.

Regards,
Divyendu

stleary:
**Proposal: Enhancing JSON-java accessibility**

The JSON-java project includes everything needed to be incorporated into user projects. However, improvements could be made to improve accessibility. None of these changes would involve executable code. 

There are multiple options for managing the code, but for now I will look at the simplest and easiest to perform choices.

Providing a jar file on the GitHub page would go a long way towards making the software more accessible. However, GitHub does not directly support adding  executables for download. Some projects work around this by providing a link on a wiki page to deliverables on an external server. 

The GitHub way to provide deliverables is via releases (See https://github.com/blog/1547-release-your-software). Release artifacts may include release notes, javadoc jar, source jar, bytecode jar, source zip, and tar.gz zip files. A release need not be tied to a single commit, but can span multiple commits. 

In the past, snapshots of the JSON-java project have been made available in the Maven central repository at http://mvnrepository.com/artifact/org.json/json. This seems to be a popular source, but has not been updated since Nov 2014. 

Gradle is a popular build framework that can pull files from the Maven central repository. Although JitPack.io could be used to provide direct access to a GitHub release, just using the existing Maven repository seems more straightforward. It would be reasonable to provide some text in the release notes on how to add a Maven dependency for Gradle builds

GitHub recommends using Semantic versioning (See http://semver.org/), and this seems reasonable for JSON-java. I think a case could be made for making the Java 1.8 commit the 1.0.0 release.

Please post any questions, comments, and concerns to this thread. After a few weeks we should be ready to proceed to the next step.

johnjaylward:
I think it may make sense to continue with the date versioning scheme since that is what is already in the maven repository. Maven/Gradle would see the date version (20141113) as being more recent than the 1.0.0 version, since both tools assume semantic versioning by default and the date would be used as the Major number. I believe Gradle could be configured not to assume semantic versions, but having mixed version numbers may confuse it as well.

Using a maven command like:
```shell
mvn versions:display-dependency-updates
```
will always show the older version using the date format.

Assuming we don't mind this caveat where historic versions look newer, the proposed plan looks acceptable to me.

johnjaylward:
Also, I'm unsure if we could convince apache/maven to delete or re-version the old libraries in the repository. I haven't heard of them doing so previously. If we can convince them, then using semantic versions seems more reasonable.

stleary:
Date versioning also sounds reasonable, as does maintaining consistency with the Maven repository.

vdurmont:
I may be a little late in the discussion but I just wanted to mention an argument in favor of the semantic versioning (something like MAJOR.MINOR.PATCH).

As explained on [semver.org](http://semver.org), the evolution of the versions has a real meaning:

> * MAJOR changes when you make incompatible API changes,
> * MINOR changes when you add functionality in a backwards-compatible manner, and
> * PATCH changes when you make backwards-compatible bug fixes.

For example, you just released a version `20150729` (which is great, thanks!) but I can't know if I can safely update my dependencies unless I dive into the code and check every commit since the last version. It's a lot of work that could easily be avoided with a semantic versioning.

I know that switching to semantic versioning can bring its own problems but there are successful examples of projects that did it ([commons-io](http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22commons-io%22%20AND%20a%3A%22commons-io%22) comes to mind).

douglascrockford:
There will be no more major changes of this package, and ideally there should not few minor changes.

PR #146, opened 2015-08-16
Implement a method for converting XML read from a Reader object to a JSONObject

r0b0tbuilder:
I have implemented a method that overloads XML#toJSONObject(String), but accepts a Reader object as an argument for reading the XML from a character stream.  Most likely, this would be used to convert a XML document to a JSONObject.

The following is example usage of this method to convert an XML file to a JSONObject.

```java
try {
	JSONObject fromXML = XML.toJSONObject(new FileReader(new File("foo.xml")));

	System.out.println(fromXML);

} catch (JSONException e) {
	e.printStackTrace();
} catch (FileNotFoundException e) {
	e.printStackTrace();
}
```

### DIFF
--- a/XML.java
+++ b/XML.java
@@ -367,23 +367,14 @@ public class XML {
      */
     public static JSONObject toJSONObject(Reader reader) throws JSONException {
     	
-    	String string = "";
-    	
-    	BufferedReader br = new BufferedReader(reader);
-    	
-    	String line = "";
-    	
-    	try {
-		while( (line = br.readLine()) != null) {
-			string += line;
- 		}
+        JSONObject jo = new JSONObject();
+        XMLTokener x = new XMLTokener(reader);
 
-	} catch (IOException e) {
-		throw new JSONException(e);
-	}
-    	
-    	return toJSONObject(string);
-    	
+        while (x.more() && x.skipPast("<")) {
+            parse(x, jo, null);
+        }
+
+        return jo;
     }
 
     /**
--- a/XML.java
+++ b/XML.java
@@ -357,6 +357,34 @@ public class XML {
         return jo;
     }
 
+    /**
+     * Works like {@link XML#toJSONObject(String)}, but will read the XML 
+     * from a reader object.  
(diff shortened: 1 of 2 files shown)


stleary:
This is an enhancement; it provides convenient functionality, does not break any unit tests, and I don't think it will affect existing code. Seems reasonable to me. A little inefficient to convert a reader to a string, only to have XMLTokener convert it back into a reader. Might want to consider something like:

XMLTokener.java
```Java
    public XMLTokener(Reader reader) {
       super(reader);
    }
```

XML.java
```Java
    public static JSONObject toJSONObject(Reader reader) throws JSONException {
        XMLTokener x = new XMLTokener(reader);
        while (x.more() && x.skipPast("<")) {
            parse(x, jo, null);
        }
        return jo;
    }
```

Any other opinions?

r0b0tbuilder:
I have used your suggestion.  A commit has also been pushed that introduces minor edits to the javadoc for the method.

r0b0tbuilder:
Does that look good?

stleary:
Looks fine, will need the usual unit tests. You are welcome to work on those if so inclined (https://github.com/stleary/JSON-Java-unit-test). My preference is that all of the tests that exercise XML.toJSONObject(String) also exercise XML.toJSONString(Reader). Right now my top item is https://github.com/douglascrockford/JSON-java/pull/143. That requires some background effort and will be a little while before it is completed.

r0b0tbuilder:
I have never done unit testing before, so I think it would be best if you could first one as an example.  Thanks! :octocat:

stleary:
See https://github.com/stleary/JSON-Java-unit-test/pull/22. Somehow some unrelated changes to comments which I thought I had already pushed are also included in the pull request. But the modified test should be easy to spot. Anyone making additional changes to XMLTest should make pull requests to the xml-reader-remote branch, but if not sure how, push them to the master branch and I will move them over.

stleary:
Not planning to implement this at the present time.

PR #155, opened 2015-10-10
Allow JSONArray constructor that takes object collections to take collections of any kind

bguedas:
Trying to build a new JSONArray from a Collection\<E\> (where E != Object) throws an exception because it uses another constructor that takes an Object as argument which is supposed to be an array.

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -151,10 +151,10 @@ public class JSONArray implements Iterable<Object> {
      * @param collection
      *            A Collection.
      */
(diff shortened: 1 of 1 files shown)


bguedas:
I just noticed that a similar pull request has been closed.

Updating the JSON library from a version without generics (before may 6th 2014!) broke our code because Collection<String> (for instance) is a subtype of Collection, but not of Collection<Object>.

stleary:
Please continue this discussion in https://github.com/douglascrockford/JSON-java/pull/153.

PR #159, opened 2015-10-12
Properly overrides the Exception class.

johnjaylward:
Overriding the exception class should not keep it's own ```cause``` field.

### DIFF
--- a/JSONException.java
+++ b/JSONException.java
@@ -28,7 +28,7 @@ public class JSONException extends RuntimeException {
      * @param cause
      *            The cause.
      */
-    public JSONException(String message, Throwable cause) {
+    public JSONException(final String message, final Throwable cause) {
 	super(message, cause);
     }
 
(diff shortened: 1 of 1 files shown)


stleary:
Can you say a few words about why this needs to be fixed?

johnjaylward:
The Exception base class already has a variable to store the "cause" and the super.getCause() method already handles the cause properly. This pull request simplifies the implementation as well as creates less stack overhead when throwing an exception.

stleary:
Moved to a branch commit and pull request. Please see https://github.com/douglascrockford/JSON-java/pull/161

johnjaylward:
Not to be a jerk, but the way you are merging my changes loses the history that shows me as the committer. If you need to create personal branches to test pull requests that's fine, but you should merge the original pull request once approved to keep the proper committer history.

johnjaylward:
I had already adjusted the date after seeing it modified in your PR. Not sure if that will affect the test.

stleary:
Completed unit tests successfully. No local changes needed. Will leave this pull request open for a few days for comment.

stleary:
The JSONException API already has a Throwable cause, initialized in the ctor and accessed via a getter.

**Changes to the API, code behavior, unit tests, or documentation?**
No

PR #160, opened 2015-10-12
Fixes possible NPE

johnjaylward:
The comparison for ```(object == null) ? "null" : escape(object.toString())``` was doing nothing, ```escape(object.toString())``` would always be called since ```object``` could never be null at this point. I have added a null check around the possible NPE areas so this check actually does something.

If we want the NPE, then I propose we just change the ```(object == null) ? "null" : escape(object.toString())``` tertiary to ```escape(object.toString())```

### DIFF
--- a/XML.java
+++ b/XML.java
@@ -468,10 +468,12 @@ public class XML {
 // XML does not have good support for arrays. If an array appears in a place
 // where XML is lacking, synthesize an <array> element.
 
-        } else {
+        }
+        if(object!=null){
             if (object.getClass().isArray()) {
                 object = new JSONArray(object);
             }
+            
             if (object instanceof JSONArray) {
                 ja = (JSONArray)object;
                 length = ja.length();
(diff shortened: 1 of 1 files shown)


stleary:
Looks like the intended behavior is for XML.toString(Object) to emit the string "\"null\"" on line 483 when the object is null. However, an NPE is thrown when line 472 is executed, so this never happens. The unit tests expose the bug, but this was not understood at the time. Proposal is to fix this by rearranging the flow of control so that when the object is null, line 472 is not executed but line 483 is executed. The unit tests will need to be updated as well. See https://github.com/stleary/JSON-Java-unit-test/tree/XML-npe-test-fix.

stleary:
XML.toString(null) is intended to return "null" (in quotes), but instead a NullPointerException is thrown. This code enables the correct behavior.

**Changes to the API?**
No

The only change is within XML.toString(). The flow of control is corrected so that "null" is returned when the parameter is null.

Yes,  https://github.com/stleary/JSON-Java-unit-test/pull/26 will need to be committed when this change is pushed.

No, this is an internal bug fix.

PR #169, opened 2015-10-26
Empty arrays fix

4boss:
If array field is empty in source json then it will be completely omitted during json-> xml transformation. Here is small fix for this problem.

### DIFF
--- a/XML.java
+++ b/XML.java
@@ -430,6 +430,9 @@ public class XML {
                 } else if (value instanceof JSONArray) {
                     ja = (JSONArray)value;
                     length = ja.length();
+					if(length == 0) {
+                        sb.append(toString("", key));
+                    }
                     for (i = 0; i < length; i += 1) {
                         value = ja.get(i);
                         if (value instanceof JSONArray) {
--- a/XML.java
+++ b/XML.java
@@ -430,7 +430,7 @@ public class XML {
                 } else if (value instanceof JSONArray) {
                     ja = (JSONArray)value;
                     length = ja.length();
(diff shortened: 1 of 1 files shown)


stleary:
Are you proposing this as a bug fix, or an enhancement? If a bug fix, please include a description of how it is causing a problem. Commits that change behavior are likely to be declined unless we can justify the change.

johnjaylward:
From what I can see this PR would break XML output for json arrays:

Test:
```java
JSONObject jo1 = new JSONObject();
jo1.put("arr",new JSONArray(new String[]{"One", "Two", "Three"}));
JSONObject jo2 = new JSONObject();
jo2.put("arr",new JSONArray());

String output1 = XML.toString(jo1,"jo1");
String output2 = XML.toString(jo2,"jo2");
```

Output Before:
```xml
<jo1><arr>One</arr><arr>Two</arr><arr>Three</arr></jo1>
<jo2></jo2>
```

Output After:
```xml
<jo1><arr>One</arr><arr>Two</arr><arr>Three</arr></jo1>
<jo2><arr/></jo2>
```

The "after" output, when read back into a JSON object would create a structure that looks something like this:
```json
{
"arr": [null]
}
```

or possibly:
```json
{
"arr": [""]
}
```

4boss:
I am using produced XML to generate xsd schema, which is weird a little bit, I know :) but from this perspective empty tag much better than nothing.

stleary:
Thanks, but can't include this change at the present time.

PR #170, opened 2015-10-26
Handle arrays consistently

johnjaylward:
There is some code in the XML converter that tries to wrap arrays into a JSONArray, but does so inconsistently. This causes some disconnect between the handling of a native Java Array and the JSONArray object when conversion to XML occurs.

This PR also updates the loops over the JSONArrays to use the new iterable for loops introduced in Java5 as well as updates some comment formatting to be more consistent with other files in the project.

### DIFF
--- a/XML.java
+++ b/XML.java
@@ -27,49 +27,54 @@ SOFTWARE.
 import java.util.Iterator;
 
 /**
- * This provides static methods to convert an XML text into a JSONObject,
- * and to covert a JSONObject into an XML text.
+ * This provides static methods to convert an XML text into a JSONObject, and to
+ * covert a JSONObject into an XML text.
+ * 
  * @author JSON.org
  * @version 2015-10-18
  */
+@SuppressWarnings("boxing")
 public class XML {
 
     /** The Character '&amp;'. */
-    public static final Character AMP   = '&';
+    public static final Character AMP = '&';
 
     /** The Character '''. */
-    public static final Character APOS  = '\'';
+    public static final Character APOS = '\'';
 
     /** The Character '!'. */
-    public static final Character BANG  = '!';
+    public static final Character BANG = '!';
 
     /** The Character '='. */
-    public static final Character EQ    = '=';
+    public static final Character EQ = '=';
 
     /** The Character '>'. */
-    public static final Character GT    = '>';
+    public static final Character GT = '>';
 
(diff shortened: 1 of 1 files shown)


stleary:
Thanks, investigating.

stleary:
Does not break the unit tests. A bit difficult to isolate the functional change due to all of the style updates. In toString() Array values found in a JSONObject are converted into JSONArray instances:

XML.toString()
Insert at line 409 of the original code:

    } else if (value.getClass().isArray()) {
        value = new JSONArray(value);
    }

**New behavior:**
* If within a "content" key, emit the value as a string of non-escaped line-delimited JSONArray entries, instead of an escaped value.toString(). This should not be an issue since the JSONArray contents are escaped when toString() is called recursively later.
* Otherwise, emit the value as a JSONArray instead of directly invoking toString(value, key). Each JSONArray entry that is a JSONArray is emitted as key element with value text. Other entries are emitted via toString(value, key).

**Unrelated note:**
Don't think line 410 of the original code does anything, due to the return on line 466.
                string = value instanceof String ? (String)value : null;

Seems like a good enhancement, but needs more study of how the behavior is changed.

johnjaylward:
I agree with you about line 410. I think that would be outside the scope of this PR though, unless you want me to include it.

stleary:
XML.toString() does not work as expected when a JSONObject contains Array type values.

**Changes to the API?**
No.

When processing JSONObject values, Arrays will be converted to JSONArrays before 
further processing. This results in consistent handling between raw JSON, JSONArray
values and Array values.
Example
```
{
    "arr" : [
        "One",
        [
             "Two",
             "Three"
        ],
         "Four",
         []
     ]
}
```
Produces this XML from raw JSON or internal JSONArrays (correct):
```
    <arr>One</arr>
    <arr>
        <array>Two</array>
        <array>Three</array>
    </arr>
    <arr>Four</arr>
    <arr></arr>
```
Produces this XML from internal arrays (incorrect):
```
    <arr>One</arr>
    <arr>Two</arr>
    <arr>Three</arr>
    <arr>Four</arr>
```
The code fix will bring the Array value in sync with raw JSON and JSONArray.

Yes, the unit tests will be updated. See https://github.com/stleary/JSON-Java-unit-test/pull/29.

No, this is an internal change.

PR #179, opened 2015-12-02
Update JavaDoc for JSONObject Constructors

andrew-fletcher:
Two JSONObject constructors incorrectly specify a @throws JSONException
tag in the JavaDoc for those constructors. Remove the relevant JavaDoc.


stleary:
Thanks, seems reasonable.

PR #180, opened 2015-12-07
Remove executable bit

netheril96:
The source files should not have executable permission set.


stleary:
The files in the pull request do not show any changes, so far as I can tell.
Are you sure the executable bit being set is not an artifact from your Git client?

netheril96:
@stleary 

json-java

GitHub finds 14 files with *mode* changes. Are you on Windows, where the permission system is different from what git expects?

stleary:
Yes, I am on Windows.

netheril96:
@stleary Windows has a different permission system so you may not find anything different. It surely is different on Unix.

stleary:
Are the permissions breaking anything?

netheril96:
@stleary Not yet. But it is inconsistent and annoying.

johnjaylward:
@stleary this is what it currently looks like on OS X or linux
execbit

On Unix, this makes the source code executable to the OS. In general this is frowned upon as a security risk

stleary:
@johnjaylward thanks, I think it would be reasonable to have the code be as correct as possible, including the permissions. Are you recommending just making sure the executable bit is cleared for all files?

johnjaylward:
yeah, that's what the little note that Github is showing you this pull request does:
    CDL.java 100755 → 100644
755 is
* Executable bit set for "User, Group, Other"
* Read set for "User, Group, Other"
* Write set for "User"

644 is:
* Executable bit not set
* Read set for "User, Group, Other"
* Write set for "User"

johnjaylward:
here's a brief overview of the numeric notation github is showing: https://en.wikipedia.org/wiki/File_system_permissions#Numeric_notation

stleary:
@netheril96 OK, seems reasonable, thanks for the suggestion. It would be helpful if you could update the "version" field in the JavaDoc for each file, replacing the date you find there with  the current date, using this format: yyyy-mm-dd.  The commit comment should be something simple, such as,  "Remove executable permission bit from file mode", or whatever you think is appropriate. Thanks, Sean

netheril96:
@stleary Done

stleary:
Thanks, files have the correct mode bits set now, unit tests pass, will include this in the next release.

johnjaylward:
Verified the correct permissions are set.

execbit

PR #181, opened 2015-12-09
multi-catch fixed for source release compatibility 1.6

jinahya:
Two lines in each file fixed for source code compatible `1.6`.
I believe these changes don't affect any functionalities and API changes.
Note that sticking source release `1.6` doesn't mean target class version should be `1.6`.

```
$ javac -source 1.6 -target 1.6 -d target *.java
warning: [options] bootstrap class path not set in conjunction with -source 1.6
1 warning
$ javac -source 1.6 -target 1.7 -d target *.java
warning: [options] bootstrap class path not set in conjunction with -source 1.6
1 warning
$ javac -source 1.6 -target 1.8 -d target *.java
warning: [options] bootstrap class path not set in conjunction with -source 1.6
1 warning
$
```
Note, also, that even [Java 9 might support `1.6`](http://openjdk.java.net/jeps/182).

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -593,7 +593,9 @@ public class JSONArray implements Iterable<Object> {
                 return myE;
             }
             return Enum.valueOf(clazz, val.toString());
-        } catch (IllegalArgumentException | NullPointerException e) {
+        } catch (IllegalArgumentException iae) {
+            return defaultValue;
+        } catch (NullPointerException npe) {
             return defaultValue;
         }
     }
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -900,7 +900,9 @@ public class JSONObject {
                 return myE;
             }
             return Enum.valueOf(clazz, val.toString());
-        } catch (IllegalArgumentException | NullPointerException e) {
+        } catch (IllegalArgumentException iae) {
+            return defaultValue;
+        } catch (NullPointerException npe) {
             return defaultValue;
         }
     }


stleary:
Additional discussion of this code change can be found at the bottom of https://github.com/douglascrockford/JSON-java/issues/156

stleary:
I think the first item is to determine what Android Development Studio thinks of the JSON-java code, unless anyone has other suggestions.

johnjaylward:
Some links for ease of not switching between issues:
* [Android Market Share](https://developer.android.com/about/dashboards/index.html). API 19 and above can use Java1.7 features. At this time it's about 66% of android devices. About 33% require Java1.6 or lower.
* [Oracle Java EOL](http://www.oracle.com/technetwork/java/eol-135779.html). Java 1.7 and lower have been EOL by Oracle. Only java8+ runtimes will receive security and bug fixes.
* [json-retrotranslated](https://github.com/jinahya/json-retrotranslated) a project that forks this one to support older STBs and PCBs that run java 1.3

stleary:
Agreed that the only code preventing 1. 6 compatibility are the multi-catches, which were a recent enhancement. However, I have been unable to make the case that the existing JSON-java code fails to compile in Android Studio, and deploy/execute on an Android device (see my comment, above). What exactly is your use case for making this change?

jinahya:
I officially withdraw this PR. You can ignore/decline this pull request. Thanks.

stleary:
@jinahya feel free to reopen if you come across any situation where reverting to 1.6 compatibility would provide a benefit to users.

PR #186, opened 2016-01-01
Update README with number information

stleary:
See https://github.com/douglascrockford/JSON-java/issues/162


kushieda-minori:
You should fix the wrapping of the urls to be on 1 line. Otherwise looks good.

johnjaylward:
You should fix the wrapping of the urls to be on 1 line. Otherwise looks good.

Issue #192, opened 2016-02-03
Maven or Gradle project?


nhajratw:
Any interest in having this be a proper maven or gradle project? I'd be happy to contribute a pull request for either. Let me know, and i'll get working :-)

erosb:
Recently I found this repo: https://github.com/eskatos/org.json-java
I think this is where the maven arifacts are built from and are deployed to maven central. It includes https://github.com/stleary/JSON-java as a git submodule. But I'm not sure.

Anyway I'd also like to see this repo as a maven project, so +1

nhajratw:
interesting -- this is kind of a strange setup. Seems like it should all be in a single repo.

johnjaylward:
no, that one is different.

The main reason that this is not a gradle or a maven project is that when it first started there were no test cases, and still to this day, there are no external dependencies. This makes building the project very straightforward and easy; e.g. ```javac *.java```

Historically and PR that have been given to mavenize or gradleize the project have been denied due to the original maintainers lack of interest in it and the easy of use in building it. You can search the old PR and issues.

However, now that we have a decent test bench, I think it makes more sense to combine the 2 projects into a single one and possible use Maven or Gradle to manage the tests. It would make contributing easier since we wouldn't need 2 pull requests, one for the bug/feature and the other for tests.

If you do create a PR for this, I'd recommend including the test bench code and updating the README to reflect the new tests location. The test bench currently had some gradle build info in it that may be a good place to start.

nhajratw:
Thanks @johnjaylward -- I am certainly happy to pull all of this together. I expect it won't take more than a couple hours. I'll give it a shot and submit a pull request. If it ultimately is denied, it will have been an interesting exercise anyway.

stleary:
@nhajratw, appreciate the effort of putting together the pull request. Agree with @johnjaylward that it is clumsy to have the unit tests in a separate project. However, for now I think there is still value in a simple project that only requires the tools provided by the JDK. Will leave this issue open for a few more days in case anyone wants to voice an opinion.

erosb:
@nhajratw kudos for working on it! I'd really like to see it merged. Not only because of gradle, but because having the tests in the same repo is essential for external contributors. When #187 came up, the lack of tests stopped me to start fixing it on my own (I couldn't find them anywhere).

nhajratw:
[...]

PR #193, opened 2016-02-03
Create gradle project

nhajratw:
Converted to a Gradle project, and incorporated tests, per issue #192


johnjaylward:
Just set the source and target comparabilities to 1.7 instead of 1.8

stleary:
Thanks, not not accepting this at the present time.

PR #202, opened 2016-02-27
Add equals and hashCode to JSONObject and JSONArray

joeferner:


johnjaylward:
Map/Array equality is often not a use-case for equality of JSON Objects or JSON Arrays. Also, both objects already have a ```similar``` method that can be used.

* http://stleary.github.io/JSON-java/org/json/JSONObject.html#similar-java.lang.Object-
* http://stleary.github.io/JSON-java/org/json/JSONArray.html#similar-java.lang.Object-

joeferner:
How is `similar` different from `equals`? It looks like the existing implementation is very similar to the existing implementation of `java.util.Map#equals` With the implementation of `equals` in both `JSONObject` and `JSONArray` this method should no longer be needed.

I also have a use case to store `JSONObject`s in a `java.util.Set` and would like to obviously avoid duplicates. To do this `equals` and `hashcode` must be implemented.

I would also like to have `equals` to allow `http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertEquals(java.lang.Object, java.lang.Object)` to work.

stleary:
@johnjaylward perhaps the intent is to check the equality of nested JSONObjects and JSONArrays, which might be a helpful enhancement.

I think this was last proposed in https://github.com/stleary/JSON-java/pull/33. At that time, the concern was for cyclic objects and imperfect equality implementation. Perhaps the former could be addressed with a "Warning: This method assumes that the data structure is acyclical." comment on the equals() method. The new equals() implementation itself seems to be OK.

What was not brought up in 33 was whether existing implementations might rely on the default equals() behavior. But that seems to be a minor corner case.  Also, this PR does not break the unit tests. 

What is your recommendation?

johnjaylward:
The only worry is have at this change is performance for large objects.
Currently all my personal code works fine assuming each instance is a
different object since I only use JSON as an intermediary between systems.
The original maps and sets handle any duplicates before conversation to
json.

If people are using JSONObject as a first class object and not just for
information exchange, then I think this pull request would be required.

However, I don't think this library was really designed for that. JSON
itself is only meant to exchange information between systems and this
library implementation shows that.

Json is really meant to serialize and deserialze Java beans. It is not
performant for comparisons and other intensive tasks.

joeferner:
@johnjaylward I don't see why `JSONObject` and `JSONArray` are all that different from `Map` and `List`. Should they not exhibit a similar behavior?

johnjaylward:
For me they are the result of serialization or the input for
deserialization. As an end/start product, I don't need the extra overhead
of checks that already occurred elsewhere in my applications.

If we look at other implementations like in JavaScript

```javascript
o1 = { "id" = 1 };
o2 = { "id" = 1 };
if o2 == o1 {
   // mostly equal
}
if o2 === o1 {
  // exactly true
}
```

Neither of the JavaScript if statements evaluate to true.

johnjaylward:
Hmm, email replies stink... Sorry for the bbad formatting

johnjaylward:
@joeferner To answer your question about differences between similar and equals:

To quote http://www.json.org/, "JSON (JavaScript Object Notation) is a lightweight data-interchange format. It is easy for humans to read and write. It is easy for machines to parse and generate."

As such, each JSON is unique in that each one is a new message in the data interchange. We can have similar messages that look the same, but they are still unique messages.

Once the messages are decoded, the receiving system can then determine equality. Sometimes this may be the entire message, sometimes it may be a unique ID, and other times still it may be a combination of fields within the message.

By having a separate ```similar``` method, this lets people check for exact duplicate messages before decoding them without forcing a generic equality check on all messages that may or may not be accurate.

stleary:
Thanks, but can't accept this pull request at the present time, given the concerns about:
* Performance with large datasets
* The meaning of equality in the context of the underlying serialized data
* The focus of this project as a reference app

PR #203, opened 2016-02-27
Adds JSONArray toList method and JSONObject toMap method

joeferner:


stleary:
Seems like a reasonable enhancement. 

Should the methods be renamed as getters, e.g. getMap() instead of toMap() ?
Should the return values be Map<?,?> and List<?> ?

Otherwise, no objections here. 
Any other opinions?

johnjaylward:
I think the toMap and toList are reasonable names as they imply a
conversion, which is what the implementation is doing.

johnjaylward:
I didn't check the javadoc included, but be sure a note for cyclical
objects is there.

joeferner:
Just added the warning to both methods

stleary:
Provide methods to facilitate access to a shallow copy of the data stored in JSONObject and JSONArray. During the copy, nested JSONObjects are converted into maps and nested JSONArrays are converted into lists. 

**Changes to the API?**
Yes, 2 new methods are added:
````
JSONArray: public List<Object> toList();
JSONObject public Map<String, Object> toMap();
````

No.

No, but new unit tests will be added after this commit. See https://github.com/stleary/JSON-Java-unit-test/pull/43.

No, this will be rolled into the next release, which is not yet scheduled.

The online JavaDoc should be split into latest release and latest code branches. The latest code branch should be updated with the new methods. The readme file will be updated to point to both JavaDoc pages. It is not necessary to complete this before merging the code.

PR #206, opened 2016-03-05
Added getDate and optGetDate to JSONObject

pasqumirk:
Hi, I added two functions to JSONObject to help retrieving a date from the object


erosb:
Date is not a built-in type in JSON, therefore I don't think it is a good idea to add these methods. I think the scope of this library should remain "handling JSON data" and it doesn't have to support additional kinds of data people use JSON to represent.

PR #208, opened 2016-03-09
 JSONObject from an Object for private members

anjalipatelit93:
JSONObject from an Object was not working for private members.

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -290,9 +290,11 @@ public class JSONObject {
         for (int i = 0; i < names.length; i += 1) {
             String name = names[i];
             try {
-                this.putOpt(name, c.getField(name).get(object));
-            } catch (Exception ignore) {
-            }
+           				Field field = c.getDeclaredField(name);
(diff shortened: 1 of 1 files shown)


johnjaylward:
JSON is a data interchange format. As such, only public methods and fields should be exposed via JSON. Any protected, private, or package level items should not be serialized. Unlike Java binary serialization, JSON does not promise to maintain object state.

amitbhoraniya:
If we want convert POJO classes to JSON String, then currently we can not convert POJO class to JSON String, because as JAVA Standard fields of POJO classes are private but getter and setter methods are public, So We should provide provision for that, else we should convert to JSON string using getter methods.

stleary:
@amitbhoraniya, I think JSON-Java does support getters and setters via the Bean API. 
@anjalipatelit93, Thanks for the pull request, but it can't be accepted at this time. 

Thought I had responded to this yesterday, but must have forgotten to hit the comment button.

PR #215, opened 2016-04-06
add method hasIngoreCase of JSONObject

MrGlory:
In my project,I need find a ignore case key.So I add it


stleary:
Thanks, but don't see a use case for including this API method at the present time.

PR #216, opened 2016-04-07
Added support to preserve the order when converting XML to JSON

susinda:
Introduced (overload)method XML.toJSONObject(String string, boolean preserveOrder)

### DIFF
--- a/XML.java
+++ b/XML.java
@@ -332,6 +332,37 @@ public class XML {
         }
         return jo;
     }
+    
+    /**
+     * Convert a well-formed (but not necessarily valid) XML string into a
+     * JSONObject. Some information may be lost in this transformation because
+     * JSON is a data format and XML is a document format. XML uses elements,
+     * attributes, and content text, while JSON uses unordered collections of
+     * name/value pairs and arrays of values. JSON does not does not like to
+     * distinguish between elements and attributes. Sequences of similar
+     * elements are represented as JSONArrays. Content text may be placed in a
+     * "content" member. Comments, prologs, DTDs, and <code>&lt;[ [ ]]></code>
+     * are ignored.
+     * 
+     * @param string 
+     *            The source string.
+     * @param boolean 
+     *            flag saying whether to maintain order or not.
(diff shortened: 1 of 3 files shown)


stleary:
Thanks, but can't accept this pull request. The spec says JSON objects are unordered.

susinda:
Hi Stleary
Thanks for looking my PR.
What I meant by order is to maintain the same order as in xml (when converting xml to json). For simple xml files order is not changed (even without this fix) issue comes for complex xmls. If we maintain the same order, is it a violation of spec.

Thanks
Susinda

stleary:
Maintaining order in JSON objects, even for a good reason, would cause JSON-Java to diverge from the spec, and lesson its value as a reference app. If you check the closed pull requests, this is probably the 2nd most popular topic, after Mavenizing the project.

johnjaylward:
By definition, if you are converting to JSON, it means you don't care what order your object keys are in. If you need ordered keys, you should use a different data interchange format.

susinda:
Thanks stleary and johnjaylward.
Understood and agree with your explanation.  Is there any chance that you could make this functionality extendable so that ppl(like me) who needs the order can extend and use it.
Anyway i'd like to thank you and all the developers for your effort on development and maintenance this.

XML.java
```diff
@@ -332,6 +332,37 @@ public static JSONObject toJSONObject(String string) throws JSONException {
...
+     * @param boolean 
+     *            flag saying whether to maintain order or not.
+     * @return A JSONObject containing the structured data from the XML string.
+     * @throws JSONException
+     */
+    public static JSONObject toJSONObject(String string, boolean preserveOrder) throws JSONException {
+        if (!preserveOrder) {
+        	return toJSONObject(string);
+        } else {
+        	JSONObject jo = new OrderedJSONObject();
```
meetjoeblack: As I see we have new JSONObject() creation inside XML#parse https://github.com/stleary/JSON-java/blob/28e09dc4938e65bbf022302972559ac0179b5ebc/XML.java#L327 method and your changes doesn't cover this case.

Issue #218, opened 2016-04-12
RFC 6901 (JSON Pointer) support


erosb:
What do you think about adding [JSON Pointer](https://tools.ietf.org/html/rfc6901#section-5) support to the library?

From the API point of view I think it would mean a new method with a signature like `Object query(String jsonPointer)`, both in `JSONObject` and `JSONArray`.

@stleary if you give green light for this proposal then I may work on it.

johnjaylward:
is this like XML Path but for JSON?

erosb:
Exactly (and simpler too).

stleary:
Torn between liking the idea and thinking maybe it is not part of the purpose of the lib. Hard to decide because I am not familiar with it - I use JsonPath in the unit tests. JsonPointer does not seem to have wildcards. Is this the case?

erosb:
I've never used JsonPath, but it doesn't seem to be a standard, so I would still prefer implementing JSON Pointer. JsonPath also seems to be a more complex language, deserving its own library, while a JSON Pointer implementation would be small-enough to fit into only a few methods in this project.

stleary:
Please proceed with JSON Pointer if you are still interested. Let's see what it would look like.

stleary:
Please see https://github.com/stleary/JSON-java/pull/222

erosb:
@stleary will youl merge #222 ?

stleary:
@erosb I expect so. The implementation looks good so far. Have not had a chance to go over the unit test cases or coverage yet.

PR #222, opened 2016-04-26
JSON Pointer implementation

erosb:
See #218 for details

### DIFF
--- a/JSONPointer.java
+++ b/JSONPointer.java
@@ -5,6 +5,7 @@ import static java.util.Collections.emptyList;
 
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
+import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -37,8 +38,34 @@ SOFTWARE.
  * <a href="https://tools.ietf.org/html/rfc6901">RFC 6901</a>. 
  */
 public class JSONPointer {
+    
+    private static final String ENCODING = "utf-8";
 
-    private List<String> refTokens;
+    public static class Builder {
+        
+        private final List<String> refTokens = new ArrayList<String>();
+
+        public Builder append(String token) {
+            refTokens.add(token);
+            return this;
+        }
+
+        public JSONPointer build() {
+            return new JSONPointer(refTokens);
+        }
+
+        public Builder append(int arrayIndex) {
+            refTokens.add(String.valueOf(arrayIndex));
+            return this;
+        }
+        
+    }
+    
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    private final List<String> refTokens;
 
     /**
      * Pre-parses and initializes a new {@code JSONPointer} instance. If you want to
@@ -59,7 +86,7 @@ public class JSONPointer {
         if (pointer.startsWith("#/")) {
             pointer = pointer.substring(2);
             try {
-                pointer = URLDecoder.decode(pointer, "utf-8");
+                pointer = URLDecoder.decode(pointer, ENCODING);
             } catch (UnsupportedEncodingException e) {
                 throw new RuntimeException(e);
             }
@@ -72,6 +99,10 @@ public class JSONPointer {
         for (String token : pointer.split("/")) {
             refTokens.add(unescape(token));
         }
+    }
+
+    public JSONPointer(List<String> refTokens) {
+        this.refTokens = refTokens;
     }
 
     private String unescape(String token) {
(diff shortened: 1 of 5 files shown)


johnjaylward:
This looks like a nice simple implementation to me. I have not looked at the test cases to see if they cover the different pointer types available.

@erosb can you also update the README to indicate which RFC this is implementing?

erosb:
@johnjaylward  I've just added it in https://github.com/stleary/JSON-java/pull/222/commits/cbb1546c530d73d55c51b903b9c14962f4f40671

stleary:
This is a new feature for JSON-Java. It implements JsonPointer as specified in https://tools.ietf.org/html/rfc6901.

**Changes to the API?**
Yes. 2 new public methods are defined:
* Object JSONObject.query(String)
* Object JSONArray.query(String)

2 new classes are also defined:
* JSONPointer
* JSONPointerException

No, but new unit tests will be added after this commit. Those tests will fail unless this code is committed. See https://github.com/stleary/JSON-Java-unit-test/pull/46.

No, but this will be rolled into the next release, which is expected in May 2016.

Yes. The JavaDoc and the README should be updated.

johnjaylward:
I finally got around to looking at the test cases and saw this:
```java
assertSame(document.get("c%d"), query("#/c%25d"));
```

I was wondering if it made sense to provider a "builder" where you can add parts. Something like this maybe:
``java
JSONPointerBuilder jpb = new JSONPointerBuilder(); // will return the whole document
jpb.append("c%25");
```
the ```append``` method would return ```this``` like any normal builder so you can chain the commands.

So the example from the RFC ```"/foo/0"``` would look something like this with a builder:
```java
JSONPointer jp = new JSONPointerBuilder().append("foo").append(0).compile();
```

maybe this would be better in a separate pull request?

erosb:
Hello @johnjaylward , the above `"#/c%25d"` query has the same semantics as
`"/c%d"`, the difference is just the first one being in [
https://tools.ietf.org/html/rfc6901#section-6](URI fragment representation)
while the first one is [https://tools.ietf.org/html/rfc6901#section-5](JSON
string representation). In other words the percent-encoding is optional and
is necessary only if the pointer starts with `#`. For this purpose I don't
think it is necessary to create a builder API. Maybe in the future, if it
will be requested.

2016-04-29 18:05 GMT+02:00 johnjaylward <notifications@github.com>:

> I finally got around to looking at the test cases and saw this:
>
> assertSame(document.get("c%d"), query("#/c%25d"));
>
> I was wondering if it made sense to provider a "builder" where you can add
> parts. Something like this maybe:
> ``java
> JSONPointerBuilder jpb = new JSONPointerBuilder(); // will return the
> whole document
> jpb.append("c%25");
>
> the ```append``` method would return ```this``` like any normal builder so you can chain the commands.
>
> So the example from the RFC ```"/foo/0"``` would look something like this with a builder:
> ```java
> JSONPointer jp = new JSONPointerBuilder().append("foo").append(0).compile();
>
> maybe this would be better in a separate pull request?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/stleary/JSON-java/pull/222#issuecomment-215781110>
>



-- 


*Bence Erősme@github <https://github.com/erosb>*

erosb:
@johnjaylward @stleary I've just updated the PR according to the above suggestions.

JSONPointer.java
```diff
@@ -37,8 +38,34 @@ of this software and associated documentation files (the "Software"), to deal
  * <a href="https://tools.ietf.org/html/rfc6901">RFC 6901</a>. 
  */
 public class JSONPointer {
+    
+    private static final String ENCODING = "utf-8";
```
johnjaylward: when we drop Java6 support, this can be changed to https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html

java.nio.charset.StandardCharsets.UTF_8

maybe put in a comment noting that.

JSONPointer.java
```diff
@@ -74,6 +101,10 @@ public JSONPointer(String pointer) {
         }
     }
 
+    public JSONPointer(List<String> refTokens) {
```
johnjaylward: I we want this could either be a protected constructor, or the List should be copied so the Pointer can't be modified outside of the class.

this.refTokens = new ArrayList<>(refTokens);

johnjaylward:
@erosb I made a few inline comments on the commits. I think the one constructor needs modification, but other than that it looks good. Pretty much what I was thinking of.

JSONPointer.java
```diff
@@ -0,0 +1,226 @@
...
+        } else {
+            throw new IllegalArgumentException("a JSON pointer should start with '/' or '#/'");
+        }
+        refTokens = new ArrayList<String>();
+        for (String token : pointer.split("/")) {
+            refTokens.add(unescape(token));
+        }
+    }
+
+    public JSONPointer(List<String> refTokens) {
```
johnjaylward: This could either be a **protected** constructor, or the **List** should be copied so the Pointer can't be modified outside of the class.

this.refTokens = new ArrayList<>(refTokens);

JSONPointer.java
```diff
@@ -0,0 +1,226 @@
...
+SOFTWARE.
+*/
+
+/**
+ * A JSON Pointer is a simple query language defined for JSON documents by
+ * <a href="https://tools.ietf.org/html/rfc6901">RFC 6901</a>. 
+ */
+public class JSONPointer {
+    
+    private static final String ENCODING = "utf-8";
```
johnjaylward: when we drop Java6 support, this can be changed to https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html

java.nio.charset.StandardCharsets.UTF_8

maybe put in a comment noting that.

erosb:
@johnjaylward good point, I've just added token list copying to the constructor. I kept it `public` though, it might be useful for somebody creating the reference token list in a dynamic way.

erosb:
@stleary will you merge it in the next release?

stleary:
@erosb thanks for being patient. No review comments or code changes in over a week. No cautions about adding code that is not central to the purpose of the app. Still need to come up with some content for the readme. Will do a final code review and test run, then should be ready for the merge. Should be able to get to this over the weekend.

PR #229, opened 2016-05-06
Fixed the parsing of the xsi:nil attribute, now outputs a JSON object with "null" for the value.

ahalldin:
This change will now allow XML.java to properly convert XML elements with the xsi:nil=true attribute. Previously elements with this attribute would be converted into JSON elements with "xsi:nil" being added as a separate element underneath. For example, consider the following XML:
`<book> <title>Beowulf</title> <author xsi:nil="true"/> </book>`

Previously, this would be converted to the following format:
`{"book": { "author": {"xsi:nil": true}, "title": "Beowulf" }}`

With this change, the converted JSON will now look like this:
`{"book": { "author": "null", "title": "Beowulf" }}`

### DIFF
--- a/XML.java
+++ b/XML.java
@@ -223,6 +223,7 @@ public class XML {
             tagName = (String) token;
             token = null;
             jsonobject = new JSONObject();
+            boolean nil = false;
             for (;;) {
                 if (token == null) {
                     token = x.nextToken();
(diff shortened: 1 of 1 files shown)


johnjaylward:
As we don't currently support any Schemas in XML processing, I think this would be better handled as a post-process of the JSONObject than as an additional "if-else" block within the XML reader.

What may be a better solution as a whole would be to create an event-listener setup within the XML reading process that can be used.

for example with pseudocode:
```
If an element is read and there is an element listener
 - call the element listener
 - if the element listener wants to modify the element contents, use the listeners result
if an attribute is read and there is an attribute listener
 - call the attribute listener
 - if the attribute listener wants to modify the attribute, use its' result
 - if the attribute listener wants to modify the containing element, use it's result as the element value.
```

stleary:
Thanks for the pull request, but can't accept it at this time. Not convinced that XML schema instance elements should be interpreted as part of the XML-to-JSON conversion, or that if so, only this element should be parsed. Agreed that XML.java is somewhat broken, but would prefer to see proposals to fix the design. I think it would be difficult to fix it without breaking existing code.

Issue #233, opened 2016-05-14
JSONPointer should not always throw JSONPointerException 


stleary:
A user may want to evaluate a JSONPointer which is not expected to match the document. For example, when writing a test case to confirm that a specified item is not present. Having to catch exceptions makes this more difficult. It might be better if the JSONPointer query were to return null instead. I don't think there is a conflict with actual null values, since I believe they are stored as JSONObject.NULL.

Certain other cases may justify throwing an exception. For example, using a syntactically incorrect JSONPointer string.

I don't think the spec requires exceptions to be thrown, nor does it seem to require the same behavior for all types of errors.

All of the cases where JSONPointer throws an exception should be evaluated to determine if returning null might be better. Compatibility of JSONPointer returning null and JSONObject.NULL content should be confirmed.

erosb:
The RFC [says the followings](https://tools.ietf.org/html/rfc6901#page-6): 

>   In the event of an error condition, evaluation of the JSON Pointer
>   fails to complete.
>
>   Error conditions include, but are not limited to:
>
>   o  Invalid pointer syntax
>
>   o  A pointer that references a nonexistent value

In my understanding the current behaviour is correct.

stleary:
I am working from Effective Java, item 57 (Use exceptions only for exceptional conditions) and item 58 (Use checked exceptions for recoverable conditions and runtime exceptions for programming errors). A pointer not maching the data does not seem to rise to the level of exceptional condition or programming error. Besides, it seems burdensome to require the developer to wrap each query in a try/catch block in case the data does not happen to be present. No problem with the invalid syntax part.

johnjaylward:
The spec doesn't say how the error should be generated, just that it should error. I'm not sure if I'd consider retuning null constitutes an error, but it's how many other libraries indicate a "not found" status.

Maybe a better solution would be to offer an "opt" getter like JSONObject and JSONArray currently offer. The normal getters throw an exception if not found, and opt returns either null or a default value.

stleary:
Good idea. I think adding optQuery() methods to JSONObject and JSONArray would be a reasonable compromise. No change would be needed to JSONPointer.

stleary:
Will be fixed in https://github.com/stleary/JSON-java/pull/234

PR #234, opened 2016-05-16
fixing stleary/JSON-Java#233

erosb:


stleary:
Users should have the option of performing a JSONObject or JSONArray query without an exception being thrown when the string does not match an item in the document. The optQuery() methods support this option.


Changes to the API?
Yes. 2 new public methods are defined:

Object JSONObject.optQuery(String)
Object JSONArray.optQuery(String)

No, but new unit tests will be added after this commit. Those tests will fail unless this code is committed. See https://github.com/stleary/JSON-Java-unit-test/pull/48.

No, but this will be rolled into the next release, which is expected in May-June 2016 timeframe.

Yes. The JavaDoc should be updated.

PR #236, opened 2016-06-02
Make nextString throw a JSONException instead of a NumberFormatExcept…

madsager:
…ion for malformed input.

The documentation for the JSONObject(String) constructor says that JSONExceptions are thrown for invalid input. However, the implementation can throw a NumberFormatException as well. This fixes that.

### DIFF
--- a/JSONTokener.java
+++ b/JSONTokener.java
@@ -281,7 +281,7 @@ public class JSONTokener {
                     try {
                         sb.append((char)Integer.parseInt(this.next(4), 16));
                     } catch (NumberFormatException e) {
-                        throw this.syntaxError("Illegal escape.");
+                        throw this.syntaxError("Illegal escape.", e);
                     }
                     break;
                 case '"':
@@ -437,6 +437,16 @@ public class JSONTokener {
         return new JSONException(message + this.toString());
     }
 
+    /**
+     * Make a JSONException to signal a syntax error.
+     *
+     * @param message The error message.
+     * @param causedBy The throwable that caused the error.
+     * @return  A JSONException object, suitable for throwing
+     */
+    public JSONException syntaxError(String message, Throwable causedBy) {
+        return new JSONException(message + this.toString(), causedBy);
+    }
 
     /**
      * Make a printable string of this JSONTokener.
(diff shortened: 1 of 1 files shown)


johnjaylward:
I'd like to see the original NumberFormatException preserved. Could you update the ```syntaxError``` method to take an optional ```Throwable``` and pass that as the ```causedBy``` in the generated ```JSONException```?

madsager:
Good idea. Updated the pull request.

The error will now look something like this:

Exception in thread "main" org.json.JSONException: Illegal escape. at 9 [character 10 line 1]
	at org.json.JSONTokener.syntaxError(JSONTokener.java:448)
	at org.json.JSONTokener.nextString(JSONTokener.java:284)
	at org.json.JSONTokener.nextValue(JSONTokener.java:365)
	at org.json.JSONObject.<init>(JSONObject.java:206)
	at org.json.JSONObject.<init>(JSONObject.java:319)
	at fix.JSONFix.main(JSONFix.java:8)
Caused by: java.lang.NumberFormatException: For input string: "rl":"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at org.json.JSONTokener.nextString(JSONTokener.java:282)
	... 4 more

stleary:
Looks reasonable. Let me know if you are planning on updating the unit tests.

madsager:
I didn't see any tests in the repo, so it wasn't clear to me where to add it. I will be happy to add a regression test for this. Is that in a separate repo?

stleary:
Yeah, they are kept in a separate project: https://github.com/stleary/JSON-Java-unit-test

madsager:
Thanks. Pull request created for a simple regression test that check that the right type of exception is thrown.

stleary:
Thank you, will review this evening, if everything seems OK will open a pull request. We generally leave PRs open for about 3 days for comment before merging.

stleary:
JSONTokener.nextString() JavaDocs state that JSONExceptions are thrown, but NumberFormatExceptions may also be thrown.

**Changes to the API?**
No.

This change will be rolled into the next release, which is expected in June 2016 timeframe.

No, this brings the code in line with the documentation.

**Change to unit tests?**
A new unit test is added, which will only succeed with the code change.

PR #239, opened 2016-06-12
possible tweaks from intervision

roydmerkel:
Here are some possible tweaks I came up with to this library while writing java code for my organization.

1. Make the map implementation of the JSONObject a LinkedHashMap instead of HashMap, this allows me to enforce that the order I put the properties into the JSONObject is the order I will pull them out when writing the JSONObject to string (this was required due to organizational coding standards, and because of some really bad front end jquery code I had to work arround).
2. Allow JSONArrays to be be built from generic Collections (this was required when I needed to build a JSONArray for serialization from a dynamically build array, or possibly when I was trying to import an array of json objects containing json arrays).
3. Adding a JSONSerializable interface, this commit is optional (I put it here because it made sense to me to put it with the other json functions), but it made sense to include an interface that can be used to define a class that can be exported to a JSONObject to go with the JSONString class.

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -170,7 +170,11 @@ public class JSONArray implements Iterable<Object> {
      */
     public JSONArray(Object array) throws JSONException {
         this();
-        if (array.getClass().isArray()) {
+        if (array instanceof Collection) {
+            for (Object o: (Collection)array)){
+                this.put(JSONObject.wrap(o));
+            }
+        } else if (array.getClass().isArray()) {
             int length = Array.getLength(array);
             for (int i = 0; i < length; i += 1) {
                 this.put(JSONObject.wrap(Array.get(array, i)));
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -171,7 +171,7 @@ public class JSONArray implements Iterable<Object> {
     public JSONArray(Object array) throws JSONException {
         this();
         if (array instanceof Collection) {
-            for (Object o: (Collection)array)){
+            for (Object o: ((Collection)array)){
                 this.put(JSONObject.wrap(o));
             }
         } else if (array.getClass().isArray()) {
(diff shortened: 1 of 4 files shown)


johnjaylward:
I can't see this getting merged as-is. You'd be better off breaking each feature into it's own pull request.

* The JSONSerializer class doesn't appear to have any links into the code base, so I doubt that should be included
* The compile.sh should also not be included as it's not needed to compile and is not cross platform.
* The LinkedHashMap modification has been denied about 5-10 times already. Please see past pul requests for why or read the [FAQ](https://github.com/stleary/JSON-java/wiki/FAQ)
* The last portion of your changes with the Collections for JSONArray look interesting and has the most promise of getting accepted.

johnjaylward:
Although, for the JSONArray Collection change, can you explain why the ```JSONArray(Collection<?>)``` constructor did not work for you and why you needed to modify the ```JSONArray(Object)``` constructor instead?

stleary:
@roydmerkel thanks for the pull request. As @johnjaylward suggested, it cannot be accepted in its current form:
**JSONObject**
This change can't be accepted.  From the [FAQ](https://github.com/stleary/JSON-java/wiki/FAQ):
**Why isn't ordering allowed in JSONObjects?**
Support for ordering is one of the most popular requests, along with adding some type of build support. However, from the JSON spec RFC 7159: "An object is an unordered collection of zero or more name/value pairs". As a reference app, the design is required to stay as close to the spec as possible, so ordering in JSONObjects is not supported.

**JSONArray**
Should be a separate pull request and include explanation of why this change is needed. Probably unlikely to be accepted.

**JSONSerializable**
Sounds reasonable. Please open a separate pull request. Will need to use a different name - There is already a built-in Serializable interface which has a different meaning in Java.

Issue #54, opened 2012-04-17
Improve performance of toString() methods


vvalchev:
I've played a little bit with the code to improve the performance of toString() methods.

Here is a summary of most important changes:
- added write(Writer writer, int indentFactor, int indent) to both JSONObject and JSONArray, so now you can use indentation even with the write methods
- made toString() methods to use the write() method, but writing to a StringWriter. This also reduced the duplicate code! 
- appending to a StringBuffer/StringWriter is synchronized to the StringBuffer. This makes append operations faster.

I've wrote a simple test to measure the performance between changes and here is the results:
D:\develop\json\douglascrockford-JSON-java-d98dc0b>java -cp org.apache.commons.io.jar;. PerfTest
str: 5206
D:\develop\json\douglascrockford-JSON-java-d98dc0b>java -cp org.apache.commons.io.jar;. PerfTest
str: 5241
D:\develop\json\douglascrockford-JSON-java-d98dc0b>java -cp org.apache.commons.io.jar;. PerfTest
str: 5280
D:\develop\json\douglascrockford-JSON-java-d98dc0b>cd ..\new-json
D:\develop\json\new-json>java -cp org.apache.commons.io.jar;. PerfTest
str: 2912
D:\develop\json\new-json>java -cp org.apache.commons.io.jar;. PerfTest
str: 2946
D:\develop\json\new-json>java -cp org.apache.commons.io.jar;. PerfTest
str: 2912

This means that the average milliseconds for the old code is 5242 and 2923 for the new code. This means that the new code is ~1.8 times faster that the original code.

Here is a zip file (http://depositfiles.com/files/vnfv5budt) that includes:
- updated JSONObject and JSONArray code
- PerfTest.java
- screenshot of the output from PerfTest.java

I cannot include the JSON file I've been testing with, but it contains various data types, JSON objects and arrays. It is at about 74K big.

PR #241, opened 2016-06-15
Use Appendable in place of Writer

run2000:
This change modifies JSONWriter to take an Appendable instead of a Writer as its constructor parameter. Appendable is an interface implemented by Writer, as well as classes such as StringBuilder.

This refactor allows JSONStringer to use a StringBuilder directly, instead of a StringWriter.

It also flows through to writeValue() in JSONObject. This in turn allows valueToString() to be simplified, simply delegating to writeValue() with StringBuilder as a parameter.

I ran the test suite in JSON-Java-unit-test, which continues to pass. The code coverage varies in minor ways in JSONObject, JSONArray, and JSONWriter, due to the code reorganisation.


johnjaylward:
Overall this looks like a great refactor to me. I'd just like to ensure that all the branch points that were in ```valueToString``` are appropriately represented in ```writeValue``` since the refactor simplifies the logic there to remove duplicate code. I'm not sure I have time to go through that check myself though. 

Thanks for the great contribution.

johnjaylward:
On an initial inspection, it looks like ```writeValue``` is missing:
* The conversion of ```Map<?>``` to ```JSONObject```
* The conversion of ```Collection<?>``` to ```JSONArray```

```valueToString``` also has an explicit check for ```Boolean``` or ```JSONObject``` or ```JSONArray```, but the same logic should be covered by the final ```else``` clause in ```writeValue```

```writeValue``` does have an explicit check for ```Boolean```, but I'm thinking it could be removed in favor of dropping into the final ```else```.

Does that look right? If I missed any of them feel free to just update the pull request accordingly.

**Comment by johnjaylward** [comment]
Nevermind about the first part. I mistakenly looked at the diff instead of the full file. The conversion for ```Map<?>``` and ```Collection<?>``` are there.

I'm still in favor of removing the ```instanceof Boolean``` check though. It doesn't seem needed with the fallthrough ```else```.

**Comment by johnjaylward** [comment]
One last item. I'm not sure I'm a fan of this check in the ``instanceof JSONString``` branch:

```java
if (o != null) {
    writer.append(o);
} else {
    quote(value.toString(), writer);
}
```

If ```toJSONString()``` return null, then null is what should get written. We should not override the return value of that method for any reason.

run2000:
For the instanceof Boolean test, it's writing the value without going through the quote() method. This is so that boolean values are written as bare words, not quoted strings.

Regarding the instanceof JSONString, I was just changing the logic from a ternary operation so that I could call the two parameter form of quote(), passing in the Appendable directly, rather than going through a separate StringBuilder accumulator. You're quite correct that this is a logic change from what was in the stringToValue() method, which throws an exception instead.

johnjaylward:
Good point on the boolean check, I didn't think of that.

I can't decide if the exception being thrown on null is really a good idea either. I guess the original implementation expected "" to be written out for nulls, but then you can never have ```toJSONString``` pass back the null keyword. "", "null", and null are all distinct values in JSON and I think that toJSONString should probably support it. Although, I'm not sure if an instance of an object should ever serialize to null... Maybe there is some edge-case where that could/should happen in someone's business logic?

run2000:
It's a tough one to resolve satisfactorily. The Javadoc of JSONString doesn't say what should happen if null is returned from toJSONString(), so we're in to an area of programming by co-incidence.

If a JSONString object was used in a JSONObject or JSONArray object model, it would appear to work, though maybe unexpectedly, by falling back on its toString() method. If it was used in a JSONWriter or JSONStringer, it would fail with an exception.

Solutions look like:
* conservative behaviour -- always fall back on toString(), and preserve compatibility with any code that depended on this behaviour previously. It's doubtful whether any existing code depended on an exception being thrown here.
* remove bugs that could tickle someone in the future -- always throw an exception on null, forcing a decision on how to update any existing code. Defensible given that exceptions could already occur depending on how you used the API.
* silently insert a JSON null value -- probably what I would do if I was writing a JSON parser from scratch, without an existing code base, but could lead to silent unexpected failures in code that depends on existing behaviour.

Personally, I think the behaviour of (3) is too subtle a change to consider as a change for an existing API. Either (1) or (2) could be justified, but it depends on how conservative you want to be.

johnjaylward:
@stleary I think @run2000 made a decent list of the differences and problems each solution faces. I like the 2nd and 3rd options the best. I don't think we should ever fall back on toString() as that breaks the point of having the interface.

stleary:
Thanks for the thorough refactor and discussion, and appreciate that this does not break the unit tests. Regarding behavior changes, will need a little time to look at them. For example, it is possible that the unit test coverage was not adequate, and should have broken on a behavior change. All things being equal, strongly prefer conservative changes that do not affect existing code.

stleary:
Any concerns about the file mode changes?

johnjaylward:
No, it is removing the execute flag.

stleary:
And 	the 826ceef commit...

johnjaylward:
Looking at the combined changes, the permissions are not changed.

stleary:
OK, I think I understand what is going on. Most of the changes were committed around June 18 2016. At the time, I committed to reviewing the unit tests, since even though they passed, they may have missed something. That was about the time I became heavily committed at work, and never got around to the unit tests. In the meanwhile, 826ceef was committed on July 1 2016, which in addition to the accidental mode change to 100755, appears to include functional changes to JSONArray.join(), JSONObject.valueToString(), JSONObject.writeValue(), and JSONWriter.value().  The mode change was corrected in 2917cc5, also on July 1 2016.

Someone, hopefully @johnjaylward, needs to look at these changes and provide an approval or comments.  Someone else (myself) needs to run the unit tests to make sure nothing is broken, and then make sure the code coverage of the changed code is acceptable, and that there are no missing tests that should be added. 

So no clock reset, because as it turns out, I never started the merge countdown. Please let me know if I am missing or misunderstanding something here.

johnjaylward:
I took a quick look at the latest commits, but didn't fully review them to see why the changes were made. I probably won't have time to look at them in depth until Wednesday

run2000:
Apologies, I was just commiting some outstanding code I'd chaned since sending the merge request, not realising that it would affect the merge.

The chane is adding an overload to writeValue () to take two parameters, since there two callers that took 0, 0 as the third and fourth parameters.

Sorry for the noise.

stleary:
@run2000 No worries, this is a good exercise to make sure the changes are fully understood and work as intended.

johnjaylward:
Sorry, I finally got around to reviewing the latest changes and they all look good. The main changes were to remove references to the `valueToString` method that was refactored to just call `writeValue`. This should reduce stack overhead with 1 less function call.

The last change of significance is in `JSONWriter` with the replacement in the `value(double)` function. This change is also to reduce stack overhead by shortcutting to the actual call further in the stack. No functional change is here.

The last thing to decide on before approval of this PR is how we want to handle the exception case for writing out `JSONString` implementations that return `null`. (https://github.com/stleary/JSON-java/pull/241#issuecomment-226360247)

run2000:
Is there anything additional I need to do to help this along? Or are you comfortable shepparding this along now? I think my comments from Jun 16 still stand.

stleary:
@run2000, there seems to be at least one item left from @johnjaylward's most recent comment, regarding the exception case for JSONString implementations that return null. 

Otherwise, I still need to do a final review of the changes, and confirm the unit test coverage is good. If you want, you could look at the unit tests to confirm all new code has coverage and that the unit tests are "reasonable". That is, that they adequately test the functionality, ensure there are no unexpected changes to existing functionality, and there are not any uncaught bugs. If so, please post your results here.

run2000:
I've been looking at test cases for `JSONObject` and `JSONArray` today. My focus so far has been covering off the `toString()` and `write()` methods, including the overloads for non-zero indentation. I've also created some tests for the new `toMap()` and `toList()` methods.

When I next find a few hours, I'll have a closer look at where `JSONString` is called, and how the calling code deals with nulls and exceptions.

These are in `run2000/JSON-Java-unit-test` if you're interested.

stleary:
Thanks, tests, and progress, looks good.

stleary:
@run2000 
* Looks like the merge of #249 has introduced some conflicts. Can you resolve them and resubmit? 
* Does https://github.com/stleary/JSON-Java-unit-test/pull/53 include all of the tests you are planning for this pull request?
* Do you feel you have addressed @johnjaylward's concern about the exception case for JSONString implementations that return null?

run2000:
Hi @stleary 

* Resolved the conflict. It was a line of Javadoc in JSONArray
* Yes, I'm happy with the tests I've added to stleary/JSON-Java-unit-test#53
* In the case of JSONString returning `null`, I've gone with the conservative fix as outlined in https://github.com/stleary/JSON-java/pull/241#issuecomment-226360247 . This is the most backward-compatible fix short of adding new code to valueToString() to explicitly throw new JSONException in this case. My temptation is to prefer consistent behaviour between valueToString() and writeValue().

Let me know if there's any other concerns.

johnjaylward:
I personally don't like falling back on the toString behavior. If the implemetor of the API specifies the JSONString interface on one of their objects, then calling toString does not seem like the best option when null is returned. That feels more like an easter egg than feature.

In either case, @run2000, can you update the javadoc on JSONString to indicate what happens when null is returned. I don't believe I saw that in your commits.

run2000:
Ok, will do. Since I'm in that area, I notice JSONString doesn't have the json.org copyright notice. Is this an oversight or intentional?

johnjaylward:
it would be an oversight. If you could add that too that would be best.

run2000:
Okay, Javadoc has been expanded on to document the `null` return scenario.

stleary:
I am not strongly in favor of this change. However, if the preparation is done to spec, the commit can proceed.
- Need a clear justification for why the new code is an improvement over the old
- Format changes should not be included in the commit, they just make it harder to compare versions
- Any time the functionality is changed, e.g. JSONArray line 1447 removing synchronized, a comment should be added explaining why this is was necessary or an improvement.

JSONArray.java
```diff
@@ -447,7 +445,7 @@ public String join(String separator) throws JSONException {
             if (i > 0) {
                 sb.append(separator);
             }
-            sb.append(JSONObject.valueToString(this.myArrayList.get(i)));
+            JSONObject.writeValue(sb, this.myArrayList.get(i));
```
run2000: Write directly to the one StringBuilder, avoiding creation of temporary buffer.

JSONArray.java
```diff
@@ -1159,10 +1157,8 @@ public String toString() {
      * @throws JSONException
      */
     public String toString(int indentFactor) throws JSONException {
-        StringWriter sw = new StringWriter();
-        synchronized (sw.getBuffer()) {
```
run2000: This reverts a change related to https://github.com/stleary/JSON-java/issues/54

Basically, now that we use StringBuilder, which is not synchronizing on a buffer, instead of StringWriter, which does, the rationale for synchronizing no longer applies.

JSONObject.java
```diff
@@ -1465,51 +1461,84 @@ public static Writer quote(String string, Writer w) throws IOException {
...
+        w.append('"');
         for (i = 0; i < len; i += 1) {
             b = c;
             c = string.charAt(i);
             switch (c) {
             case '\\':
             case '"':
-                w.write('\\');
-                w.write(c);
+                if(prev < i) {
```
run2000: Appending sequence ranges here:

* If the target Appendable is unbuffered, a sequence will write more quickly than single characters.
* If the target Appendable is a StringBuffer or StringBuilder, writing sequences will help reduce buffer resizes

In any circumstance, writing a single character at a time, with a method call per character, is likely to be slower than gathering sequences where possible.

JSONObject.java
```diff
@@ -1465,51 +1461,84 @@ public static Writer quote(String string, Writer w) throws IOException {
...
-                    w.write("\\u");
+                    if(prev < i) {
+                        w.append(string, prev, i);
+                    }
                     hhhh = Integer.toHexString(c);
-                    w.write("0000", 0, 4 - hhhh.length());
-                    w.write(hhhh);
-                } else {
-                    w.write(c);
+                    w.append("\\u0000", 0, 6 - hhhh.length());
```
run2000: Combine the output of "\\u" and the number padding of "0000", since the substring calculation is sufficiently easy.

JSONObject.java
```diff
@@ -1465,51 +1461,84 @@ public static Writer quote(String string, Writer w) throws IOException {
...
+                    w.append(hhhh);
+                    prev = i + 1;
                 }
             }
         }
-        w.write('"');
+        if(prev == 0) {
+            w.append(string);
```
run2000: For the degenerate case of writing sequences that require no escaping, writing the entire CharSequence to the Appendable means the fewest temporary copies. Situations where this could be important include base-64 encoded binary data.

JSONObject.java
```diff
@@ -1699,10 +1728,8 @@ public String toString() {
      *             If the object contains an invalid number.
      */
     public String toString(int indentFactor) throws JSONException {
-        StringWriter w = new StringWriter();
-        synchronized (w.getBuffer()) {
-            return this.write(w, indentFactor, 0).toString();
-        }
+        StringBuilder w = new StringBuilder();
```
run2000: Same change related to https://github.com/stleary/JSON-java/issues/54 as above. Using StringBuilder no longer requires the synchronization.

JSONObject.java
```diff
@@ -1828,14 +1822,23 @@ public static Object wrap(Object object) {
      * @return The writer.
      * @throws JSONException
      */
-    public Writer write(Writer writer) throws JSONException {
+    public <T extends Appendable> T write(T writer) throws JSONException {
         return this.write(writer, 0, 0);
     }
 
-    static final Writer writeValue(Writer writer, Object value,
+    static <T extends Appendable> T writeValue(T writer, Object value)
```
run2000: New method used by JSONObject.valueToString(), JSONArray.join(), and JSONWriter.appendValue(). Simple delegation.

JSONObject.java
```diff
@@ -1849,26 +1852,30 @@ static final Writer writeValue(Writer writer, Object value,
...
             try {
                 o = ((JSONString) value).toJSONString();
             } catch (Exception e) {
                 throw new JSONException(e);
             }
-            writer.write(o != null ? o.toString() : quote(value.toString()));
+            if (o != null) {
+                writer.append(o);
+            } else {
+                quote(value.toString(), writer);
```
run2000: This version of quote() writes directly to the Appendable, avoids another buffer copy.

JSONWriter.java
```diff
@@ -50,11 +49,11 @@ of this software and associated documentation files (the "Software"), to deal
  * <p>
  * The first method called must be <code>array</code> or <code>object</code>.
  * There are no methods for adding commas or colons. JSONWriter adds them for
- * you. Objects and arrays can be nested up to 20 levels deep.
+ * you. Objects and arrays can be nested up to 200 levels deep.
```
run2000: Javadoc out-of-date, or typo.

JSONWriter.java
```diff
@@ -102,7 +101,7 @@ public JSONWriter(Writer w) {
     }
 
     /**
-     * Append a value.
+     * Append a JSON-encoded value.
```
run2000: Javadoc tweak, to distinguish from the method added below.

JSONWriter.java
```diff
@@ -130,6 +129,35 @@ private JSONWriter append(String string) throws JSONException {
     }
 
     /**
+     * Append a value, converting it into a JSON string.
+     *
```
run2000: Similar to append() method above, writes directly using JSONObject.writeValue() to encode the value.

JSONWriter.java
```diff
@@ -300,7 +328,7 @@ public JSONWriter value(boolean b) throws JSONException {
      * @throws JSONException If the number is not finite.
      */
     public JSONWriter value(double d) throws JSONException {
-        return this.value(new Double(d));
+        return this.append(JSONObject.numberToString(Double.valueOf(d)));
```
run2000: Avoid the delegation chain to JSONWriter.value(Object), to JSONObject.writeValue(Object), to JSONObject.numberToString(). Avoid the chain and go there directly. Still retains the testValidity() call.

run2000:
* Changes methods that take a `java.io.Writer` to use a
  `java.lang.Appendable` instead. Situations where `Appendable`
  can be used where `Writer` cannot include the `StringBuffer`,
  `StringBuilder`, and `CharBuffer` classes.
* `JSONObject.valueToString()` simplified to delegate to
  `JSONObject.writeValue()`. This causes a small change in
  behaviour where the value implements `JSONString` and the
  `toJSONString()` method returns `null`. Now the behaviour is
  identical to the `writeValue()` behaviour.
* Some work has also been done to reduce buffer copies, where
  intermediate `StringBuilder` objects are used to create `String`
  objects that get written to a `Writer`. Write directly to the `Writer`
  instead.

**Changes to the API?**
Some method parameter types have changed from `Writer` to `Appendable` in `JSONObject`, `JSONArray`, and `JSONWriter`.

Can be rolled into the next release.

May want to document the `JSONObject.valueToString()` behaviour change noted above.

**Change to unit tests?**
New unit tests are added, calling out the `valueToString()` 
`writeValue()` methods and the small differences. See 
 https://github.com/stleary/JSON-Java-unit-test/pull/53

stleary:
The change from Writer to Appendable looks OK. Can't accept the change to JSONObject.valueToString() at the present time. Behavior changes have a high bar for acceptance, which this does not meet. Or the changes to buffer copies, since there is no compelling case for changing the code. If still interested, please resubmit just the Appendable changes. Also, going forward, recommend that code comments actually be in the code, not the pull request.

erosb:
> Also, going forward, recommend that code comments actually be in the code, not the pull request.

+1 for this

Issue #240, opened 2016-06-14
JSONObject: tabs should not be allowed


mmoghadas:
If I'm understanding the JSON standards correctly, tabs are not allowed.  I'm posting a json to my api (the whitespace at the end of the value is a tab character).  This is what's happening in the background:

String string = {"key":"value	"};
JSONObject json;
try {
            json = new JSONObject(string);
        } catch (JSONException e) {
}

the assignment above should not succeed, but it does. 
you can use this sample project to see for yourself
https://github.com/mmoghadas/gs-rest-service/blob/master/complete/src/main/java/hello/GreetingController.java#L26

git clone git@github.com:mmoghadas/gs-rest-service.git
cd gs-rest-service/complete
./gradlew build
java -jar build/libs/gs-rest-service-0.1.0.jar

http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf
Insignificant whitespace is allowed before or after any token. The whitespace characters are: character
tabulation (U+0009), line feed (U+000A), carriage return (U+000D), and space (U+0020). Whitespace is not allowed within any token, except that space is allowed in strings.

johnjaylward:
Page 5, Figure 5 of EMCA-404 indicates that a String can contain an valid code point except " \ or a control character. I  believe that most systems do not consider \t tab to be a control character

stleary:
For what it's worth, http://JSONLint.com seems to agree that tabs should not appear in strings.

johnjaylward:
well, I'll say at best the document is inconsistent. I'd think this would be a possible breaking change for existing users though.

stleary:
Neither spec is especially well-written, and this particular item is already prone to confusion because a tab can either be represented as an escape sequence or as the character itself. My reading is that tabs are allowed in strings only when they are represented by the char sequence "\t". However, I am not inclined to fix this since the bug seems unlikely to likely to break applications, but the fix might. Other opinions?

johnjaylward:
I'm OK with that. Maybe just place an Errata in the README?

johnjaylward:
Also, [RFC 7159 page 8](https://tools.ietf.org/html/rfc7159#page-8) specifies a string to be the same as the flowchart#5 indicates on page 5 of EMCA-404.

So I'm OK with leaving it as it is.

johnjaylward:
gah, nevermind again. I see this in RFC7159:

   All Unicode characters may be placed within the
   quotation marks, except for the characters that must be escaped:
[...]

PR #242, opened 2016-06-16
Fixes Issue #240

johnjaylward:
Update Readme to note that the parser allows some invalid JSON text while the generator never should.

### DIFF
--- a/README
+++ b/README
@@ -80,6 +80,14 @@ in the form of get(), opt(), and put() API methods.
 Although 1.6 compatibility is currently supported, it is not a project goal and may be
 removed in some future release.
 
+In compliance with RFC7169 page 10 section 9, the parser is more lax with what is valid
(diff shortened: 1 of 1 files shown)


Parakoopa:
It should be RFC7159, not RFC7169

johnjaylward:
thanks fixed

stleary:
This is a change to the README documentation. Normally it would just be committed, but in this case it will be left up for review for a few days. The project diverges from the specs in that when parsing strings, control chars other than '\n', '\r', and binary 0 are allowed. The team believes that the benefit of following the spec exactly does not outweigh the potential cost of breaking existing applications.

**Changes to the API?**
No.

This change will be rolled into the next release, which is expected in June 2016 timeframe.

Yes, this is the doc change (README only).

**Change to unit tests?**
A new unit test is added, which documents the existing behavior. See stleary/JSON-Java-unit-test#50

PR #244, opened 2016-06-21
Update JSONArray.java

jorgesartori:
Modified to consider similar arrays with the same elements but in different orders.

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -1026,24 +1026,32 @@ public class JSONArray implements Iterable<Object> {
             return false;
         }
         int len = this.length();
+        boolean[] verify = new boolean[len];
         if (len != ((JSONArray)other).length()) {
             return false;
         }
         for (int i = 0; i < len; i += 1) {
-            Object valueThis = this.get(i);
-            Object valueOther = ((JSONArray)other).get(i);
-            if (valueThis instanceof JSONObject) {
-                if (!((JSONObject)valueThis).similar(valueOther)) {
-                    return false;
-                }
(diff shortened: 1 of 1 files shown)


johnjaylward:
What is this supposed to be doing?

jorgesartori:
Before, if two arrays contained the same elements, but not in the same order, this method returned false. 
Example:
array1 = [{"id":2,"color":"white"},{"id":5,"color":"blue"}]
array2 = [{"color":"blue","id":5},{"color":"white","id":2}]

This arrays must be considered similar, but in the original code does not, because compares array1[1] and array2[1], and after, array1[2] and array2[2].
On the modification proposed, for each array1 position are checked all array2 positions.

johnjaylward:
No, they are not similar because both ECMA-404 and RFC 7159 define an array as being ordered. The javadoc also states that the 2 lists must have similar sequences.

[1,2,3] is not the same sequence as [3,2,1]

stleary:
@jorgesartori  Thanks for your pull request. It cannot be accepted due to the reasons stated. If you decide to keep it for your own use, make sure it passes this test case:
````
    @Test
    public void notReallySimilar() {
        String s1 = "[\"red\", \"red\"]";
        String s2 = "[\"red\", \"green\"]";
        JSONArray ja1 = new JSONArray(s1);
        JSONArray ja2 = new JSONArray(s2);
        assertFalse(ja1.similar(ja2));
    }
````

jorgesartori:
Sorry and thanks!

PR #246, opened 2016-06-22
Add License file

johnjaylward:
I just copied the license text from a source file into a general LICENSE file. It seems like we get asked about the license somewhat frequently, this should help reduce confusion from lazy people.


stleary:
This is an exact copy of the copyright comment found in the JSON-Java source files. If there are no questions or concerns, it will be merged to the build and included in the next release (now expected to be in July 2016).

Puy2308 (approved):

PR #249, opened 2016-07-08
reduces the use of unnecessary exceptions

Simulant87:
Creating an Exception is an expensive operation, because of the collection of the stacktrace.

If the Exception is caught and ignored, this is even an unnecessary operation. Therefore I added null checks und class instance checks to prevent ClassCastExceptions and NullPointerExceptions from being created and caught afterwards.

The changes are passing all test from the test suite as the behavior when a JSONException is thrown or the defaultValue is returned has not been changed.

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -72,13 +72,13 @@ import java.util.Map;
  * <li>Strings do not need to be quoted at all if they do not begin with a quote
  * or single quote, and if they do not contain leading or trailing spaces, and
  * if they do not contain any of these characters:
- * <code>{ } [ ] / \ : , #</code> and if they do not look like numbers and
- * if they are not the reserved words <code>true</code>, <code>false</code>, or
+ * <code>{ } [ ] / \ : , #</code> and if they do not look like numbers and if
+ * they are not the reserved words <code>true</code>, <code>false</code>, or
  * <code>null</code>.</li>
  * </ul>
  *
  * @author JSON.org
- * @version 2016-05-20
+ * @version 2016-07-08
  */
 public class JSONArray implements Iterable<Object> {
 
@@ -156,9 +156,9 @@ public class JSONArray implements Iterable<Object> {
     public JSONArray(Collection<?> collection) {
         this.myArrayList = new ArrayList<Object>();
         if (collection != null) {
-        	for (Object o: collection){
-        		this.myArrayList.add(JSONObject.wrap(o));
-        	}
+            for (Object o : collection) {
+                this.myArrayList.add(JSONObject.wrap(o));
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


johnjaylward:
I don't have any major issue with this, however, please verify that this test case returns the same value before and after your changes:

```java
JSONObject jo = new JSONObject("{\"int\":\"123\",\"true\":\"true\",\"false\":\"false\"}");
assert jo.optBoolean("true",false)==true : "unexpected optBoolean value";
assert jo.optBoolean("false",true)==false : "unexpected optBoolean value";
assert jo.optInt("int",0)==123 : "unexpected optInt value";
assert jo.optLong("int",0)==123 : "unexpected optLong value";
assert jo.optDouble("int",0.0)==123.0 : "unexpected optDouble value";
assert jo.optBigInteger("int",BigInteger.ZERO).compareTo(new BigInteger("123"))==0 : "unexpected optBigInteger value";
assert jo.optBigDecimal("int",BigDecimal.ZERO).compareTo(new BigDecimal("123"))==0 : "unexpected optBigDecimal value";

JSONArray ja = new JSONArray("[\"123\",\"true\",\"false\"]");
assert jo.optBoolean(1,false)==true : "unexpected optBoolean value";
assert jo.optBoolean(2,true)==false : "unexpected optBoolean value";
assert jo.optInt(0,0)==123 : "unexpected optInt value";
assert jo.optLong(0,0)==123 : "unexpected optLong value";
assert jo.optDouble(0,0.0)==123.0 : "unexpected optDouble value";
assert jo.optBigInteger(0,BigInteger.ZERO).compareTo(new BigInteger("123"))==0 : "unexpected optBigInteger value";
assert jo.optBigDecimal(0,BigDecimal.ZERO).compareTo(new BigDecimal("123"))==0 : "unexpected optBigDecimal value";

```

johnjaylward:
I added the test cases here:
https://github.com/stleary/JSON-Java-unit-test/pull/51

from what I can tell your PR should pass them both before and after, but I did not run them on your PR

Simulant87:
I run the tests before and after my changes against the updated test repository including your new test case and they both passes the test successfully.

johnjaylward:
@stleary This PR looks good to me. I'm OK with including this in the next release if you are.

stleary:
The changes are probably OK. However, it should not be necessary to wade through pages of formatting in order to find the real code changes. It is actually helpful to convert tabs to 4-spaces, so that is OK. And if you find egregious formatting problems, those are OK to fix as well (don't think I saw any of those in this review). Otherwise only lines containing actual code changes should be touched. Please resubmit without the format changes. If someone else wants to do this task, that would be fine as well.

Simulant87:
I reverted my last commit and created the changes to reduce the exceptions again (they are the same changes like in the first commit) without the changes to the formatting. There are just a hand full of "important" formating changes in this commit.
I repeated the tests and again they are all passing.

stleary:
Thanks, this is helpful and simplifies the review.

stleary:
Minor refactoring to more efficiently handle intermediate exceptions. Instead of a throw/catch, 

**Changes to the API?**
No change to the API

This change will be rolled into the next release, which is expected in July or August 2016 timeframe.

Not required

**Change to unit tests?**
See stleary/JSON-Java-unit-test#51

Issue #243, opened 2016-06-20
Bug in version 20160212?


sunmingtao:
`XML.toJSONObject("<code>29.</code>"); `returns `{"code": 29}`, rather than `{"code": "29."}`

Earlier versions seem fine.

stleary:
Will investigate why the behavior has changed, but this looks correct. If a value can be converted to numeric, the conversion should be performed.

johnjaylward:
@stleary, it changed back in a commit I made for the number parsing. I deprecated the `XML.stringToValue` method and had it call `JSONObject.stringToValue` instead for consistency. PR #188 was the one.

@sunmingtao, there are a lot of complaints with how we handle the XML to JSON conversions. I doubt this will get fixed short of modifying the XML reader to always use string values, but I'm sure someone will complain about that as well.

stleary:
@johnjaylward Thanks for clarifying. So now we know why it changed, and it seems to be the right thing to do.

We know that JSON/XML conversions will always be broken to some degree, because of the nature of the format: data vs document. With that in mind, it may be worth considering allowing some degree of tuning by the developer. For example, the default would be the code as currently implemented, but a user may choose to emit values as strings, instead of  conversion to numeric or other representations.  Other settings could be considered as well. Please post here if you have any thoughts or suggestions.

johnjaylward:
@stleary there was a simliar PR for the JSONML class, #97.  The implementation looks decent enough there and may be applicable to handling this scenario

stleary:
I think this issue is resolved.

Issue #230, opened 2016-05-10
Strings that are valid numbers in scientific notation get transformed when converting XML into JSON


oleg-gr:
If you have a string in your XML, that also happens to be a valid number in scientific notation, it will be transformed into a java representation of that number when XML is transformed into JSONObject:

```
XML.toJSONObject("<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?>\n" +
        "<str1>045e5000</str1>\n" +
        "<str2>45e5000</str2>\n" +
        "<str3>045e50</str3>\n" +
        "<str4>45e5</str4>");
```
will return
```
{"str1":"045e5000","str4":4500000,"str3":4.5E51,"str2":"45e5000"}
```

I understand that vanilla XML doesn't have a way to define types, but I still think this is a very unpredictable behavior. I am not sure how to solve it. Possibly, do not convert e notations and leave them as strings?

stleary:
@oleg-gr, Agreed, but I think you would get the same output with this code:
````
        JSONObject jsonObject  = new JSONObject("\"str1\":045e5000,\"str2\":45e5000,\"str3\":045e50,\"str4\":45e5}");

````
In the case of your example, I think the numbers with exponent 5000 could not be converted to double, so were handled as strings. This seems reasonable to me.

**Comment by oleg-gr** [comment]
@stleary, thank you for the reply.

Yeah, I know that the numbers with huge exponent weren't converted because they were too large, but in fact, in my opinion, it's a more expected behaviour. I am more concerned with other values being converted to Double instead of String since the original value gets mangled completely in some cases ("45e5" turns into 4500000 instead of staying "45e5"). Since as I was saying, in my case, those were intended to be strings (e.g. they are IDs or something else that just happens to have 'e' in them), I was surprised at this behaviour, but I see where it's coming from.

Since xml doesn't differentiate types, how do we deal with it? Why is it assumed that those should be tried to be converted as numbers getting the result like this:
```
new org.json.JSONObject("{\"str1\":045e5000,\"str2\":45e5000,\"str3\":045e50,\"str4\":45e5}");
```
and not the result like this:
```
new org.json.JSONObject("{\"str1\":\"045e5000\",\"str2\":\"45e5000\",\"str3\":\"045e50\",\"str4\":\"45e5\"}");
```
?

[...]

Issue #194, opened 2016-02-05
Implement Independent Serialization Policies for Primitive Data Types when converting XML to JSON


Twigglius:
[This is an enhancement request, not a true issue]

When converting XML to JSON the code does its best to determine proper serialization/deserialization but it is sometimes incorrect. I should have the ability to tell the program to treat primitive data types according to my requirements. 

For example I have an XML message I want to convert to JSON, but I want every field to written as a String. I have an element in my XML with a value of '5237E8' after conversion, the JSON shows the value as 5.237E11. It assumed it was a double and converted it. 

Another example is I have elements with leading 0's that need to be maintained in the conversion to json, but those 0's are removed in the serialization/deserialization.

This is standard functionality in other similar libraries such as GSON.

johnjaylward:
This project does not support schema validation. While this feature is often requested for the XML processing, it would be a lot of work to fit it in.

Also, looking at the [JSON number spec](http://json.org), your example matches a number. If you can control the XML generation at all, I would suggest modifying the output of 5237E8 to have quotes around it, or other non-numeric characters.

For the leading 0 problem, I agree that the JSON spec does not allow for them (except for leading a decimal like 0.5), so those values should be treated as string data. If you want, you can open a issue for that case specifically.

mikesterlingw:
I'm working on a pull request to address this. I think what @Twigglius is trying to get at is not schema validation but a way to treat all numbers as strings.

stleary:
@Twigglius What option does GSON provide that allows you to work around the problem?

johnjaylward:
@mikesterlingw any movement on the pull request for this?

stleary:
Feel free to submit a pull request if you come up with an implementation that meets your requirements while not breaking existing applications.

Issue #69, opened 2012-11-14
Converting strings to other types can change values


yaccob:
Hi Douglas

The attempt to guess the type of a text node or an attribute value may change the type which may be unintended. If the intended type is a string, "01" is not equal with "1". So operations change values which can cause unwanted effects. 
Furthermore the behavior is not consistent: "00" remains unchanged (remains a String) while "01" is converted to the number 1 as you can see in the sample below.
Also automatic conversion to boolean might have unwanted effects. Imagine the title of a book's chapter is "True" - intentionally with upper case "T". As far as I could see currently I am not able to convert this title to Json without implicitly changing it to a boolean and therefore losing information. 

Sample: 

xml:
&lt;root&gt;
 &nbsp;&nbsp;&lt;id&gt;01&lt;/id&gt;
 &nbsp;&nbsp;&lt;id&gt;1&lt;/id&gt;
 &nbsp;&nbsp;&lt;id&gt;00&lt;/id&gt;
 &nbsp;&nbsp;&lt;id&gt;0&lt;/id&gt;
 &nbsp;&nbsp;&lt;title&gt;True&lt;/title&gt;
&lt;/root&gt;

code:
println org.json.JSONML.toJSONArray(xml)

output:
["root",["id",1],["id",1],["id","00"],["id",0],["title",true]]

I would suggest to either change this or at least to add an option to control the behavior in a static way. (e.g. adding a non-final static boolean field KEEP_STRINGS which can be set to true)

Best regards,
yaccob

PR #253, opened 2016-07-18
Optional type conversion for XML reading

johnjaylward:
When reading XML and JSONML offer new option to not convert strings. This was mostly from PR #70 and is in reference for many issues brought up about conversions that are unwanted during XML processing, the lastest of which is #243 

It expands on pr #70 by bringing the conversion option directly into the XML class and not just in JSONML. It also prefers polymorphism instead of new function names.

Test cases from the original PR will be included in a separate commit in the test project.

### DIFF
--- a/JSONML.java
+++ b/JSONML.java
@@ -36,24 +36,6 @@ import java.util.Iterator;
  * @version 2016-01-30
  */
 public class JSONML {
-
-    /**
-     * Parse XML values and store them in a JSONArray.
-     * @param x       The XMLTokener containing the source string.
-     * @param arrayForm true if array form, false if object form.
-     * @param ja      The JSONArray that is containing the current tag or null
-     *     if we are at the outermost level.
-     * @return A JSONArray if the value is the outermost tag, otherwise null.
-     * @throws JSONException
-     */
-    private static Object parse(
-        XMLTokener x,
-        boolean    arrayForm,
-        JSONArray  ja
-    ) throws JSONException {
-		return parse(x, arrayForm, ja, false);
-	}
-
     /**
      * Parse XML values and store them in a JSONArray.
      * @param x       The XMLTokener containing the source string.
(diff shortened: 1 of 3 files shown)


stleary:
No problem with the XML changes, but concerned about opening the API to JSONML, JSONObject, and JSONArray to ignore type conversions. Is there any way to do this without changing the public API for those classes?

johnjaylward:
JSONObject and JSONArray should be unaffected. JSONML is another XML representation and should be be consistent with the standard XML class I think. I'm personally unsure why JSONML is not a subclass of the XML class. On the surface, there appears to be a lot of duplicate code between the 2 but I didn't take the time to go through them and reduce duplication.

stleary:
Thanks for clarifying about JSONObject and JSONArray, agreed. I believe we have a certain freedom to work with XML, primarily because it was a candidate for deletion from the library. However, the same may not hold true for JSONML. Don't mind if they diverge a little, but unless there is a compelling argument otherwise, would prefer to avoid changes to the API that are not bug-related.

johnjaylward:
Any existing code should still work as expected with JSONML. The caller would have to specify the new parameter with a `true` value to get the new parsing. I think the overall benefits of keeping the XML and the JSONML classes with similar APIs and features outweighs not implementing it. At the time of the original PR there also seems to be a decent amount of support for it.

johnjaylward:
Just for reference, here are the other issues I found that reference XML improperly converting values that this PR may be a valid fix:
* #230
* #194
* #69
* #49
* #27

stleary:
Thanks for doing the research, will need a little time to review those PRs.

stleary:
A number of developers have requested a little more control over XML parsing. This gives the ability to specify parsing of XML values to strings instead of other JSON types.

**Changes to the API?**
Yes, the XML and JSONML APIs are extended to allow the developer to specify string parsing. The default behavior is still first attempt to parse to JSON types.

This change will be rolled into the next release, which is expected in July or August 2016 timeframe.

Yes, since the API is changing, the new behavior should be called out to users. Some additional text in the README would be appropriate.

**Change to unit tests?**
New unit tests are added, which will only succeed with the code change. See stleary/JSON-Java-unit-test#52

stleary:
Adding this to bump the comments. This PR will be kept open for 7 days, in case XML or JSONML users want to comment.

PR #256, opened 2016-08-05
Preserving XML ordem when parse from XML string to Json

PedroHReis22:
Currently, if we parse a xml string to a JsonObject, we have no guarantee that the xml order will be preserved (because JSONObject uses a HashMap).
 May be interesting keep this order (although it does not make a difference in the Json's structure).
 This pull request swaps from HashMap to LinkedHashMap.


PedroHReis22:
It's duplicate from #106

PR #259, opened 2016-08-08
JSONWriter uses Appendable (v2)

run2000:
* Changes methods that take a `java.io.Writer` to use a
  `java.lang.Appendable` instead. Situations where `Appendable`
  can be used where `Writer` cannot include the `StringBuffer`,
  `StringBuilder`, and `CharBuffer` classes.

**Risks**
Low risk. This is modifying an API method to take an interface instead of a base class, which is generally a good practice. The positive unit test coverage is good. But JSONWriter is lacking coverage for most of the thrown exceptions, including in modified methods. OK to commit, but better unit test code coverage will be required before this code can be included in a release.  See https://github.com/stleary/JSON-Java-unit-test/issues/56. 

**Changes to the API?**
The constructor for JSONWriter has changed from `Writer` to `Appendable`.

Can be rolled into the next release, pending completion of the unit tests.

Not required.

No, the current tests run successfully. However, new unit tests have been added. See 
https://github.com/stleary/JSON-Java-unit-test/pull/53

### DIFF
--- a/JSONWriter.java
+++ b/JSONWriter.java
@@ -1,7 +1,6 @@
 package org.json;
 
 import java.io.IOException;
-import java.io.Writer;
 
 /*
 Copyright (c) 2006 JSON.org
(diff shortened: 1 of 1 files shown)


run2000:
This is a much more straightforward change. The tests from the previous test cases apply.

johnjaylward:
This looks simple enough.

Your other pull request did have other good changes in it as well. Maybe if you find some good places to split the functional changes into separate pull requests it will be easier to get them approved.

I would especially like to see consistency in how the JSONString interface is used within the library, so if you could open a pull request (or at least an issue) for that, I would appreciate it. It should note where the inconsistencies happen and the current way the interface is used at each point. We can then discuss the appropriate way that the interface should be handled from there.

stleary:
Looks good, will leave the pull request open for a few days for comments.

Issue #260, opened 2016-08-09
optBoolean isn't optional


quidryan:
[JSONObject::optBoolean](https://github.com/stleary/JSON-java/blob/master/JSONObject.java#L944) no longer catches the exception thrown by [JSONObject::get](https://github.com/stleary/JSON-java/blob/master/JSONObject.java#L472).  This means that while it's called "opt" (optional), it throws an exception if not found. It's broken in a few recent versions of org.json.

stleary:
Thanks for catching this! Will fix it soonest.

johnjaylward:
@stleary it looks like all the opt functions have this issue, or at least most. They should be changed from
```java
Object object = this.get(key); 
```
to
```java
Object object = this.opt(key);
```
to prevent the missing key exception.

stleary:
Agreed, another code change may fix it. But the unit tests did not catch it, so they will need to be tightened up too. If this simple bug is in the commit, others may have been included as well. There was no strong case for this pull request, it was supposed to be minor refactoring.  On the plus side, it looks like it can be reverted without affecting your latest commit.

quidryan:
With the better tests proving that this is fixed, we can maybe fix forward.

stleary:
Addressing this in #261, please post comments there going forward.

johnjaylward:
I'm updating the test cases now to take this into account

PR #261, opened 2016-08-09
Revert "reduces the use of unnecessary exceptions"

stleary:
Reverts stleary/JSON-java#249

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -78,7 +78,7 @@ import java.util.Map;
  * </ul>
  *
  * @author JSON.org
- * @version 2016-07-19
+ * @version 2016-05-20
  */
 public class JSONArray implements Iterable<Object> {
 
@@ -156,9 +156,9 @@ public class JSONArray implements Iterable<Object> {
     public JSONArray(Collection<?> collection) {
         this.myArrayList = new ArrayList<Object>();
         if (collection != null) {
-            for (Object o : collection) {
-                this.myArrayList.add(JSONObject.wrap(o));
-            }
+        	for (Object o: collection){
+        		this.myArrayList.add(JSONObject.wrap(o));
+        	}
         }
     }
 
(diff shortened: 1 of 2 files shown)


johnjaylward:
@stleary I don't think this was necessary. See my comment on #260

stleary:
#249 was supposed to be a minor refactoring, but it contains an uncaught bug that breaks the opt*() methods. It needs to be reverted,

**Changes to the API?**
No change to the API

Yes, the 20160807 Maven release contains this bug.

Not required

**Change to unit tests?**
No changes for the revert, but new unit tests will be added later.

**Note**
This pull request is being fast-tracked, so if you have comments, get them in quickly. Will leave the request open for 1 day.

johnjaylward:
@stleary looking at the reversal, this does not solve the issue.

stleary:
@johnjaylward please elaborate.

johnjaylward:
the functions optBoolean, optLong, optInt, etc were calling the get method internally before the original changes were made that this is reverting.

johnjaylward:
https://github.com/stleary/JSON-java/blob/master/JSONObject.java#L513

it is still calling get() without a try-catch block, so the issue in #260 is not solved.

johnjaylward:
oh nevermind. I thought this was merged already, but I now see it's not. I was mistaken when I saw the closed status above.

stleary:
No worries, let me know if you see any other reason not to revert, am otherwise strongly inclined to proceed.

johnjaylward:
I ran the tests against the reverted branch just now, and they all seem to pass, while there are failures on master. At this time I see no reason to hold the revert.

johnjaylward:
You may want to just add a new commit after the revert to correct the comment fixes that @Simulant87 made in #249. Those are valid fixes.

Simulant87:
@johnjaylward is right. changing

`this.get(key);`

to

`this.opt(key);`

in all the optXXX methods fixes the bug. I think I can submit a new merge-request to fix this bug in 12 hours, if you can wait for this. Otherwise someone else can do the fix.

PR #262, opened 2016-08-09
Consistently use this.opt()

### LINKED ISSUE #260: optBoolean isn't optional

quidryan:
Fixes #260

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -942,7 +942,7 @@ public class JSONObject {
      * @return The truth.
      */
     public boolean optBoolean(String key, boolean defaultValue) {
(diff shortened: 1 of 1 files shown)


stleary:
Thanks, but in this case a revert is preferred. Please see #261

Issue #255, opened 2016-07-29
warp of Sets of Enums is broken


Simulant87:
When I use an Enum with a "getter"-method and put this value into a set and this set as a value of a JSONObject the String representation of the Enum is not the same as when I put the Enum directly as a value of the JSONObject. This issue depends on the `wrap` method of JSONObject that is called when a JSONArray (used for the Set) is created from a collection.

Here is an example.
```
    enum ToStringEnum{
        a,
        d(false);

        private final boolean flag;

        ToStringEnum() {
            this(true);
        }

        ToStringEnum(boolean disabledOnDefault) {
            this.flag = disabledOnDefault;
        }

        public boolean isFlag() {
            return this.flag;
        }
    }

    ToStringEnum a = ToStringEnum.a;
    JSONObject object = new JSONObject();
    Set<ToStringEnum> set = new HashSet<ToStringEnum>();
    set.add(a);
    object.put("testset", set);
    object.put("test", a);
    System.out.println(object.toString());
    //result: {"testset":[{"flag":true}],"test":"a"}
    //expected: {"testset":["a"],"test":"a"}
```

stleary:
Enums have partial support in JSON-Java. You can get an idea of the extent of that support by reviewing the tests and comments in  https://github.com/stleary/JSON-Java-unit-test/blob/master/src/test/java/org/json/junit/EnumTest.java. Let me know if you think this needs to be more clearly documented in the JSON-Java README file.

johnjaylward:
@stleary I think this is a valid concern we should look at fixing. Now that we are openly supporting java1.6+ I think it would be good to come up with a good definition of how Enums are handled and make sure it's consistent.

Having the same value produce 2 different results does not seem optimal to me. Would you like me to start a PR for this?

Issue #145, opened 2015-08-09
JSONObject.wrap does not handle enums


abewell:
Steps to reproduce:
Create a new JSONObject();
try to add an enum as a value
dump the jsonObject via the toString() method

expected:
we see the enum values

actual:
we do not see the enum values

Please see below stack. It seems that the JSONObject.wrap function fails to recognize the enum and tries to use the bean constructor from the JSONObject class. My enum does not have getters/setters like (since it is not a bean) and so nothing gets dumped from the toString()

JSONObject.wrap(Object) line: 1730	
JSONArray.<init>(Collection<Object>) line: 159	
JSONObject.writeValue(Writer, Object, int, int) line: 1767	
JSONObject.write(Writer, int, int) line: 1836	
JSONObject.toString(int) line: 1614	
JSONObject.toString() line: 1591	
JSONOutputRenderer.buildVariantsArray(Set<VariantInfo>) line: 100	
JSONOutputRenderer.getJson(BuildInfo) line: 83	
JSONOutputRenderer.generateBuildInfo(String, BuildInfo) line: 50	
GenBuildOutputHandlerMain.outputBuildInfo(String, BuildInfo) line: 24	
GenBuildOutputHandlerMain.processFile(String) line: 18	
GenBuildOutputHandlerMain.main(String[]) line: 34

I have pulled the latest source and built the following suggestion to fix the bug:

Added at line 1722 of JSONObject.java
            if(object instanceof Enum){
                return object.toString();
            }

stleary:
Thanks for pointing this out. Some explicit support for enum was added a few weeks ago. Please see the discussion in https://github.com/douglascrockford/JSON-java/pull/127. At the time, wrap() was mentioned but not included in the final list. Based on the constraints in the discussion thread, if you think you see a way that wrap() support for enum could be allowed, please let us know your ideas. Will leave this issue open a few days for feedback.

abewell:
Ah yeah, I read through your mentioned discussion. I see that it is more ambiguous than my simple case was. I don't usually use enums with getters/setters, but I can see how it would make sense too. I can't think of anything more than what you all already discussed for a general solution. So I'm not sure what the appropriate fix for this project would be. I will plan to just stick with my one off case for now. Thanks for pointing me to that thread. Take care.

stleary:
[...]

PR #271, opened 2016-08-11
Update enum support to be more fully featured.

johnjaylward:
Adds support to JSONObject wrap and write methods to explicitly handle Enums.

Enums were handled inconsistently as noted in the original improvement (#140 and #127) and brought up again in #145 and #255 

The new way enums are handled is to always place the actual enum in the JSONObject/JSONArray (not wrapped, similar to other value types). When writing, we always write the actual "name" of the enum, so even with a toString override on the enum class, the value remains consistent and compatible with the optEnum/getEnum methods.

The constructor JSONObject(Object) functions the same way as before when passing an enum and is consistent with other "value" types. For example, when creating a JSONObject with Long, Boolean, BigDecimal as the constructor parameter, the value will be treated as a "bean".

**Risks**
Low risk. Currently inserting enums into JSONObject and JSONArray are not consistent. Anyone depending on the existing handling of enums probably does so with if blocks to ensure they have the correct data. Additionally, some cases of inserting an enum would produce results that could not be read back using get|optEnum methods.
```java
Set<MyEnumField> set = new EnumSet.of(MyEnumField.VAL1);
jo.put("value", set);

assert (new JSONObject(jo.toString()).optEnum(MyEnumField.class, "value") != null
    : "Oops, couldn't read back the enum value.";
```

Given that the data is not always reversible, it's unlikely that people are depending on this behaviour.

**Changes to the API?**

Not required.

Yes, new unit tests have been added.

**Review status**
ACCEPTED. Starting 3-day comment window 15 Aug 2016

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -1708,6 +1708,9 @@ public class JSONObject {
         if (value.getClass().isArray()) {
             return new JSONArray(value).toString();
         }
+        if(value instanceof Enum<?>){
+            return quote(((Enum<?>)value).name());
+        }
         return quote(value.toString());
     }
 
@@ -1730,12 +1733,9 @@ public class JSONObject {
             }
             if (object instanceof JSONObject || object instanceof JSONArray
                     || NULL.equals(object) || object instanceof JSONString
-                    || object instanceof Byte || object instanceof Character
-                    || object instanceof Short || object instanceof Integer
-                    || object instanceof Long || object instanceof Boolean
-                    || object instanceof Float || object instanceof Double
-                    || object instanceof String || object instanceof BigInteger
-                    || object instanceof BigDecimal) {
+                    || object instanceof Number || object instanceof Character
+                    || object instanceof Boolean || object instanceof String
+                    || object instanceof Enum) {
                 return object;
             }
 
@@ -1797,6 +1797,8 @@ public class JSONObject {
             writer.write(numberToString((Number) value));
         } else if (value instanceof Boolean) {
             writer.write(value.toString());
+        } else if (value instanceof Enum<?>) {
+            writer.write(quote(((Enum<?>)value).name()));
         } else if (value instanceof JSONString) {
             Object o;
             try {
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -1733,9 +1733,12 @@ public class JSONObject {
             }
             if (object instanceof JSONObject || object instanceof JSONArray
                     || NULL.equals(object) || object instanceof JSONString
-                    || object instanceof Number || object instanceof Character
-                    || object instanceof Boolean || object instanceof String
-                    || object instanceof Enum) {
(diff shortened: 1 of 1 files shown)


stleary:
Thanks, will take a look at this tonight. Please add a Risk section with your evaluation of the risk of this change. You can see examples in #259 and #261.  For example, in what ways might existing users be impacted by the functional change?  

Agreed that enums are currently not well supported.

johnjaylward:
done.

stleary:
Is there a reasonable way to implement this such that existing users are unaffected, but new users can choose the new behavior?

johnjaylward:
I don't think so since it was inconsistent. Sometimes values would be
generated as a single string, and others as a Bean. However, the Bean
output was not that useful because neither the name or the ordinal of the
enum was ever output.

stleary:
Enum support is connected in an odd way to Big numbers: both represent Java features that did not exist when the lib was first implemented. At the time, we were fortunate that Douglas gave us pretty clear guidance on what we should and should not do, when implementing support for new features.  That is, it needs to be exactly correct, and it can't break existing code. Of course, there will be a little give and take between those two constraints. I am still trying to get a sense of whether this enhancement is within the gray area (_"Some rules can be bent; some can be broken" - Morpheus_), or if we are straying too far into the realm of breaking existing code.

johnjaylward:
I'd say that since the current implementation is inconsistent that this is
closer to a bug fix than an enhancement.

johnjaylward:
Also to note, this library did have support for enums a while ago, but
Douglas removed it when I pointed out the readme said it compiles on java
1.4.

stleary:
Good information, thanks. Would that be the change and revert to JSONObject.wrap() in 2012? Or was there more?

johnjaylward:
Yes, just those 2 commits as far as I know. One in July adding support,
then the reversal in October.

stleary:
Taking small steps...

- Of course the unit tests will fail if they are positively tracking the expected behavior, so that is not, in itself, a problem.
- Would prefer to omit any changes not directly related to Enum, so it would be better if the Number changes are not included in this commit.
- Have not analyzed it yet, but re-implementing Douglas' enum change should be fine. That would be the wrap() change, I guess.
- It seems that the valueToString() change is a necessary action if wrap() inserts the enum. Assuming that was the purpose, and this was missed in the 2012 checkin because there were no (or at least no good) unit tests at the time.

Let me know what you think.

johnjaylward:
That sounds good. The changes to valueToString are indeed needed to prevent
default use of the enums toString method. We want to garuntee that the name
of the enum is output.

For the number changes, I'll revert those here and place them in a separate
pull request.

johnjaylward:
ok, cleanup up the commit as well as merging in the latest changes you made for the comments. Also update my notes in comment#1

stleary:
Thanks, starting a 3 day comment window.

PR #277, opened 2016-08-18
refactor JSONObject.write to use entrySet

Simulant87:
Using the entrySet of the backing HashMap has a better perfomance than iterating over the keySet and resolving the value with an additional get on the map. Reading the code makes it also more clear, that the key and the value are always from the same entry.

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -1835,23 +1835,23 @@ public class JSONObject {
     public Writer write(Writer writer, int indentFactor, int indent)
             throws JSONException {
         try {
-            boolean commanate = false;
             final int length = this.length();
-            Iterator<String> keys = this.keys();
+            Iterator<Entry<String, Object>> iterator = map.entrySet().iterator();
             writer.write('{');
 
             if (length == 1) {
-                Object key = keys.next();
-                writer.write(quote(key.toString()));
+                Entry<String, Object> entry = iterator.next();
+                writer.write(quote(entry.getKey()));
                 writer.write(':');
                 if (indentFactor > 0) {
                     writer.write(' ');
                 }
-                writeValue(writer, this.map.get(key), indentFactor, indent);
+                writeValue(writer, entry.getValue(), indentFactor, indent);
             } else if (length != 0) {
+                boolean commanate = false;
                 final int newindent = indent + indentFactor;
-                while (keys.hasNext()) {
-                    Object key = keys.next();
+                while (iterator.hasNext()) {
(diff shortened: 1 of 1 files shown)


johnjaylward:
I wouldn't mind seeing this taken a step further to use the newer for loop syntax. I don't think we need to maintain the iterator in the loop.

The initial `if` check on the length can be removed altogether as well since it's just code duplication
```java
for (Entry<String,?> entry : map.entrySet() ) {

}
```

stleary:
Thanks, but not at this time, don't think the implementation is inefficient enough to justify a change. If you can quantify otherwise, no objections to looking at it again.

johnjaylward:
@stleary there are documented performance notices on the use of entryset over keyset/get in multiple places:

* Stack Overflow
 * http://stackoverflow.com/questions/3870064/performance-considerations-for-keyset-and-entryset-of-map
 * http://stackoverflow.com/questions/8962459/java-collections-keyset-vs-entryset-in-map
* http://www.learnjavaj2ee.com/2015/12/keyset-vs-entryset-in-map.html

I believe the first link with give you a good idea of the benefits of using the entryset directly vc the keyset/get.

In summary though, the keyset if derived from the entryset, then you have to call get for each value which is approximately O(n) for the keyset generation and O(1) for the get.

Using the entryset you have a O(1) for the entryset generation, then a O(1) for getting the value.

Simulant87:
I used the test demonstrated here:
http://howtodoinjava.com/core-java/related-concepts/performance-comparison-of-different-ways-to-iterate-over-hashmap/

I added a loop aroud the get-part to test this more often and also repeated the test on a smaller map.

Using a Map with 50 entries
retrival repeated 10,000,000 times for each loop implementation
Using keySet() 	 in for-each loop 	: 3300 ms
Using keySet() 	 and iterator 		: 3278 ms (current implementation)
Using entrySet() in for-each loop 	: 2299 ms
Using entrySet() and iterator 		: 1871 ms (target implementation)

johnjaylward:
Yes, 50 items would not make a large difference as the majority of the work for the keyset implementation is building the keyset itself. Try a test with 50,000 entries (and less iterations) and you may see better performance for the entry set.

However, I think your test is more indicative of actual use-case sizes for more JSONObject map sizes. I doubt many people would have large objects with 50,000 keys.

I mostly just think the for-each loop on the entry set is much more readable. Having iterators and looking up the value a second time.

I'm not sure those performance numbers justify making a change now that I see them, but I would still like it for readability if nothing else.

PR #283, opened 2016-09-07
Tweaked JSONObject.getString() to evade exceptions

CPColin:
Right now, if the value `JSONObject.getString()` is trying to retrieve happens to be a number (or anything else that's not a `String`), you get a `JSONException`. Looking in from the outside of the API, I would expect `getString()` to return the `String` representation of the value, without being sensitive to the type it _should have_ been.

So this change calls `toString()` on the value `JSONObject.get()` returned and returns that, instead of throwing an exception.

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -706,20 +706,17 @@ public class JSONObject {
     }
 
     /**
-     * Get the string associated with a key.
+     * Get the string representation of the value associated with a key.
      *
      * @param key
      *            A key string.
-     * @return A string which is the value.
+     * @return The string representation of the value.
      * @throws JSONException
-     *             if there is no string value for the key.
+     *             if there is no value for the key.
      */
     public String getString(String key) throws JSONException {
         Object object = this.get(key);
-        if (object instanceof String) {
-            return (String) object;
-        }
-        throw new JSONException("JSONObject[" + quote(key) + "] not a string.");
+        return object.toString();
     }
 
     /**
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -715,8 +715,7 @@ public class JSONObject {
      *             if there is no value for the key.
      */
     public String getString(String key) throws JSONException {
-        Object object = this.get(key);
-        return object.toString();
+        return this.get(key).toString();
     }
 
     /**


CPColin:
Right. Everything _should_ have some sort of string representation, so the conversion can't fail, unlike with integers and other data types. Although I suppose calling `getString()` on an object that's actually an array or other complex object could trip somebody up.

JSONObject.java
```diff
@@ -706,20 +706,17 @@ public long getLong(String key) throws JSONException {
...
-     *             if there is no string value for the key.
+     *             if there is no value for the key.
      */
     public String getString(String key) throws JSONException {
         Object object = this.get(key);
-        if (object instanceof String) {
-            return (String) object;
-        }
-        throw new JSONException("JSONObject[" + quote(key) + "] not a string.");
+        return object.toString();
```
johnjaylward: Just a note for other reviewers, the get(key) will throw an exception if the key or the value is null, so no null check is needed here.

I'm not sure if you want to make this more succinct though as
```java
return this.get(key).toString();
```

I'm personally not a fan of local variables that are only used once.

CPColin: Sure, I can switch that. I might have left it at two lines so one fewer line would change in the pull request, haha.

stleary:
Please include a statement about the effect of this code change on existing applications.

CPColin:
Applications that formally encountered `JSONException`s when calling `JSONObject.getString()` for a value that wasn't actually a String will no longer see those exceptions. Instead, the `toString()` method will be called on whatever object `JSONObject.get()` would have returned and that String will be returned. An application that, for some reason, relied on the exception being thrown will have to call `JSONObject.get()` directly and pay closer attention to the type of object returned.

stleary:
Respectfully disagree with @johnjaylward about changing this method, since it seems to me that the other get[type]() methods are intended to return the value only if it matches that type. Agreed that sometimes those methods do perform conversions, though. Also, prefer not to change the API and advertised behavior except in the case where the existing code is clearly incorrect or does not comply with the spec, and even then a strong case has to be made. That does not appear to be the case here. However, no objection if you want to provide a new API method that unconditionally returns a string representation of the value, for example, public String getAsString(String key).

CPColin:
The problem, for me, is that I see `getString()` and already assume it should work like `getAsString()` would. It feels natural to me that calling `getInt()` would throw an exception if the value were not an integer or could not be converted to one. It also feels natural to me that calling `getString()` would throw an exception if the value were not a String or could not be converted to one. The issue is that there _isn't_ a value that `get()` can return that can't be converted to a String, so you can see why I was surprised when it threw an exception.

The main thing I'm afraid of is having a developer on my team who is not 100% familiar with this library look to see which method is appropriate for retrieving a String, choose `getString()`, and have that code blow up much later when, say, user fiddling or a loose design winds up submitting an integer instead. I would not expect people to see `getString()` in their IDE's list of autocompletions, but opt for `getAsString()` instead.

I do understand, though, if you want to reject this pull request, based on your reasoning.

johnjaylward:
@CPColin your note of `getAsString()` is good. Maybe it would be better to create a new function to perform that task instead of modifying the existing `getString()`

CPColin:
Incidentally, the reason this is coming up at all is because we're trying to update from an ancient version of this library to the latest version, so we can take advantage of `JSONException` extending `RuntimeException`. The version we have is from before commit f4cb14728f13629972a0ea76bb3dc0705a735fa8, which is where the `instanceof String` check was added. (So it's definitely our self-interests that are motivating this pull request.)

stleary:
Thanks for the request. Appreciate your project's requirements, but precedence must be given to applications that expect the API to work in the advertised way. No objections to a pull request for a new API method that unconditionally returns a String value.

PR #292, opened 2016-10-05
Provides "#" string evaluation support for JSON Pointer

erosb:


stleary:
In the JSON Pointer spec, "#" and "" are equivalent strings

**Risks**
Low. This adds required functionality to an ancillary JSON-Java feature. While existing apps may erroneously depend on the existing behavior, this is still a relatively new feature.

**Changes to the API?**
No.

May be rolled into the next release.

No.

No, but a new unit test is added to exercise this code change.

**Review status**
RECOMMEND ACCEPT. Starting 3 day comment window with this post.

PR #297, opened 2016-11-01
Add equals() and hashCode() implementations to JSONArray and JSONObject

ndrewln:
Tests for this functionality are located here https://github.com/stleary/JSON-Java-unit-test/pull/63

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -186,6 +186,20 @@ public class JSONArray implements Iterable<Object> {
         return myArrayList.iterator();
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (o == this) return true;
+        if (!(o instanceof JSONArray)) return false;
+
+        JSONArray otherArray = (JSONArray) o;
+        return myArrayList.equals(otherArray.myArrayList);
+    }
+
+    @Override
+    public int hashCode() {
+        return myArrayList.hashCode();
+    }
+
     /**
      * Get the object value associated with an index.
      *
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -1631,6 +1631,20 @@ public class JSONObject {
         }
     }
 
+    @Override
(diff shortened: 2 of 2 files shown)


johnjaylward:
These are not doing any kind of deep comparison. Equals for both objects is shallow here and will almost always return false for the map/array comparisons. 

I'm not sure I see a benefit to using these implementations over the default ones built into Java. 

Is there a usecase for your application that requires a non-default implementation?

amolokoedov:
+1

ndrewln:
According to my experience and official documentation, standard java containers do deep comparison in equals() implementations:
https://docs.oracle.com/javase/7/docs/api/java/util/AbstractList.html#equals(java.lang.Object)
https://docs.oracle.com/javase/7/docs/api/java/util/AbstractMap.html#equals(java.lang.Object)

I tried to use comparison of org.json containers in REST API tests:
```java
assertEquals(new JSONArray("[1,2,3]"), myAPI.requestIds());
```

That check failed despite actual response data were same as reference one. I thought it would be a good idea to make json containers use deep comparison as all other containers do.

johnjaylward:
We currently have `similar` methods to do the deep comparisons. For your test bench you can try something like this instead:

```java
assertTrue(new JSONArray("[1,2,3]").similar(myAPI.requestIds()));
```

Code for similar here:
* https://github.com/stleary/JSON-java/blob/master/JSONObject.java#L1483
* https://github.com/stleary/JSON-java/blob/master/JSONArray.java#L1024

Docs:
* http://stleary.github.io/JSON-java/org/json/JSONObject.html#similar-java.lang.Object-
* http://stleary.github.io/JSON-java/org/json/JSONArray.html#similar-java.lang.Object-

stleary:
Thanks, but cannot accept this pull request.

PR #299, opened 2016-11-09
add new method call removeAll

MetaiR:
it's remove a group of keys.


johnjaylward:
This can already be done through the `JSONObject.keySet().removeAll(Collection<?>)` method.

stleary:
Thanks for the pull request, but not accepting for the reason cited.

PR #301, opened 2016-11-17
Add missing `break` statements

Arinerron:


johnjaylward:
I'm pretty sure those fall-throughs were purposeful.

stleary:
Surprisingly it does not break the unit tests, although maybe this just means the tests are not thorough enough. Thanks for the pull request, but can't accept changes without context or justification. If you can find a use case where this change fixes a bug or is otherwise beneficial, feel free to re-submit.

PR #306, opened 2017-01-04
There are use cases that require a canonical form of the JSON string …

PyMeH:
…(or at least a form where JSON attributes appear in a sorted order)

E.g. security applications where client and server need to construct a JSON structure and sign it, and the other party need to validate the signature.
With this change JSONObject allows its children to override the type of java.util.Map being used, thus allowing for TreeMap, thus allowing sorted JSON, thus allowing two clients to generate the same JSON string, given they provide the same attributes.

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -41,6 +41,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.ResourceBundle;
 import java.util.Set;
+import java.util.TreeMap;
 
 /**
  * A JSONObject is an unordered collection of name/value pairs. Its external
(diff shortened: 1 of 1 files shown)


PyMeH:
...I've read the FAQ, and I understand that by design JSON is a collection of unordered attributes.
What I need is to have **the option** to implement a canonicalizer, comparator and a more human-friendly String representation.
IMO https://tools.ietf.org/html/draft-staykov-hu-json-canonical-form-00 captures the needs very well: "... While convenient for human interaction, this  flexibility is inconvenient for cases where a machine is used to  assess the logical equivalence of documents."

johnjaylward:
I'm not sure this is comprehensive. If you add a Map to a JSONArray a JSONObject will be added and not your possible subclass. This, and other examples, could lead to unexpected inconsistencies in ordering. I think as long as you document it, it should be fine.

You should also remove the TreeMap from the imports as it's not used.

stleary:
Thanks for the suggestion. Although it cannot be accepted due to the reasons stated in the Wiki, it seems like a reasonable enhancement to include in a forked project. Per @johnjaylward's suggestion, recommend looking into other classes that create new JSONObjects, including JSONArray, Cookie, JSONML, JSONTokener, JSONWriter, Property, and XML.

PR #310, opened 2017-01-13
bug 309 accept multi-line string values

arikast:
This PR fixes my use case, but of course pls check for regressions at your end.

While fixing I also made a quick maven project and test case (not included in this PR).  Let me know if you'd be interested in receiving those in another PR.

### DIFF
--- a/JSONTokener.java
+++ b/JSONTokener.java
@@ -264,8 +264,6 @@ public class JSONTokener {
             c = this.next();
             switch (c) {
             case 0:
-            case '\n':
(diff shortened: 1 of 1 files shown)


johnjaylward:
There is already a test project, see https://github.com/stleary/JSON-Java-unit-test

it happens to be a gradle project

arikast:
Ah ok.  Running those tests now I see lots of fails with my PR.  I'll look into it and submit a revised PR.

PR #312, opened 2017-01-17
stored the key names in order and add isInt

yfor:
stored the key names in order.Most time hash code order is ok,some case order it is  useful. With type get method ,the opt** is just try cast ,we need is** (isInt).

### DIFF
--- a/JSONObjectKeyOrder.java
+++ b/JSONObjectKeyOrder.java
@@ -1,0 +1,58 @@
+package org.json;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * sometimes we need key order in input,hashmap`s keyset is chaos.
+ * @author wangq
+ *
+ */
+public class JSONObjectKeyOrder extends JSONObject {
+	// store the name in order
+	private List<String> names;
+	
+	public JSONObjectKeyOrder(String source) throws JSONException {
+
+		this(new JSONTokenerKeyOrder(source));
+	}
+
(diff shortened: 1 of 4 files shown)


stleary:
Support for ordering is one of the most popular requests. However, from the JSON spec RFC 7159: "An object is an unordered collection of zero or more name/value pairs". As a reference app, the design is required to stay as close to the spec as possible, so ordering in JSONObjects is not supported.

yfor:
ok. The other thing is 'Get an optional ** value ',such as optString(String key), use  it get the type of value  is need some trick.

stleary:
你好, OK, let's talk about that. First, some ground rules. Please submit one change at a time, lest the entire pull request gets rejected because of a single problem. The existing  JUnit test framework should be used for tests: https://github.com/stleary/JSON-Java-unit-test. Regarding your API change, proposed enhancements and bug fixes require a compelling use case.  For example:

* Why is the method JSONObject.isInt() needed?
* What about all of the other numeric and non-numeric types?
* Will this have any effect on existing applications?

FYI, there are currently two active issues on this project: 

* XML support on the Android platform. Problematic because our license terms have resulted in the open source community forking the project at an earlier point, resulting in different classes sharing the same namespace and class names.
* To what extent the project should be modified to allow for more flexible XML processing. There are ongoing requests to make the parsing code more XML-friendly. We need a balanced response that supports the users without breaking the integrity of the project. This is probably the biggest challenge right now.

Any work you might want to do on the above issues would be appreciated, and would be more likely to be accepted.
谢谢, Sean

Issue #237, opened 2016-06-07
Ignoring caught exceptions makes this library unusable


successhawk:
The profuse usage of the anti-pattern of ignoring caught exceptions makes this library unusable.
try{...} catch (Exception ignore) {}

Please modify all occurrences to perform proper exception handling like throwing some form of RuntimeException wrapping the cause.
At a minimum, make it be able to log the exceptions in some type of debug mode.  I had to download and modify the source to debug my problem.

I found at least 5 occurrences in JSONObject.

johnjaylward:
I've never had a problem with it using the library for 10+ years. can you give some examples of your data input, calls, etc so we can know what you are trying to do? In  many cases in the library, like all the opt* methods, it is proper and expected to eat the exceptions and return either null or a default value.

successhawk:
All of my problems stemmed from the class not being public.  Here is a simple example.  I admit that I should have read the javadocs closer.  But it was surprising to me that when I explicitly told it what field to parse, it didn't throw an exception when it couldn't, and it wasn't easy for me to determine why it couldn't (there are many things that could go wrong when you catch Exception) until I modified the JSONObject code to print the exception.
 
`
System.out.println( new JSONObject(new Test(), new String[] {"message"}) );
`
```
// non-public class
class Test {
	public String message = "test";
}
```

johnjaylward:
yes, I see what you mean now. It would make sense to split that exception handling up and ignore only specific exceptions that we knew about, or possibly expand the logic to prevent them in the first place.

stleary:
Thanks for the suggestion. If interested, you can offer a pull request with your proposed changes. More information about what kinds of code changes are accepted can be found here: https://github.com/stleary/JSON-java/wiki/FAQ

Mormuig:
It's a reasonably old post, but I run into that kind of problem to.
I have made some changes so that it's possible to get logs of otherwise ignored Exceptions, without changing the actual behavior.
May I post a pull request ?

stleary:
Sure, if you can do it without using 3rd party libs, and if the logging is opt-in (not enabled by default).

Mormuig:
I've done it using only Java Logging, so no 3rd party libs.
To keep the logging disabled by default, I used a low log level, as the goal is to give the possibility to log just by tweaking JSONObject and/or JSONArray loggers (or parents).

PR #313, opened 2017-01-26
Either log or rethrow otherwise invisible Exceptions.

Mormuig:
When possible, just put the original Exception as the cause of thrown
JSONException.
Using log to give possibility to trace other exceptions, with FINE level
so they don't show up in a default configuration (default ConsoleHandler
log from INFO to SEVERE).
Linked to #237 issue.

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -35,6 +35,8 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * A JSONArray is an ordered sequence of values. Its external text form is a
@@ -82,6 +84,11 @@ import java.util.Map;
  */
 public class JSONArray implements Iterable<Object> {
 
+    /**
+     * Logger for retrieving otherwise invisible Exceptions.
+     */
+    private static final Logger LOGGER = Logger.getLogger(JSONArray.class.getName());
+    
     /**
      * The arrayList where the JSONArray's properties are kept.
      */
@@ -244,7 +251,7 @@ public class JSONArray implements Iterable<Object> {
             return object instanceof Number ? ((Number) object).doubleValue()
                     : Double.parseDouble((String) object);
         } catch (Exception e) {
-            throw new JSONException("JSONArray[" + index + "] is not a number.");
+            throw new JSONException("JSONArray[" + index + "] is not a number.", e);
         }
     }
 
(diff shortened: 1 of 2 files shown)


Mormuig:
It use standard Java Logging, so it should not error in a standard JRE.
The default output is console (defined by the root logger ConsoleHandler), but it can be changed either by config file or incode.

stleary:
Thanks, looks good. 

It would be helpful to have some text that describes concrete scenarios about how someone would make use of the new code. For example:

- How to configure logging in order to view an ignored exception, with a code fragment that causes the exception
- Are there any considerations, and how to set up the logging config for executing the code in an IDE (e.g. Eclipse)

I don't think that most of the changed code is exercised by the unit tests: https://github.com/stleary/JSON-Java-unit-test. I can do this when time permits. But if you have time and want to add the unit tests, that would help move things along.

Mormuig:
Here is an example with the code of the referenced issue.
In this example, no other configuration is needed, it while show in the IDE console while executing the main method.
I will provide another example using a file configuration.

```
package main;

import java.util.logging.ConsoleHandler;
import java.util.logging.Level;
import java.util.logging.Logger;

import org.json.JSONObject;

public class MainInCode {

        // Nothing logged, just print an empty JSONObject
        System.out.println(new JSONObject(new Test(), new String[] { "message" }));

        // Get the logger associated with JSONObject
        Logger logger = Logger.getLogger(JSONObject.class.getName());
        // Setting the log Level to FINE (INFO is default)
        logger.setLevel(Level.FINE);
        // Tell the logger to not use the parents Handlers
        logger.setUseParentHandlers(false);

        // Creating a new ConsoleHandler (could be any handler)
        ConsoleHandler console = new ConsoleHandler();
        // Setting the log level of this Handler to FINE
        console.setLevel(Level.FINE);

        // Associate the two
        logger.addHandler(console);

        // Error logged, and still print the empty JSONObject
        System.out.println(new JSONObject(new Test(), new String[] { "message" }));
    }

    // non-public class
    static class Test {
        public String message = "test";
    }
}
```

johnjaylward:
I don't think there is a good way to test loggers short of writing your own "test" implementation to the interface that will trigger something.

@stleary , these changes look good to me as long as they aren't breaking any tests. The changes don't appear to change any logic flow.

stleary:
@johnjaylward, no objection to proceeding, unit tests are still needed.

Mormuig:
I used a StreamHandler on the Logger to catch emitted logs.
I'm working on the normally unreachable logs (those in catch block), hope to be done shortly.

johnjaylward:
@Mormuig I incorporated the exception wrapping changes into the latest master.  If you can merge or rebase your PR we can look at just the logging changes then.

stleary:
Thanks for the pull request, but can't be accepted at this time.

Issue #315, opened 2017-02-10
Construct a JSONObject from an Object using bean getters is inconsistent across locales


jakope0003:
Construct a JSONObject from an Object using bean getters produce unexpected result when locale is set to Turkey.

Having following two classes `Bean.java` and `JSONIssue.java`.

```java
public class Bean {

    private final String id;

    public Bean(String id) {
        this.id = id;
    }

    public String getId() {
        return id;
    }

}
```

```java
import java.util.Locale;
import org.json.JSONObject;

public class JSONIssue {

    public static void main(String[] args) {
        Locale.setDefault(new Locale("en"));
        JSONObject jsonen = new JSONObject(new Bean("beanId"));
        System.out.println("jsonen " + jsonen);

        Locale.setDefault(new Locale("tr"));
        JSONObject jsontr = new JSONObject(new Bean("beanId"));
        System.out.println("jsontr " + jsontr);
    }

}
```

Output differs for different locales.

```
jsonen {"id":"beanId"}
jsontr {"?d":"beanId"}
```

I'm expecting to have consistent id key no matter what locale is set.

johnjaylward:
Can you print both JSON after the local change? 

Save the first JSON string to a var, change the local,  then print both?

jakope0003:
Please find proposed patch:

```java
diff --git a/JSONObject.java b/JSONObject.java
index 4df256a..14a86d5 100644
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -1159,7 +1159,7 @@ public class JSONObject {
                         if (key.length() == 1) {
                             key = key.toLowerCase();
                         } else if (!Character.isUpperCase(key.charAt(1))) {
-                            key = key.substring(0, 1).toLowerCase()
+                            key = key.substring(0, 1).toLowerCase(Locale.ROOT)
                                     + key.substring(1);
                         } 
```

jakope0003:
This is the case described in `String#toLowerCase` javadoc:

> Note: This method is locale sensitive, and may produce unexpected results if used for strings that are intended to be interpreted locale independently. Examples are programming language identifiers, protocol keys, and HTML tags. For instance, "TITLE".toLowerCase() in a Turkish locale returns "t\u0131tle", where '\u0131' is the LATIN SMALL LETTER DOTLESS I character. To obtain correct results for locale insensitive strings, use toLowerCase(Locale.ROOT).

johnjaylward:
Thanks!

stleary:
[...]

PR #317, opened 2017-02-10
make sure locale independent data is not upper/lowercased incorrectly…

johnjaylward:
…. See #315

### DIFF
--- a/HTTP.java
+++ b/HTTP.java
@@ -25,6 +25,7 @@ SOFTWARE.
 */
 
 import java.util.Iterator;
+import java.util.Locale;
 
 /**
  * Convert an HTTP header to a JSONObject and back.
@@ -74,7 +75,7 @@ public class HTTP {
         String         token;
 
         token = x.nextToken();
-        if (token.toUpperCase().startsWith("HTTP")) {
+        if (token.toUpperCase(Locale.ROOT).startsWith("HTTP")) {
 
 // Response
 
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -1157,9 +1157,9 @@ public class JSONObject {
                             && Character.isUpperCase(key.charAt(0))
                             && method.getParameterTypes().length == 0) {
                         if (key.length() == 1) {
-                            key = key.toLowerCase();
+                            key = key.toLowerCase(Locale.ROOT);
                         } else if (!Character.isUpperCase(key.charAt(1))) {
-                            key = key.substring(0, 1).toLowerCase()
+                            key = key.substring(0, 1).toLowerCase(Locale.ROOT)
                                     + key.substring(1);
                         }
 


johnjaylward:
Just for reference, Locale.ROOT was added in java6, which is why it was not in the original implementation. So this meets our current build requirement of JDK6.

stleary:
Sounds reasonable. I thought there was a Character.toLowerCase() in XMLTokener, is that not an issue? Unit tests will be needed as well.

johnjaylward:
I didn't see a locale option for the Character.toLowerCase() calls. Only for String.

johnjaylward:
@stleary maybe in the XML tokener instead of calling toLowerCase for each character, we call it with the toString on https://github.com/stleary/JSON-java/blob/master/XMLTokener.java#L139

like this: 
```java
StringBuilder sb = new StringBuilder();
        for (;;) {
            char c = next();
            if (Character.isLetterOrDigit(c) || c == '#') {
                sb.append(c);
            } else if (c == ';') {
                break;
            } else {
                throw syntaxError("Missing ';' in XML entity: &" + sb);
            }
        }
        String string = sb.toString().toLowerCase(Locale.root);
        Object object = entity.get(string);
        return object != null ? object : ampersand + string + ";";
```

johnjaylward:
Actually, for the ```nextEntity``` call, the locale should probably set to Locale.EN since all entities are either numeric or latin1 encoded.
* `<` = `&lt;` or `&#0060`

stleary:
I can add unit tests tonight, unless you will be making further changes to the code.

johnjaylward:
I won't make any more changes unless you think we need to modify the XMLTokenizer code. Thanks for taking care of the tests.

stleary:
Glad we have unit tests! Looks like JSONPointer needs more work, but not yet sure where.

```
public class MyLocaleBean {
    private final String üx = "Tlocaleüx";
    public String getÜx() {  return üx;    }
}

MyLocaleBean myLocaleBean = new MyLocaleBean();
JSONObject jsonObject = new JSONObject(myLocaleBean);
String str1 = (String)jsonObject.get("üx");
String str2 = (String)jsonObject.query("/getÜx");
System.out.println("str1: " + str1 + " str2: " + str2);
```
str1: Tlocale?x 
str2: null

stleary:
On second thought:
```
String str1 = (String)jsonObject.get("üx");
String str2 = (String)jsonObject.query("/üx");
```
str1: Tlocale?x str2: Tlocale?x

johnjaylward:
I think it's OK for the values. I'm not too clear on how the locale effects
string output.  Maybe test the output in Locale.EN as well and see what you
get?

On Feb 11, 2017 01:32, "Sean Leary" <notifications@github.com> wrote:

On second thought:

String str1 = (String)jsonObject.get("üx");
String str2 = (String)jsonObject.query("/üx");

str1: Tlocale?x str2: Tlocale?x

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/stleary/JSON-java/pull/317#issuecomment-279125437>, or mute
the thread

stleary:
Turned out I was just using the wrong query string, fixed now. But this code has to be saved as UTF, so I am putting the tests into a different file.  Don't want someone saving JSONObjectTest as ASCII and having the tests fail.

stleary:
The following tests are successful **without** using the code in this pull request.
```
public class MyLocaleBean {
    private final String üx = "Tlocaleüx";
    private final String ü = "Tlocaleü";
    public String getÜx() {        return üx;    }
    public String getÜ() {        return ü;    }
}
        assertTrue("expected Tlocaleüx",
                "Tlocaleüx".equals(jsonObject.getString("üx")));
        assertTrue("expected Tlocalü",
                "Tlocaleü".equals(jsonObject.getString("ü")));
        assertTrue("expected Tlocaleüx",
                "Tlocaleüx".equals((String)(jsonObject.query("/üx"))));
        assertTrue("expected Tlocalü",
                "Tlocaleü".equals((String)(jsonObject.query("/ü"))));
```

To run the test I had to add this to build.gradle:
```
tasks.withType(JavaCompile) {
     // this subproject requires -parameters option
     options.compilerArgs << '-parameters'
     options.encoding = 'UTF-8'
}
```
My compiler version is: 1.8.0_05

Any suggestions?

johnjaylward:
Try using the same example as the submitter. Specifically the Id with the
switch from EN to TR.

stleary:
JSONObject keys obtained from parsing beans should be invariant regardless of locale. This change enforces Locale.ROOT rules for lowercasing of the first letter of the key in get...() or is...() methods. 

**Risks**
Low risk. This is likely a one-off that most developers are unlikely to see. 

**Changes to the API?**
The first char of the key name is lowercased in a locale-invariant way when parsing beans.

This change can be rolled into the next release.

Not necessary at this time.

No, but a unit test was added that will cause breakage until this change is committed. See https://github.com/stleary/JSON-Java-unit-test/pull/64

No, just the requested change was made.

**Review status**
ACCEPT. Starting 3 day comment window with this post.

Issue #319, opened 2017-02-16
XML.toString outputs null strings rather than empty tags


mhubbard:
This is a reopening of issue #129 

As described in issue #129 when converting JSON to XML using XML.toString() null values withing the JSON results in the string "null" since XML has no concept of null. Instead, empty tags should be utilised as per the spec https://www.w3.org/TR/xml11/#sec-starttags.

XML.toString looks like it even accounts for this on lines 576-579 when it appends and empty tag

https://github.com/stleary/JSON-java/blob/5ff8b4cb0891db3b234e76b912e81ca27b0502ad/XML.java#L576-L579

```java
} else if ("".equals(value)) {
    sb.append('<');
    sb.append(key);
    sb.append("/>");
} else {
```

My proposed solution is to change line 536 as shown below 

https://github.com/stleary/JSON-java/blob/5ff8b4cb0891db3b234e76b912e81ca27b0502ad/XML.java#L536

```java
while (keys.hasNext()) {
    key = keys.next().toString();
    value = jo.opt(key);
    if (value == null || JSONObject.NULL.equals(value)) { // line 536
        value = "";
    }
...
```

This way we account for JSONObject.opt(key) returning null when the key is null, and also for when value is the JSONObject.NULL object (when the value is null in the json).

I tested the outlined fix and get empty xml tags for json null values. With approval, I'd be happy to submit the fix.

stleary:
@mhubbard, you can submit a pull request. The code change breaks 2 unit test now, in XMLTest.shouldHandleNullNodeValue() and JSONObjectTest.jsonObjectNullOperations(). The PR should make a case for why these tests should be modified or removed, and also why this won't break existing applications that might have been written to expect the "null" text value. No argument that the current transformation rules are arbitrary, and may have been better implemented, but the potential for breaking existing code needs to be considered before proceeding.

PR #321, opened 2017-02-17
JSONObject.NULL now outputs an empty tag when converting to xml with …

mhubbard:
This is to address issue #319 

To address @stleary's concerns I will be making a pull request to JSON-Java-unit-test as well with modifications to both XMLTest.shouldHandleNullNodeValue() and JSONObjectTest.jsonObjectNullOperations(). XMLTest.shouldHandleNullNodeValue already had code commented out to test the JSONObject.NULL -> empty xml tag, because this case isn't covered by any other tests I think it's important to modify the test rather than remove it. For the same reason I think modifying JSONObjectTest.jsonObjectNullOperations is the correct move, not to mention that test handles multiple other cases and removing it remove all JSONObject.NULL coverage.

Unfortunately there is some potential to break existing code, though anyone using XML.toString should be able to handle the new empty tags because it's a valid scenario (when the input json has a value that is an empty string) it's possible that some code expects a string of "null" in the instance of null in the json. I suspect that most code that's expecting a string of "null" is to treat it as a null value, but the potential for breakage is still there and should be considered.

### DIFF
--- a/XML.java
+++ b/XML.java
@@ -378,7 +378,7 @@ public class XML {
                     if (jsonobject.length() > 0) {
                         context.accumulate(tagName, jsonobject);
                     } else {
-                        context.accumulate(tagName, "");
+                        context.accumulate(tagName, JSONObject.NULL);
                     }
                     return false;
 
@@ -533,8 +533,8 @@ public class XML {
             while (keys.hasNext()) {
                 key = keys.next();
                 value = jo.opt(key);
-                if (value == null) {
-                    value = "";
+                if (JSONObject.NULL.equals(value)) {
+                    value = null;
                 } else if (value.getClass().isArray()) {
                     value = new JSONArray(value);
                 }
@@ -553,7 +553,7 @@ public class XML {
                             i++;
                         }
                     } else {
-                        sb.append(escape(value.toString()));
+                        sb.append(escape(String.valueOf(value)));
                     }
 
                     // Emit an array of similar keys
@@ -576,9 +576,19 @@ public class XML {
                 } else if ("".equals(value)) {
                     sb.append('<');
                     sb.append(key);
-                    sb.append("/>");
+                    sb.append('>');
+                    sb.append("</");
+                    sb.append(key);
+                    sb.append('>');
 
                     // Emit a new tag <k>
+
+                } else if (value == null) {
+                    sb.append('<');
+                    sb.append(key);
+                    sb.append("/>");
+
+                    // Emit a new empty tag <k/>
 
                 } else {
                     sb.append(toString(value, key));


XML.java
```diff
@@ -533,7 +533,7 @@ public static String toString(Object object, String tagName)
             while (keys.hasNext()) {
                 key = keys.next();
                 value = jo.opt(key);
-                if (value == null) {
+                if (value == null || JSONObject.NULL.equals(value)) {
```
johnjaylward: you don't need the OR here, simply `if (JSONObject.NULL.equals(value)) {` will do the same thing

stleary:
I think the current behavior of XML allows string transformations to be restored. For example, the following code fragment outputs {"abc":null}
```
        String str = "{ abc : null }";
        JSONObject jo = new JSONObject(str);
        str = XML.toString(jo);
        jo = XML.toJSONObject(str);
        System.out.println("Restored: " +jo.toString());
```
Simiilarly, this code fragment outputs \<abc>null\</abc>
```
        String str = "<abc>null</abc>";
        JSONObject jo = XML.toJSONObject(str);
        str = XML.toString(jo);
        System.out.println("Restored: " +str);
```

I think the ability to restore the original string is important. Not sure that this will work with the proposed code change.

XML.java
mhubbard: Ah you're right, I even had just looked at the JSONObject.NULL.equals method.

mhubbard:
@stleary that is not something I had thought about, you're right though the two to one mapping onto an empty tag means the original string can not be restored. The current implimentation makes sense to be now, though there is still a two to one mapping of "null" and JSONObject.NULL onto <tag>null</tag>.

I will try to think of a better solution.

mhubbard:
I have a solution that fixes both two to one mapping issues. The existing mappings for null/empty values are follows

{ "abc": "" }       <-> <abc/>
{ "abc": null }    <-> <abc>null</abc>
{ "abc": "null" }   -> <abc>null</abc>

My solution would map them like this

{ "abc": "" }       <-> <abc></abc>
{ "abc": null }    <-> <abc/>
{ "abc": "null" } <-> <abc>null</abc>

<element></element> and <element/> are functionally the same and are to be treated as such by xml parsers and readers, the specification for this is here https://www.w3.org/TR/REC-xml/#NT-content. Therefore this solution shouldn't break any existing code.

mhubbard:
So after looking for into it, due to XML.stringToValue(String string) calling JSONObject.stringToValue(string), any json keywords can not be recovered if they're represented as a string in the json. Meaning my last mapping of {"abc":"null"} can not be recovered, similarly neither can {"abc":"true"} or {"abc":"false"}; however I think this is the desired behavior. As it stands right now this PR changes conversion of null and "" in the following way.
```
{ "abc": "" }      <-> <abc></abc>
{ "abc": null }    <-> <abc/>
```
This however causes 14 tests in XMLTest to fail which I will modify to accomodate the new behavior as these cases should be covered. As stated in previous comments both types of empty elements are functionally the same to xml parsers and readers so this should not break any existing code

stleary:
This is characteristic of the problem we keep running into with JSON-XML transformations. There is no natural mapping, so something will always be broken. The sensible answer, already suggested, is to drop support for XML completely. Don't want to do that, but can't accept this pull request in its current form, nor do I think further changes along these lines will result in something we can use. For now, its probably best to stay with the current behavior.

seelenvirtuose:
Just to add something to this discussion (because we stumbled upon this issue last days):

Converting `{ "abc" : null }` to one of these possibilities

    <abc>null</abc>
    <abc/> 

is simply wrong because the first one should match the JSON `{ "abc" : "null" }` and the second one should match the JSON `{ "abc" : "" }`.

A `null` value in XML can be expressed in two ways:

1. Completely omit the element. This would also match the JSON equivalent to not having the element at all. However, a round-trip would be impossible with that approach.

2. Mark the element as _nilled_:

    <abc xsi:nil="true" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"/>

This is the correct expression for a `null` element in XML. The round-trip now would also be possible, as the only possible JSON conversion back is again `{ "abc" : null }`.

johnjaylward:
@seelenvirtuose thanks for the feedback. The XML support in this library is extremely simplified and we don't support namespaces. Outputting an attribute that depends on a namespace would be highly unusual in this library even if it is technically correct.

For more full-featured XML support I personally recommend the Jackson set of libraries. I believe that supports namespaces as well as a lot more functionality in the JSON<->XML conversions

PR #322, opened 2017-02-25
Allow user to invoke query and optQuery ,with a JSONPointer

dtalex:
It would be good for user's code readability to be able to invoke methods "query" and "optQuery" with a Provided JSONPointer


dtalex:
Sorry did a mistake.
I'll Publish a new pull request

PR #324, opened 2017-02-25
Allow user to invoke query and optQuery ,with a JSONPointer

dtalex:
It would be good for user's code readability to be able to invoke methods "query" and "optQuery" with a user-provided JSONPointer

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -980,7 +980,30 @@ public class JSONArray implements Iterable<Object> {
      * @return the item matched by the JSONPointer, otherwise null
      */
     public Object query(String jsonPointer) {
-        return new JSONPointer(jsonPointer).queryFrom(this);
+        return query(new JSONPointer(jsonPointer));
+    }
+    
+    /**
+     * Uses a uaer initialized JSONPointer  and tries to 
+     * match it to an item whithin this JSONArray. For example, given a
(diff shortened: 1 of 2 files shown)


stleary:
Looks reasonable, if nothing comes up in the unit tests, and no objections from @erosb. I will add some tests. Comments need to be cleaned up. For example spelling on line 1365, and param is not a string on line 1380.

erosb:
I checked this PR yesterday and I have no objections. Thanks.

stleary:
This is an enhancement, not a bug fix. API methods using JSONPointer for queries are added as a convenience.

**Risks**
Very low risk. 

**Changes to the API?**
Yes, new API methods are added:
```
jsonObject.query(JSONPointer)
jsonObject.optQuery(JSONPointer)
jsonArray.query(JSONPointer)
jsonArray.optQuery(JSONPointer)
```

No, this change can be rolled into the next release.

Not needed.

No.

Yes. The query(String) and optQuery(String) methods were changed to call the new API. This was a reasonable change.
**NOTE: ** There are a few typos in the comments, which should be fixed the next time this code is touched.

**Review status**
Approved. The usual 3 day review period is skipped, since this pull request has already been up for a month without objection. Will merge in 1 day.

PR #326, opened 2017-03-01
JSONObject and JSONArray suport stream and map/list functions

apolo92:


stleary:
Please resubmit, including a description of what you propose to change, and without the formatting updates which obscure the functional changes. You might want to open an issue first so we can discuss whether this particular change is needed.

johnjaylward:
looks like it was asking for Java8 features, which we aren't ready to support yet

Issue #37, opened 2011-12-13
Preserving insertion order of keys on JSONObject


mombiela:
Hello,

I know that JSON keys are unordered, but this make the transformation from JSON to XML unreliable, especially if you use xsd validation on the result. I suppose that some other formats are affected too.

I think that preserve the insertion order is a good improvement, with little impact on performance.

It should be as easy as replace java.util.HashMap with java.util.LinkedHashMap on JSONObject.

Thank you very much.
Joan

ygopala:
Is the order of JSON Array [] maintained when we convert from JSON to XML?

PR #329, opened 2017-03-13
To maintain the order of the fields from the input

pradeepcheers:
http://stackoverflow.com/questions/4515676/keep-the-order-of-the-json-keys-during-json-conversion-to-csv

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -155,7 +155,7 @@ public class JSONObject {
      * Construct an empty JSONObject.
      */
     public JSONObject() {
-        this.map = new HashMap<String, Object>();
+        this.map = new LinkedHashMap<String, Object>();
     }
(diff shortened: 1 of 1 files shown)


MinnDevelopment:
This was answered in many ways already but the first I found was #37 
Even the first answer on that thread says that the order of elements in a JSONObject is by definition not given.

erosb:
https://github.com/stleary/JSON-java/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aclosed%20order

https://github.com/stleary/JSON-java/wiki/FAQ

Bad luck :|

stleary:
Thanks for the pull request, but it cannot be accepted. Please see the Wiki page referenced in @erosb 's response.

PR #336, opened 2017-05-16
Numeric enhancements, Refactoring, Fix spelling

johnjaylward:
This is mostly cleanup of comments with a few changes to also remove some compile time warnings.

* JSONObject
  * `increment` I replaced the autoboxing calls with explicit accessors and also prevented runtime casting by using the proper long/float/double literals.
  * `Null` I added a hashcode function that explicitly returns 0 everytime, the same as most hash functions when handling null : `return someField == null ? 0 : someField.hashCode();`
  * Added a `put` function for `float`... for some reason we have code that handles floats in multiple places, but no way to directly add them into the JSONObject. Previously these would get converted to `double`
  * Modified the `put(literalNumber)` methods to use the valueOf static instead of always creating new instances.
* JSONPointer
  * Updated class references to be fully qualified with `this.`
  * Updated `constructor` parameters to be final. 
* XML
  * `toString(String, String)` updated parameters to be final

Let me know if I should roll any of the changes back and just keep the comment fixes.

### DIFF
--- a/CookieList.java
+++ b/CookieList.java
@@ -39,7 +39,7 @@ public class CookieList {
      * The pairs are separated by ';'. The names and the values
      * will be unescaped, possibly converting '+' and '%' sequences.
      *
-     * To add a cookie to a cooklist,
+     * To add a cookie to a cookie list,
      * cookielistJSONObject.put(cookieJSONObject.getString("name"),
      *     cookieJSONObject.getString("value"));
      * @param string  A cookie list string
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -183,7 +183,7 @@ public class JSONArray implements Iterable<Object> {
 
     @Override
     public Iterator<Object> iterator() {
-        return myArrayList.iterator();
+        return this.myArrayList.iterator();
     }
 
     /**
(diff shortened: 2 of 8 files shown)


stleary:
Looks good, but seems to be breaking one of the unit tests:
org.json.junit.JSONObjectTest > jsonObjectIncrement FAILED
    java.lang.AssertionError at JSONObjectTest.java:1319
Can you confirm?

stleary:
Looks good, but seems to be breaking one of the unit tests:
org.json.junit.JSONObjectTest > jsonObjectIncrement FAILED
    java.lang.AssertionError at JSONObjectTest.java:1319
Can you confirm?

johnjaylward:
Yup,  I'll take a look.

johnjaylward:
OK,  looks like the test is wrong due to that lack of having a float put method I mentioned. I'll update the test and push the change later today.

stleary:
This is an enhancement, not a bug fix. Some code is refactored and spelling errors are corrected.  Some improvements made to numeric handling.

**Risks**
Very low

**Changes to the API?**
No, other than support for JSONObject put(float)

No

Broke one test due to new improved increment handling. The test code fix https://github.com/stleary/JSON-Java-unit-test/pull/71 will be committed in sync with this one.

Yes

**Review status**
ACCEPTED, Starting 3 day timer.

Issue #334, opened 2017-05-15
Opt performance enhancement


TheMatthew:
Hi, I am currently using your lib to parse json logs that can be very large. I was wondering if it would be possible to make opts have an early failure. The reason I am asking is while 
<code>
try{
return getThing(string);
}catch(exception e){
return defaultValue;
}
</code>

is super readable, the throw catch exception is very slow. It would speed things up a lot having a simple check in front of the field to return if there is no field defined from the get go.

Thanks for such a great and easy to use lib.
:)

johnjaylward:
the try/catch would need to stay in case of exceptions we don't actually plan on. But a pull request to enhance the opt* functions to be more performant would be welcome. Test cases should also be updated if needed in the separate test case project.

stleary:
Sure, you can propose a pull request if you have a code change in mind. Please see the FAQ more guidance on submitting code changes (https://github.com/stleary/JSON-java/wiki/FAQ).  For performance enhancements, it will be helpful if you can provide some timing results to quantify the change.

TheMatthew:
I am checking if I can contribute to the project. Until then, here is an example of code that can do it. It is published in open source under EPL so what you do with it is up to you. ;)

guidomedina:
Hi all, that's a good catch, generally speaking relaying on exceptions for normal code flow is most likely to be slower than doing some checks like in this case if the value is absent.

Throwing exceptions will render stack traces and other things involved which will usually make the code slower by a factor, this can be easily proven, though might be unnecessary as it is a well known fact.

guidomedina:
There are many articles about this but this is one example:

http://stackoverflow.com/questions/1546514/java-exceptions-as-control-flow

johnjaylward:
@guidomedina please don't spam

stleary:
This issue has been superseded by pull request #337, please post future comments there.

PR #337, opened 2017-05-16
Optimizes opt* functions

johnjaylward:
Adjustments to the opt* functions in reference to #334

Notable changes:
* New API functions for `optFloat(key, default)` and `optFloat(key)`
* New API functions for `optNumber(key, default)` and `optNumber(key)`
* New API functions for `getNumber` and `getFloat`
* Update all opt* functions that relied on the get* function throwing an exception to no longer rely on that.

Test speedup over the old implementation is 870s -> 120s for a 2.1 GB file (about 14% of the time; or 86% time saved). See https://github.com/stleary/JSON-java/pull/337#issuecomment-302104179 and https://github.com/stleary/JSON-java/pull/337#issuecomment-302125122

This is a performance enhancement, not a bug fix. Some code is refactored to prevent exceptions from being thrown. Type coercion is also standardized between all the opt* number functions.

**Risks**
low

**Changes to the API?**
New methods added for `optFloat` and `optNumber`. 

No

Broke one unit tests for BigDecimal->BigInteger conversion due to updated coercion rules. Tests has been updated and new tests added for new methods `optFloat` and `optNumber`. See https://github.com/stleary/JSON-Java-unit-test/pull/71 for full changes.

Yes

**Review status**
ACCEPTED, Starting 3 day timer for comments

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -1075,8 +1075,8 @@ public class JSONObject {
      * if there is no value for the index, or if the value is not a number and
      * cannot be converted to a number.
      *
-     * @param index
-     *            The index must be between 0 and length() - 1.
+     * @param key
+     *            A key string.
      * @return The value.
      */
     public float optFloat(String key) {
@@ -1088,8 +1088,8 @@ public class JSONObject {
      * is returned if there is no value for the index, or if the value is not a
      * number and cannot be converted to a number.
      *
-     * @param index
-     *            subscript
+     * @param key
+     *            A key string.
      * @param defaultValue
      *            The default value.
      * @return The value.
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -1068,7 +1068,11 @@ public class JSONObject {
             // jo.optInt("double"); -- will return 1, not an error
             // this conversion to BigDecimal then to BigInteger is to maintain
             // that type cast support that may truncate the decimal.
-            return new BigDecimal(val.toString()).toBigInteger();
+            final String valStr = val.toString();
+            if(isDecimalNotation(valStr)) {
+                return new BigDecimal(valStr).toBigInteger();
+            }
+            return new BigInteger(valStr);
         } catch (Exception e) {
             return defaultValue;
         }
@@ -1759,12 +1763,22 @@ public class JSONObject {
             return false;
         }
     }
-
+    
+    /**
+     * Tests if the value should be tried as a decimal. It makes no test if there are actual digits.
+     * 
+     * @param val value to test
+     * @return true if the string is "-0" or if it contains '.', 'e', or 'E', false otherwise.
+     */
+    protected static boolean isDecimalNotation(final String val) {
+        return val.indexOf('.') > -1 || val.indexOf('e') > -1
+                || val.indexOf('E') > -1 || "-0".equals(val);
+    }
+    
     /**
      * Converts a string to a number using the narrowest possible type. Possible 
      * returns for this function are BigDecimal, Double, BigInteger, Long, and Integer.
-     * 
-     * An Exception is thrown if
+     * When a Double is returned, it should always be a valid Double and not NaN or +-infinity.
      * 
      * @param val value to convert
      * @return Number representation of the value.
@@ -1775,46 +1789,52 @@ public class JSONObject {
         char initial = val.charAt(0);
         if ((initial >= '0' && initial <= '9') || initial == '-') {
             // decimal representation
-            if (val.indexOf('.') > -1 || val.indexOf('e') > -1
-                    || val.indexOf('E') > -1
-                    || "-0".equals(val)) {
+            if (isDecimalNotation(val)) {
                 // quick dirty way to see if we need a BigDecimal instead of a Double
+                // this only handles some cases of overflow or underflow
                 if (val.length()>14) {
                     return new BigDecimal(val);
                 }
-                return Double.valueOf(val);
+                final Double d = Double.valueOf(val);
+                if (d.isInfinite() || d.isNaN()) {
+                    // if we can't parse it as a double, go up to BigDecimal
+                    // this is probably due to underflow like 4.32e-678
+                    // or overflow like 4.65e5324. The size of the string is small
+                    // but can't be held in a Double.
+                    return new BigDecimal(val);
+                }
+                return d;
             }
             // integer representation.
             // This will narrow any values to the smallest reasonable Object representation
             // (Integer, Long, or BigInteger)
+            
+            // string version
             // The compare string length method reduces GC,
             // but leads to smaller integers being placed in larger wrappers even though not
             // needed. i.e. 1,000,000,000 -> Long even though it's an Integer
             // 1,000,000,000,000,000,000 -> BigInteger even though it's a Long
-            
-            // string version
-            if(val.length()<=9){
-                return Integer.valueOf(val);
-            }
-            if(val.length()<=18){
-                return Long.valueOf(val);
-            }
-            return new BigInteger(val);
+            //if(val.length()<=9){
+            //    return Integer.valueOf(val);
+            //}
+            //if(val.length()<=18){
+            //    return Long.valueOf(val);
+            //}
+            //return new BigInteger(val);
             
             // BigInteger version: We use a similar bitLenth compare as
             // BigInteger#intValueExact uses. Increases GC, but objects hold
             // only what they need. i.e. Less runtime overhead if the value is
             // long lived. Which is the better tradeoff? This is closer to what's
             // in stringToValue.
-            
-            //BigInteger bi = new BigInteger((String)val);
-            //if(bi.bitLength()<=31){
-            //    return Integer.valueOf(bi.intValue());
-            //}
-            //if(bi.bitLength()<=63){
-            //    return Long.valueOf(bi.longValue());
-            //}
-            //return bi;
+            BigInteger bi = new BigInteger(val);
+            if(bi.bitLength()<=31){
+                return Integer.valueOf(bi.intValue());
+            }
+            if(bi.bitLength()<=63){
+                return Long.valueOf(bi.longValue());
+            }
+            return bi;
         }
         throw new NumberFormatException("val ["+val+"] is not a valid number.");
     }
@@ -1849,9 +1869,9 @@ public class JSONObject {
         char initial = string.charAt(0);
         if ((initial >= '0' && initial <= '9') || initial == '-') {
             try {
-                if (string.indexOf('.') > -1 || string.indexOf('e') > -1
-                        || string.indexOf('E') > -1
-                        || "-0".equals(string)) {
+                // if we want full Big Number support this block can be replaced with:
+                // return stringToNumber(string);
+                if (isDecimalNotation(string)) {
                     Double d = Double.valueOf(string);
                     if (!d.isInfinite() && !d.isNaN()) {
                         return d;
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -540,7 +540,7 @@ public class JSONObject {
             return new BigInteger(object.toString());
         } catch (Exception e) {
             throw new JSONException("JSONObject[" + quote(key)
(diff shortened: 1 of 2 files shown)


TheMatthew:
Tested on a 2.1 Gb json file on an i7-620 with an ssd. 
Before parsing with lots of extra logic: 870 s
After: 120 s

This is in a real-life application. I would assume that micro-benchmarks would yield a gain of at lease 20x.

All in all, this is great, thanks!

johnjaylward:
Is the 870s the original JSONObject implementation with no pre-checks, or the modified one you linked in the issue that uses the `has(key)` calls? I'd just like to verify what we are looking at.

TheMatthew:
Sorry for the delay

No changes origin/master: 870s
has(key): 122s
your change: 120s

johnjaylward:
Thanks @TheMatthew . This PR should be feature complete now. I'll update the description to reflect all the changes.

johnjaylward:
Tests updated and PR description updated to match testing changes.

stleary:
Nice, performance improved and code made more readable. Accepted pending 3 day timer.

guidomedina:
@johnjaylward 
Do you know how to squash your commits? will make it much easier now and in the future to code review.

johnjaylward:
@guidomedina it's easy to code review without erasing history. Please stop spaming

guidomedina:
1st time I have been told to "stop spamming" on Github, I only made 3 comments, valid comments so not sure what you mean.

johnjaylward:
The comments on squashing commits are not relevant to the pull request nor helpful to the review. This is not a discussion board for the merits or pitfalls of rebasing.

guidomedina:
It was only one comment, not need to tell someone not to spam because you don't like the comment, that's not very polite, you could simple state your reasons as you just did.

Did I spam when I support the exception comment? You just made my day dude, this is my last comment on your issues anyway.

JSONObject.java
```diff
@@ -578,10 +581,10 @@ public double getDouble(String key) throws JSONException {
         Object object = this.get(key);
         try {
             return object instanceof Number ? ((Number) object).doubleValue()
-                    : Double.parseDouble((String) object);
+                    : new BigDecimal((String) object).doubleValue();
```
guidomedina: Introducing a `BigDecimal` to parse a `double` will add too much overhead for no reason, if you try to represent a `double` from a `String` it will stay as is, if it was indeed a `double` then nothing will be gained or lost, you are only adding more GC overhead.

Same thing for your other `BigDecimal` introduction when working with `float` or `double`

johnjaylward: No, `Double.parse` will throw an exception if the number can not be represented as a double.
From the javadocs:

> double java.lang.Double.parseDouble(String s) throws NumberFormatException
> 
> Returns a new double initialized to the value represented by the specified String, as performed by the valueOf method of class Double.
> Parameters:s the string to be parsed.Returns:the double value represented by the string argument.Throws:NullPointerException - if the string is nullNumberFormatException - if the string does not contain a parsable double.Since:1.2See Also:java.lang.Double.valueOf(String)
> 
> 

`Double.valueOf` seems to do something closer to what you are trying to accomplish, but it would be the same amount of garbage collection.

`new BigDecimal(String)` will always parse the value as long as it's a valid decimal number (i.e. not "abc"). The `doubleValue` call will then do a narrowing conversion which may change the value to -Inf or +Inf.

Lastly, this change would be out of scope for this PR. the `get*` methods were not modified here, only the opt*. If you feel we should evaluate it further, feel free to open another pull request with the changes and notes to performance impacts.

guidomedina: `new BigDecimal(...)` and `Double.parseDouble(...)` have the same behavior, an input that fails for `parseDouble` will also fail with new BigDecimal, the JavaDoc have different wording but the end result will be the same.

As for relevance, you are trying to make opt* operations faster and yet you are introducing an expensive and unneeded instantiation when the return value is a native (small d)

johnjaylward: The JavaDoc is what to go by. If someone is running on a different VM with a different implementation, they may not function the same.

The BigDecimal conversion is not that expensive. I'd like to see some stats on it before making a change like that. We have no guarantees on the input. Just because someone asks for an `int` does not mean I would want to use `Integer.parse`.

johnjaylward: Also, I'm not "introducing" the big decimal conversion. It was there before the refactor.

guidomedina: I doubt that, things are quite clear in the Java Specification, anyone that writes a JVM must follow these rules, and most likely will use native calls:

`public static native double longBitsToDouble(long bits);`

I recommended that for `double` and `float`, it is an unnecessary instantiation.

as for `int` and `long` there are waaaaay faster and cheap ways:
- `(int)Double.parseDouble(x)`
- `(long)Double.parseDouble(x)`

johnjaylward: as I said. please post a followup pull request.

johnjaylward: * (int)Double.parseDouble(x)
* (long)Double.parseDouble(x)

do not perform the same as `new BigDecimal(x).*Value` for values larger than 2^53.

guidomedina: Which you then convert to `double`, good job then truncating it into a different number.

guidomedina: but then `int` typecast would be wrong for my proposed shortcut but it can be easily mitigated by checking if the double value is between `Integer.MIN_VALUE` and `Integer.MAX_VALUE` and if it isn't throw an exception.

johnjaylward: No, it's an opt* function. The entire point is that it does not throw an exception. I'm not disagreeing that the parsing performance can be improved. But I think it needs more than a cursory glance and wishful thinking.

This PR is focusing strictly on the request made in #337 which was to mitigate the Exception handling impact.

Any parsing impacts should be in a new PR after this one is merged.

Any narrowing coercion has the potential to change the number. This is a given and also why I added the optNumber call to the API.

guidomedina: Narrowing will not change the double for long, maybe for int, but then it is the same for BigDecimal calls, the reason I commented into this particular change is because of what I'm seeing:
```
-                    : Double.parseDouble((String) object);
+                    : new BigDecimal((String) object).doubleValue();
```

Am I reading this incorrectly? I'm seeing `Double.parseDouble(...)` changed to `new BigDecimal(...)` in this commit.

Narrowing will cause loss always, but the same will happen with `BigDecimal` hence my point of not bothering with such overhead at all.

johnjaylward: let me check that. If I did change it, I'll likely change it back.

johnjaylward: Yes, I must have changed that by accident when I was testing stuff and didn't notice. I'll revert that change back and update the respective optDouble and optFloat calls that I copied that to.

As for the narrowing, it is handled differently between BigDecimal and Double. Also, not all Long values fit in a double. see the test case changes @ https://github.com/stleary/JSON-Java-unit-test/pull/71 that I updated today. A Double only has 53 bits for it's mantissa which will not let any value over 2^53 be represented correctly.

johnjaylward:
@stleary I added the optFloat/optNumber, but no associated getFloat/getNumber. Should I add them in for posterity?

stleary:
@johnjaylward Sure, in general when an opt* method is added, a corresponding get*method is a good idea.

johnjaylward:
@TheMatthew can you retest the latest commit of this? I want to ensure the new pull out of the number parsing functions didn't adversely affect the performance of your test case.

JSONObject.java
```diff
@@ -1759,12 +1763,22 @@ public boolean similar(Object other) {
...
+     * Tests if the value should be tried as a decimal. It makes no test if there are actual digits.
+     * 
+     * @param val value to test
+     * @return true if the string is "-0" or if it contains '.', 'e', or 'E', false otherwise.
+     */
+    protected static boolean isDecimalNotation(final String val) {
+        return val.indexOf('.') > -1 || val.indexOf('e') > -1
+                || val.indexOf('E') > -1 || "-0".equals(val);
+    }
+
```
guidomedina: Nice "quick and dirty" function, what happens if running on a server on a different locale?

Anyway the following example can help, I'm guessing:
```
  static final char MINUS_SIGN;
  static final char DECIMAL_SEPARATOR;
  
  static {
    final DecimalFormatSymbols currentLocaleSymbols = DecimalFormatSymbols.getInstance();
    MINUS_SIGN = currentLocaleSymbols.getMinusSign();
    DECIMAL_SEPARATOR = currentLocaleSymbols.getDecimalSeparator();
  }
```

guidomedina: Though I'm not sure what should be done if the decimal separator for the server locale is a comma, not sure what the parse function will behave, might not be worth the effort in supporting this.

johnjaylward: We are not looking for any number, we are looking for JSON numbers, which are not locale dependent.

guidomedina: might be worth seeing how `new BigDecimal` and `parseNumber` check what's the decimal separator, the idea would be to have your function to be consistent with what Java does already, but again, maybe low priority, maybe not worth the effort.

If you run JVM in a French server for example, `Double.parseDouble` could expect a comma instead of dot, but I'm not sure of it.

johnjaylward: from BigDecimal: 
```java
                    } else if (c == '.') {   // have dot
                        // have dot
                        if (dot) // two dots
                            throw new NumberFormatException();
                        dot = true;
```

Double.parse is similar.

guidomedina: Found it, ignore this completely, `FloatingDecimal` line 224:

```
if(var4 != 46) {
  break;
}
```

TheMatthew:
Sorry, I was in a huge rush, I will retest.

TheMatthew:
Before this patch AKA current git head: 870 s
My suggested simple patch: 122 s
@johnjaylward 's first patch: 120 s
@johnjaylward 's current patch: 126 s

stleary:
Performance still looks good. @johnjaylward let me know when you are ready to proceed with the pull request.

TheMatthew:
Ok, I started profiling. the code in this commit looks moderately tight.

Some low hanging fruits I saw: 
JSONObject.keys() should probably never be used, rather it would be better to do an enhanced for loop on JSONObject#keySet()

This brings the execution time down from 126 s to 115. 

Then if you have an array let's say called "CHARS_TO_APPEND" which has 256 values, Like this:
        private static final boolean CHARS_TO_APPEND[]= new boolean[256];
	static{
		Arrays.fill(CHARS_TO_APPEND, true);
		for(int i = 0; i < ' '; i++){
			CHARS_TO_APPEND[i]=false;
		}
		CHARS_TO_APPEND[',']=false;
		CHARS_TO_APPEND[':']=false;
		CHARS_TO_APPEND[']']=false;
		CHARS_TO_APPEND['}']=false;
		CHARS_TO_APPEND['/']=false;
		CHARS_TO_APPEND['\\']=false;
		CHARS_TO_APPEND['"']=false;
		CHARS_TO_APPEND['[']=false;
		CHARS_TO_APPEND['{']=false;
		CHARS_TO_APPEND[';']=false;
		CHARS_TO_APPEND['=']=false;
		CHARS_TO_APPEND['#']=false;
	}

and replace the while loop in JSONTokener#nextValue() with 

    while (CHARS_TO_APPEND[c]) {

you reach 113s.

Here is a quick profile I did. It can be opened with visualvm 1.3.9 .
[JSON-java.nps.zip](https://github.com/stleary/JSON-java/files/1016623/JSON-java.nps.zip)
The jist of it is that most of your time is now spent in the tokener, which is good and the reader, which is excellent. 

Hope this helps.

TheMatthew:
For transparency purposes I've shared this github benchmark https://github.com/TheMatthew/json-benchmark

it's under apache so it should be compatible for any copy-pasted tests

Initial results:

org.json.simple: 140s,
org.json: 870s
after the patch: 126s

johnjaylward:
@stleary I agree, this still looks good. Let's call this one done. I just pulled in the changes recently merged to master into this PR.

@TheMatthew Lets take that optimization to a separate PR after this one is merged.

Issue #269, opened 2016-08-11
Disable trailing commas


erosb:
Currently the parser permits trailing commas after the last key-value pair of an object. For example the following statement: `new JSONObject("{\"a\":1,}");` doesn't throw any exceptions. As far as I know trailing commas are not permitted in json.

johnjaylward:
As far as I've seen, most JavaScript implementations do allow the trailing comma. I think this is similar to #240 where a request was made to be more strict.

Maybe a better option would be to create a "strict" parser option that validated the input matched actual JSON spec. If we had a flag for this, it would fix this request, as well as #240.

By default the parser would use the existing lose matching I'd think.

stleary:
@erosb thanks for the suggestion. Technically, I think you are correct. But in this case it may be better for the implementation to be more forgiving than the spec.

PR #338, opened 2017-05-17
A fix for Issue 296

effad:
This fix provides additional API for creating JSONObject and JSONArray instances in the way requested by myself in Issue 296: Java null should become JSON null. 

The behaviour of the existing API (i.e. without the useJavaNullAsJsonNull) will not be changed by this fix.

I also created some regression tests in the JSONArrayTest and JSONObjectTest from the JSON-Java-unit-test repository which I attach as ZIP-File.

[Tests.zip](https://github.com/stleary/JSON-java/files/1006845/Tests.zip)


johnjaylward:
would you be able to create a PR for the tests as well under the test project:

https://github.com/stleary/JSON-Java-unit-test

johnjaylward:
Overall these changes are straightforward and match what I expected in #296. I think if you can get the changes to the tests as a PR instead of a zip, you'll have a good case for the change. I personally have no issue with the API changes.

JSONObject.java
```diff
@@ -1797,7 +1856,7 @@ public static Object wrap(Object object) {
             }
             if (object instanceof Map) {
                 Map<?, ?> map = (Map<?, ?>) object;
-                return new JSONObject(map);
+                return new JSONObject(map, useJavaNullAsJsonNull);
```
johnjaylward: you should also update lines 1852 and 1855 to use the new JSONArray constructors:
```java
if (object instanceof Collection) {
                 Collection<?> coll = (Collection<?>) object;
                 return new JSONArray(coll, useJavaNullAsJsonNull);
             }
             if (object.getClass().isArray()) {
                 return new JSONArray(object, useJavaNullAsJsonNull);
              }
```

stleary:
Would this be a first use in this lib of a boolean constructor param to control behavior?

johnjaylward:
In these classes. The XML class uses a similar control, although I'm not a huge fan of the design pattern (I did it first in the XML class).

In another ticket I mentioned that using the boolean flags may have been a bad choice, and instead we should have a configuration object.

```java
public class XMLConfig {
   public boolean someConfigOption;
   public boolean someOtherConfigOption;
}
```

maybe to future proof this change, we should consider a similar concept.

```java
public class JSONParseConfig {
   public boolean useJavaNullAsJsonNull;
   public JSONParseConfig () { super(); }
   public JSONParseConfig setUseJavaNullAsJsonNull(final boolean useJavaNullAsJsonNull) {
      this.useJavaNullAsJsonNull = useJavaNullAsJsonNull;
      return this;
}
```
The setter would return the config object so you could chain configuration options:

```java
new JSONObject(someJsonString,
   new JSONParseConfig().setUseJavaNullAsJsonNull(true).setSomeFutureOption(true)
);
```

stleary:
Control params are arguably not a great design pattern, but the greater concern to me are the potential follow up requests for special handling of numeric types, XML, and so forth. A config object might work better, but might also open a Pandora's box of future customizations. It could be that forking the project is the best solution in this case.  Otherwise, how far are we willing to depart from being a simple reference app? Let me know what you think @effad and @johnjaylward.

effad:
I can see and understand the concern about people requesting future customization. And yes, forking is an absolutely viable way of getting the desired behaviour into a library.
However, given the very widespread use of JSON-java, I think it is also a problem if loads of forks exist because it will require people to apply the same changes over and over again to the detriment of each fork. (As could be seen here: My "fix" required careful review and a second iteration to be (hopefully) correct.)

In the special case of Issue 296 I would argue that it actually fixes/addresses a (admittedly very old) regression. The behaviour brought back here is that of a very old version of the library and people (like me) may need to have this behaviour while at the same time not wanting to write/maintain their own version of the library.

As to boolean parameter vs. configuration object: I don't have a very strong opinion there. Both ways are ok for me and I can see the advantage of extensibility in a JSONParseConfig-object.

johnjaylward:
I think we should accept this change, either as-is or swapped with the config-object. The original issue (https://github.com/stleary/JSON-java/issues/296#issuecomment-255938989) @effad opened explained it pretty well, and I think it's a valid use case for the parser/object mapper.

Using the config object could possibly also bring forward support for strict parsing like mentioned in #240 and #269. 

I'm not sure we necessarily want to support that per-say, but I think we could handle that on a PR-by-PR basis.

This PR looks simple enough and I think if @effad could post those test changes that are currently in a zip to the test project as a PR it would be easier to check and easy enough to support.

johnjaylward:
Ah, I found the tests here: https://github.com/stleary/JSON-Java-unit-test/pull/72

stleary:
@effad This project already has > 1900 forks, currently averaging 1 every day or so. Can't agree that the current behavior is a regression, since it seems clearly intended to be a design change.  But I think your other points have merit.
@johnjaylward Did you want to address the question of how far are we willing to depart from being a simple reference app? Do you not consider that a priority? I need a little time to get up to speed on your comments about #240 and #269.

johnjaylward:
For the remarks as a reference app, I feel this issue still falls in that purview. In general would like to see this project stay small, but also useful to the most people out there. If someone needs something full featured with all the bells and whistles, that's what projects like https://github.com/FasterXML/jackson are for.

Being able to maintain the "null" data in their JSON representations seems like a valid thing a reference app should handle to me. I guess what it comes down to is a difference in opinion on which features should be considered "reference". Kinda makes me wish GitHub had a talk page like wikis 😄.

My main argument for the configuration object is in case there are other "reference" items we may wish to support. The "Strict parsing" like mentioned in those other 2 tickets is what came to mind. Strict parsing seems very much like something a reference library would handle. I personally like the loose parsing and wouldn't use the strict parsing, but others have requested it.

stleary:
Fair enough, the team agrees that this is a worthwhile enhancement, so a code change can proceed, within the constraints of the project. Would it be accurate to say the entire pull request comes down to deciding whether to call jsonOjbect.wrap()?  If so, it would be better if we can come up with something that does not require passing control parameters or config objects down to the places where the code is actually needed.  New proposed code can be updated in this pull request, or a new PR started.

Not inclined to consider support for strict parsing until/unless @douglascrockford indicates that this would be a good direction to take the project.

stleary:
Thanks for the pull request, but can't be accepted at the present time.

Issue #340, opened 2017-05-22
Performance & Code Style, avoid passing iterators around


TheMatthew:
Hi,
I have an issue with the JSONObjecy#keys() method. It states it is an iterator, though the name implies it is an access to a collection. I have also noticed that is it used often in a while(iter.hasNext){foo = iter.next; ...) pattern. This is not as optimal as using an enhanced for loop on a JSONObject#keySet(). It is murky as to whether or not I am allowed to contribute code yet, but if you replace all #keys accesses by #keyset for loops you can go from parsing a 2 GB file in 125 seconds to 115. this is not an order of magnitude like @johnjaylward 's #337  patch, but it's still pretty decent.

One could argue the reason for doing the iterator is to avoid issues with concurrent modifications. I would argue that a readwritelock would be a safer approach in general.

Thanks for making such a great product by the way.

johnjaylward:
The history of those loops is from pre-java5 when the enhanced for loop was not available. I think they just haven't been updated due to them working and no complaints from performance. I don't think it should be an issue to update these.

stleary:
Everyone is welcome to submit pull requests.  Since the project is in maintenance mode, there are limits to what kind of changes are accepted. Every pull request carries a risk of introducing regressions, or negatively impacting existing users.  Please see the FAQ for more information: https://github.com/stleary/JSON-java/wiki/FAQ.

johnjaylward:
@stleary, I believe he was implying he wanted permission from his company before spending company time on it.

I opened a PR for the refactor. Let me know if it meets your needs. It's currently based off of Master and not my other PR that is scheduled for merge.

stleary:
This issue is addressed in pull request #341. Please post any future comments on that thread.

PR #341, opened 2017-05-22
Optimize loops

johnjaylward:
Changes for #340 

* Adds protected entrySet accessor to JSONObject
* Updates loops that request key/value pairs to use the new entrySet accessor

This is a performance enhancement, not a bug fix. Some code is refactored to use new loop types and map iterators introduced in Java5. One assignment line was deleted from the XML class as it didn't do anything.

**Risks**
low

**Changes to the API?**
New protected method added for internal access to the backing map entrySet. I wasn't comfortable making this a public method without input from the community as it allows unfiltered access to the entire map (keys and values).

No

No. Also, no unit tests have been added at this time.

Yes

**Review status**
NOT REVIEWED


johnjaylward:
update the branch to the latest master

johnjaylward:
@thematthew please review when convenient

TheMatthew:
I will asap, thanks @johnjaylward

TheMatthew:
The performance gain I am seeing is quite good. it is passing from 125 seconds to 115 seconds for a 2.1gb file. it is even more interesting if the file is already in ram, it goes from ~98 seconds to 85.

stleary:
Looks good, comment timer started.

PR #346, opened 2017-06-06
Using LinkedHashMap instead of HashMap to keep the order of elements

ttulka:
A test case could be:

@Test
public void testIfOrderOfElementsRemainsSame() {
    String json =
"{\"c\":{\"c\":\"c\",\"a\":\"a\",\"b\":\"b\"},\"a\":{\"c\":\"c\",\"a\":\"a\",\"b\":\"b\"},\"b\":{\"c\":\"c\",\"a\":\"a\",\"b\":\"b\"}}";
    Assert.assertEquals(json, new JSONObject(json).toString());
}

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -34,7 +34,7 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.Collection;
 import java.util.Enumeration;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Iterator;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


MinnDevelopment:
**Why isn't ordering allowed in JSONObjects?**

Support for ordering is one of the most popular requests, along with adding some type of build support. However, from the JSON spec [RFC 7159](https://tools.ietf.org/html/rfc7159): "An object is an unordered collection of zero or more name/value pairs". As a reference app, the design is required to stay as close to the spec as possible, so ordering in JSONObjects is not supported.

**Will ordering ever be allowed in JSONObjects?**

There are no current plans to support ordering. It is not expected that this will ever change.

>Taken straight from [FAQ](https://github.com/stleary/JSON-java/wiki/FAQ)

ttulka:
Okay, thanks for the info.

MinnDevelopment:
Maybe it would be a good idea to place a comment in the code explaining that HashMap is chosen on purpose

stleary:
@MinnDevelopment  Helpful code comments are always welcome, feel free to propose a pull request.

PR #347, opened 2017-06-08
a comment added to explain the use of HashMap

ttulka:
to avoid misconception of contributors about using HashMap to implement a JSON object as a unordered collection by the definition

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -164,7 +164,12 @@ public class JSONObject {
      * Construct an empty JSONObject.
      */
     public JSONObject() {
-        // HashMap is used on purpose to ensure that elements are unordered
+        // HashMap is used on purpose to ensure that elements are unordered by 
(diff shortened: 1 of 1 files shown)


stleary:
Since we are going to document this in the code, let's include a thorough explanation, e.g. what design goal is met by using an unordered container and why an ordered container is specifically not used. You can use the text from the Wiki, although I think your own interpretation could be better. This is a good example of how not all code is inherently self-documenting.

ttulka:
I tried my best. There is actually not a lot of information about this issue and I see several misconceptions in the discussion forums. So, the explanation is very welcome, I guess :-)

stleary:
Comment-only change

**Risks**
None

**Changes to the API?**
No

Can be rolled into the next release

No

No

No

**Review status**
**ACCEPTED** Starting 3 day timer.

PR #352, opened 2017-06-23
Error message position fixes

johnjaylward:
**Key Changes:**
* Fix error message position information to be accurate

Corrects incorrect position information when generating syntax errors in tokeners.

**Risks**
None. The current position information is wildly off in some cases. It's not helpful to anyone trying to debug a failing document.

**Changes to the API?**
none.

Yes, this is a bug fix that should probably be corrected as soon as possible

No

Yes. The current position information in the XML error messages is wrong/inaccurate. The changes here in regards to the `JSONTokener.more` function have fixed many of the position problems in the old code. Some other pieces of code needed to be updated as well.

See the corrected tests here: https://github.com/stleary/JSON-Java-unit-test/pull/74

Yes. The `JSONToken.more` function was changed to use the mark/reset support in the stream reader instead of stepping forward and backward. This gives us faster checking on `more()` as well as more accurate position information as we aren't moving our index around every time we want to see if there is more information in the stream.

Also, a new field `characterPreviousLine` was added to track the previous line's character count in order to preserve character information when stepping back at a new line mark.

**Review status**
ACCEPTED, starting 3 day comment window.

### DIFF
--- a/CDL.java
+++ b/CDL.java
@@ -22,7 +22,7 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
-*/
+ */
 
 /**
  * This provides static methods to convert comma delimited text into a
@@ -70,9 +70,12 @@ public class CDL {
                 c = x.next();
                 if (c == q) {
                     //Handle escaped double-quote
-                    if(x.next() != '\"')
-                    {
-                        x.back();
+                    char nextC = x.next();
+                    if(nextC != '\"') {
+                        // if our quote was the end of the file, don't step
+                        if(nextC > 0) {
+                            x.back();
+                        }
                         break;
                     }
                 }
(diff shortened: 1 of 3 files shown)


johnjaylward:
@stleary any chance to review this yet? No rush (I know it's holiday time for US and Canada), just wanted to make sure it wasn't missed or forgotten.

stleary:
Thanks for the reminder, holidays and weekends are actually the source for a lot of my free time :) Agree the bug fix should be committed. Any chance you can separate that code from the formatting changes? 
Per the FAQ, code formatting changes are normally not accepted.  I was trying to make an exception in this case, but as you noted, no progress has been made.

johnjaylward:
sure, I think I can just revert that one commit

johnjaylward:
the revert was clean. Should just be these changes now.

stleary:
Thanks for backing out the reformatting changes.

johnjaylward:
no problem. If possible though, I would like the code to be more consistent. Right now the mixed spaces/tabs and the inconsistent indentation make it harder to contribute. If we could settle on something then reformat to match, that would be very helpful.

johnjaylward:
@stleary also, I'm wondering if I should back out some of the recent changes I made that make it even harder for Android compatibility before doing a release. Not all the change in #350 , but specifically the ones for `entrySet` and maybe some others I can't think of off the top of my head. Or should we just wait for #350 to get tested and merged before worrying about android?

It does looks like my change to the Android JSONException class are getting added to Android Master soon, but still no guarantee that it will be released in a timely fashion for Android Developers to use.

stleary:
Sure! How does indent 4 spaces, no tabs sound? A commit that contains only formatting changes for one or more files, and passes unit tests,  would be fine.

stleary:
It seems to me that an actual Android developer who uses JSON-Java is needed to Beta test #350, and then recommend approval or further changes.

Issue #279, opened 2016-08-26
JSONObject#populateMap() calls methods that aren't valid object getters


run2000:
JSONObject.populateMap() can be a over-enthusiastic in calling methods on an object that aren't valid bean getter methods. For instance:

* It will call static get methods that otherwise pass all the tests for validity.
* It will call get methods with a void return type
* It will call a get method that is a synthetic method -- typically the result of overriding a superclass

My opinion is that populateMap() should be more cautious about calling methods on an unknown object. There may be side-effecty behaviour within the object that's undesirable. Code to avoid doing this is trivial and fast.

stleary:
I think the idea of more careful handling of beans is reasonable. 

Don't know if we can be sure that no users will ever want to serialize static data. For example, suppose there is only one instance of the class, and its data is static?

Making sure the method has a return type sounds reasonable.

Not sure why synthetic methods must be excluded from serialization. Can you give an example?

Might also want to address enum serialization as a bean, which I think picks up some unwanted getters.

johnjaylward:
I think this ties in a little with #264.

For Statics:
 I can't really see the need to serialize static methods at all. By definition static methods indicate an application level state (or constants that never change) and outputting those as part of the bean state seems like a bug to me.

Void return types:
 I agree, these should be ignored. These likely are causing side effects in the objects

Synthetic Methods:
 My reflection may be a little rusty, but I'm not sure I see why we should ignore these. I agree with @stleary that a use-case for ignoring these would be helpful.

johnjaylward:
@stleary enum serialization was corrected in PR #271. It should not be output as a bean anymore.

erosb:
In a slightly broader context, I have doubts about if supporting bean <-> JSONObject conversions is a good idea at all. Maybe, for the sake of simplicity, `org.json` should support only JSONObject <-> Map conversions and let people perform the more complex task (map <-> bean) perform with a library (eg. apache beanutils) which is better suited for this purpose.

Maybe it would be a better idea to deprecate the `JSONObject(Object)` constructor instead of maintaining it.

johnjaylward:
@erosb I personally do not use the constructor. I've been using my own customer bean -> JSONObject wrapper for around 10 years now. Mine does 2 of the 3 checks (no statics, and no void returns) mentioned above.

However, I think due to the handover agreement with Douglas, we can not deprecate the constructor. Bug fixes should be fine in it, but I don't think we are allowed to reduce the public API.

[...]

Issue #264, opened 2016-08-10
StackOverflow in JSONObject


auricgoldfinger:
I noticed this when updating one of the libraries in our project. I used to serialise a dto which extends another one. This super class has a getter returning an Apache `FastDateFormat`, which is a singleton and returning a static instance of itself.

Previously this wasn't an issue because the JSONObject (JSONArray in my case) took a boolean specifying whether to `includeSuperClass` or not. By default, that was `false` and only the child class was serialised. The parameter is now gone and the default behaviour is that the super class is serialised as well. Hence my issue came up.

Given this class
```java
public class StackOverflow {

   public static void main(String[] args) {
    //  final org.apache.commons.lang.time.FastDateFormat fdf = org.apache.commons.lang.time.FastDateFormat.getInstance();
      final MyDto dto = new MyDto("myvalue");
      new org.json.JSONObject(dto);
   }
}
```

Which uses these
```java
public class MyDto {

   private String value1;
   private StaticMethodClass smc = StaticMethodClass.getInstance();

   public MyDto(String value1) { this.value1 = value1; }
   public String getValue1() {  return value1;  }
   public void setValue1(String value1) {   this.value1 = value1;  }
   public StaticMethodClass getSmc() { return smc; }
   public void setSmc(StaticMethodClass smc) {  this.smc = smc; }
}

public class StaticMethodClass {

   private static StaticMethodClass _instance;

   private StaticMethodClass() {}
   
   public static StaticMethodClass getInstance() {
      if (_instance == null) _instance = new StaticMethodClass();
      return _instance;
   }
}
```

Serializing the `getInstance` method will cause a stack overflow in JSONObject since it keeps being wrapped in JSONObjects (In theory, you can keep doing StaticMethodClass.getInstance().getInstance().getInstance().getInstance()...)

Of course, serialising an instance of StaticMethodClass would cause a Stackoverflow as well.

johnjaylward:
I don't recall ever seeing a constructor for JSONObject or JSONArray that took a boolean to ignore super classes. Do you recall what version you updated from? I took a look at the history of JSONObject but didn't see that constructor in any of the versions I looked at.

johnjaylward:
Also, that StaticMethodClass in your example is not thread safe (in case you need thread safety)

johnjaylward:
[...]

PR #354, opened 2017-07-08
Updates for populateMap based on discussion in #279 and #264

johnjaylward:
**Key Changes:**

* Updates internal `populateMap` function to be more strict about what methods it calls.

Fixes issues as noted in #279 and #264. Namely, this changes populateMap to now exclude the following:
* Methods that return `void`
* [Bridge methods](https://docs.oracle.com/javase/tutorial/java/generics/bridgeMethods.html)
* Static methods

I also tightened up the exception handling in the private method to only handle the checked exceptions. This should make debugging failed serializations easier as runtime exceptions will be propagated to the caller instead of creating an empty JSONObject.

**Risks**
Low. In general the method types now excluded should generally not be serialized. Even though in #279 it was noted that static should not be included, I did run unto an issue where serializing statics lead to a stack overflow as well.

Also, as noted in multiple places, static properties are considered part of a `class` not an `instance`. References:
* http://javabeginnerstutorial.com/core-java-tutorial/transient-vs-static-variable-java/
* https://stackoverflow.com/questions/11000975/are-static-variables-serialized-in-serialization-process
* https://stackoverflow.com/questions/1008023/how-to-serialize-static-data-members-of-a-java-class

I believe keeping the current implementation of serializing static members would actually lead to unexpected behavior.

The argument @stleary made [here](https://github.com/stleary/JSON-java/issues/279#issuecomment-242755649) for static methods doesn't hold up for me. If a static singleton class needs to be serialized, it should hopefully be easy enough for a developer to handle that as a one off case instead of the norm.

**Changes to the API?**
None

No

No. All unit tests should pass. New unit tests were created that will fail against the current master, but pass with this PR. See https://github.com/stleary/JSON-Java-unit-test/pull/75

Yes, the private JSONObject.populateMap function

**Review status**
ACCEPTED

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -28,6 +28,7 @@ import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
 import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.math.BigDecimal;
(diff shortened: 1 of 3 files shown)


JSONObject.java
```diff
@@ -276,16 +277,20 @@ public JSONObject(Map<?, ?> m) {
...
+     * </p><p>
      * For example, if an object has a method named <code>"getName"</code>, and
      * if the result of calling <code>object.getName()</code> is
      * <code>"Larry Fine"</code>, then the JSONObject will contain
      * <code>"name": "Larry Fine"</code>.
-     *
+     * </p><p>
+     * Methods that return <code>void</code> as well as <code>static</code>
+     * methods are ignored.
+     * </p>
```
MinnDevelopment: There is no need for `</p>` close tags in javadocs

johnjaylward: but I like them :) and it still works. Plus Java9 javadoc will support HTML 5 which requires the close tags.

johnjaylward: ah, I see the javadoc command in java9 will instert the close tags for you. I'll get rid of them

stleary:
@johnjaylward I am on board with this. Cleaning up bean serialization is reasonable, and stack overflows that result from serialization must never occur. However, it would be better if, in each of the cases mentioned, a unit test is provided that causes incorrect processing that is fixed by the pull request. Do you agree, and is it practical to do this?

johnjaylward:
That sounds good and relatively practical.

Would you mind if I reorganized the classes in the test project a little? It's hard to quickly pick out the test classes from the classes used for mocking and testing. I'm thinking a sub package called `org.json.testdata` or maybe `org.json.junit.data` and we can put the `MyBean`, `MyEnum` etc classes there instead of having everything the test package itself.

johnjaylward:
@stleary I updated the test cases for supporting these changes. I believe I have decent coverage, but was unable to get the "isBridge" check in populateMap to get hit.

If someone else could look at my GenericBean and GenericBeanInt classes/tests and let me know what I'm missing there it would be appreciated. I thought what I did would cover the isBridge check, but I apparently missed something I'm not understanding to get the compiler to generate a bridge method. Or maybe I am, but the bridge method is not public so it's being ignored by a previous check in the same `if` statement...?

johnjaylward:
Also, I updated the `populateMap` method to check if the return type implements `Closeable`. If it does I close the resource. If anyone objects to this let me know.

johnjaylward:
Ok, I figured out the bridge method. I just needed to override the getGenericValue method

stleary:
Sure, cleanup/reorg of the unit tests is fine.

stleary:
Starting 3 day comment timer.

PR #355, opened 2017-07-11
Added unsigned number parsing for int types

MinnDevelopment:
I chose to use `getXUnsigned` as naming convention to have better suggestions when using auto-completion in an IDE.

I am aware you could simply use `getString(x)` and provide it to `Long.parseUnsignedLong` but I found this to be more convenient.

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -368,10 +368,63 @@ public class JSONArray implements Iterable<Object> {
      *             If the key is not found or if the value is not a number.
      */
     public int getInt(int index) throws JSONException {
+        return getInt(index, 10);
+    }
+
+    /**
+     * Get the int value associated with an index.
+     *
+     * @param index
+     *            The index must be between 0 and length() - 1.
+     * @param radix 
+     *            The radix to use when parsing from String
+     * @return The value.
+     * @throws JSONException
+     *             If the key is not found or if the value is not a number.
+     */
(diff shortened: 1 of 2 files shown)


johnjaylward:
While I don't necessarily object to the change, it does require java 1.8 to compile. I know that other users of the library are still requiring java 1.6 for some use cases which is what we compile against as a baseline.

If we think it's time to move away from 1.6/1.7 for the library, then this should be fine. But I don't think we've had that conversation yet.

MinnDevelopment:
Fair enough. I can keep this Pull Request open in the meantime as a reminder for the future.

I wasn't aware that these methods were introduced in 1.8 - I assumed them to have been introduced in 1.7

johnjaylward:
I updated the [FAQ](https://github.com/stleary/JSON-java/wiki/FAQ) with links to the issues and PRs that previously covered why we went down to 1.6 . It appears that about 9% of the [android user-space](https://developer.android.com/about/dashboards/index.html) is still on 1.6 (API < 19) while around 10% of the android user-space is at 1.9 (API > 23). That leaves about 79% of android users on 1.7.

While I doubt Android is the largest use of our library,  it seems a good chunk is still on 1.6.

stleary:
Thanks for the pull request. Feel free to resubmit if/when the project upgrades to 1.8 (no plans to do that at the present time).

Issue #356, opened 2017-07-15
JSONObject(2);


RMelanson:
JSONObject(2) does not indent for example:
    System.out.println("jsonObject.toString(2) = \n"+jsonObject.toString(2));
Prints Out:
{"TABLE":{"yhoo":{}}}

Should be:
{
   "TABLE":{
      "yhoo":{
      }
   }
}

stleary:
Please provide a complete example. Also, in the 'should be' section, use markdown styling to retain your preferred spacing (use 3 grave accent AKA backtick  ` chars in a row, before and after your output)

RMelanson:
Do you have any example?  I worked on this all day?

Thanks
Robin

johnjaylward:
<pre>```java
public static final void someFunction(){
    // do something
}
```</pre>
generates
```java
public static final void someFunction(){
    // do something
}
```

<pre>```json
{
    "int": 0,
    "string": "string",
    "bool": false
}
```</pre>
generates
```json
{
    "int": 0,
    "string": "string",
    "bool": false
}
```

johnjaylward:
@stleary Here is a test case:

```java
JSONObject jo = new JSONObject().put("TABLE", new JSONObject().put("yhoo", new JSONObject()));
assertEquals("\"TABLE\": {\n  \"yhoo\": {\n  }\n}", jo.toString(2));
```
and it currently fails like so:
```
org.junit.ComparisonFailure: expected:<["TABLE":{
  "yhoo":{
  ]}> but was:<[{"TABLE": {"yhoo": {}}]}>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.json.junit.JSONObjectTest.jsonObjectToStringIndent(JSONObjectTest.java:1648)
```

johnjaylward:
@RMelanson If you read my last comment on GitHub directly (I edited it a lot which doesn't go through emails) you'll see that this looks like expected behavior. When a JSONArray or JSONObject has only one element or key, then the indentation does not take effect.

@stleary I don't mind changing this, but I think updating the `toString` and `write` javadocs would be sufficient to indicate the behavior.

stleary:
@johnjaylward thanks for doing the analysis. Agree this is a case of works-as-designed, not an error. Javadoc update would be a reasonable way to address the issue.

stleary:
[...]

PR #357, opened 2017-07-15
Update javadoc according to issue #356.

johnjaylward:
javadoc only update.


stleary:
Javadoc only update, no functional changes, OK to fast track.

No. A backwards-compatible change was added to demo the documented behavior

**Review status**
ACCEPTED

PR #358, opened 2017-07-26
Provide info about duplicate keys when parsing

migueltt:
Wrap `this.putOnce(...)` to catch `JSONException` when creating `JSONObject` from a `JSONTokener`.
The new exception message will contain the duplicate key name, plus the location of the error.

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -231,8 +231,12 @@ public class JSONObject {
             if (c != ':') {
                 throw x.syntaxError("Expected a ':' after a key");
             }
-            this.putOnce(key, x.nextValue());
-
+            // Catch exception but re-throw using tokenizer syntaxError(...) - provides info about where the error is located
+            try {
+                this.putOnce(key, x.nextValue());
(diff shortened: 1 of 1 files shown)


JSONObject.java
```diff
@@ -231,8 +231,12 @@ public JSONObject(JSONTokener x) throws JSONException {
             if (c != ':') {
                 throw x.syntaxError("Expected a ':' after a key");
             }
-            this.putOnce(key, x.nextValue());
-
+            // Catch exception but re-throw using tokenizer syntaxError(...) - provides info about where the error is located
+            try {
+                this.putOnce(key, x.nextValue());
+            } catch (JSONException _e) {
```
johnjaylward: we don't really use _ (underscore) prefx for any other variables. can you rename it to just 'e'

migueltt:
Fixed (_e to e)

stleary:
Please provide an explanation of why this change is needed. 
What unit tests should be added? (See https://github.com/stleary/JSON-Java-unit-test)

PR #362, opened 2017-08-19
Fixes XML Unescaping

### LINKED ISSUE #361: Xml.java doesn't cope with encoded ampersand followed by semicolon
Hi,

I've come across an issue when using `XML.toJSONObject(String)` where the string contains an encoded ampersand followed by a semicolon (i.e. `<xml>Can cope &amp;; </xml>`)
The issue is in the `unescape` method in Xml.java. Since the `&amp;` has been converted to an `&` already in the XmlTokener, the string it's trying to unescape looks like `<xml>Can cope &; </xml>` and it assumes that there will be an XML entity to parse preceding the `&`.
Not sure if you just want to put a length check before checking the first character, i.e.
```
final String entity = string.substring(i + 1, semic);
if(!entity.isEmpty()){
                    if (entity.charAt(0) == '#') {
...
}
```

Full Stacktrace:
```
java.lang.StringIndexOutOfBoundsException: String index out of range: 0
	at java.lang.String.charAt(String.java:658)
	at org.json.XML.unescape(XML.java:193)
	at org.json.XML.stringToValue(XML.java:434)
	at org.json.XML.parse(XML.java:398)
	at org.json.XML.toJSONObject(XML.java:485)
	at org.json.XML.toJSONObject(XML.java:456)
```

Unit test for you too:
```
import org.json.XML;
import org.junit.Assert;
import org.junit.Test;

public class AuthorsNaturalOrderingPreservedInMetadataXslTest {
    @Test
    public void testXmlToJson() {

        String xml;
        xml = "<xml>Can cope &lt;; </xml>";
        Assert.assertTrue(XML.toJSONObject(xml).toString().equals("{\"xml\":\"Can cope <;\"}"));

        xml = "<xml>Can cope &amp; ; </xml>";
        Assert.assertTrue(XML.toJSONObject(xml).toString().equals("{\"xml\":\"Can cope & ;\"}"));

        xml = "<xml>Can cope &amp;; </xml>";
        xml.isEmpty()
        try {
            Assert.assertTrue(XML.toJSONObject(xml).toString().equals("{\"xml\":\"Can cope &;\"}"));
        } catch (StringIndexOutOfBoundsException e) {
            e.printStackTrace();
            Assert.fail("Failed");
        }
    }
}
```

johnjaylward:
**Key Changes:**

Fixes #361

* Removes unescape from the XML class calls
* fixes bug with unescape method
* moves unescape logic into the XMLTokener class for more consistency

Bug fix for XML Escaping that was included in a2d3b59394d62074debc4903e46c3aa97179172e as part of a fix supporting unicode entities (i.e. `&#xA` which maps to a new line character)

The original fix was causing XML string to be unescaped twice which was also causing the bug as mentioned in #361.

**Risks**
Low/none. This is a bug fix to correct incorrect parsing of XML entities.

**Changes to the API?**
One new package level method was added to `XMLTokener` that handles all valid XML entity unescaping.

Yes. The bug affects processing XML documents only, however, the bug prevents valid XML files from being read.

No

No. All unit tests should pass. New unit tests were added in order to test the changes and ensure that entities are only unencoded once. See https://github.com/stleary/JSON-Java-unit-test/pull/79

Yes. The actual implementation for the unescapeing was moved from the `XML` class to the `XMLTokener` class.

`XMLTokener` was already unescaping named entities (i.e. `&lt;` to `<` and `&amp;` to `&`) however it was not converting unicode entities back to their codepoints (i.e `&#34` to `"` or `&#x34` to `4`). To compensate for that I originally created the `XML.unescape` method and incorrectly injected it into the XML processing.

This PR corrects that by moving all the entity handling into the `XMLTokener` class and removing the `unescape` injection that I placed into the processing last time.

**Review status**
**APPROVED** Starting 3 day window for comments.

### DIFF
--- a/JSONML.java
+++ b/JSONML.java
@@ -174,7 +174,7 @@ public class JSONML {
                             if (!(token instanceof String)) {
                                 throw x.syntaxError("Missing value");
                             }
-                            newjo.accumulate(attribute, keepStrings ? XML.unescape((String)token) :XML.stringToValue((String)token));
+                            newjo.accumulate(attribute, keepStrings ? ((String)token) :XML.stringToValue((String)token));
                             token = null;
                         } else {
                             newjo.accumulate(attribute, "");
--- a/XML.java
+++ b/XML.java
@@ -141,7 +141,7 @@ public class XML {
                 if (mustEscape(cp)) {
                     sb.append("&#x");
                     sb.append(Integer.toHexString(cp));
-                    sb.append(";");
+                    sb.append(';');
                 } else {
                     sb.appendCodePoint(cp);
                 }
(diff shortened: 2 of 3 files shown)


stleary:
Sorry for the delay, back from eclipse, approved.

PR #364, opened 2017-09-28
The ability to specify your conversion to JSON

ZxUkr:
I ran into the problem that the library goes into an endless loop on objects with looped links.

### DIFF
--- a/JSONable.java
+++ b/JSONable.java
@@ -1,5 +1,7 @@
 package org.json;
 
 public interface JSONable {
-	public JSONObject toJSONObject();
+	default public JSONObject toJSONObject() {
+		return new JSONObject(JSONObject.createJSONObject(this));
+	}
 }
--- a/JSONable.java
+++ b/JSONable.java
@@ -2,5 +2,4 @@ package org.json;
 
 public interface JSONable {
 	public JSONObject toJSONObject();
-	public String toJSON();
 }
--- a/JSONable.java
+++ b/JSONable.java
@@ -2,6 +2,6 @@ package org.json;
 
 public interface JSONable {
 	default public JSONObject toJSONObject() {
-		return new JSONObject(JSONObject.createJSONObject(this));
+		return JSONObject.createJSONObject(this);
 	}
 }
--- a/JSONable.java
+++ b/JSONable.java
@@ -1,0 +1,6 @@
+package org.json;
+
+public interface JSONable {
+	public JSONObject toJSONObject();
+	public String toJSON();
+}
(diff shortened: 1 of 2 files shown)


JSONable.java
```diff
@@ -0,0 +1,6 @@
+package org.json;
+
+public interface JSONable {
+	public JSONObject toJSONObject();
+	public String toJSON();
```
johnjaylward: What is the point of the `toJSON()` interface method? It seems redundant to the [`JSONString#toJSONString()`](https://github.com/stleary/JSON-java/blob/master/JSONString.java#L17) method

ZxUkr: I think only in
object.toJSON() instead object.toJSONObject().toString()

johnjaylward: I would remove it. It's not obvious that that was the intent and I don't really see it as that helpful if there is no default implementation (like in java8 which we don't support yet)

stleary:
Please provide sample code that shows how this change solves a problem in the lib.

ZxUkr:
That is example code which cannot be processed by the library:
```
public class ClassOne {
	protected LinkedHashMap<String,Object> helpfulData;
	protected LinkedHashMap<String,Object> loopedData;

	public LinkedHashMap<String, Object> getHelpfulData() {
		return helpfulData;
	}

	public LinkedHashMap<String, Object> getLoopedData() {
		return loopedData;
	}

	public ClassOne() {
		helpfulData = new LinkedHashMap<>();
		helpfulData.put("one", "1");
		helpfulData.put("two", "2");

		loopedData = new LinkedHashMap<>();
		loopedData.put("one", this);
		loopedData.put("two", "2");
	}
}
...
ClassOne variable = new ClassOne();
JSONObject json = new JSONObject(variable);
```
After this commit I can processed this case manually.
```
public class ClassOne implements JSONable {
	protected LinkedHashMap<String,Object> helpfulData;
	protected LinkedHashMap<String,Object> loopedData;

	public LinkedHashMap<String, Object> getHelpfulData() {
		return helpfulData;
	}

	public LinkedHashMap<String, Object> getLoopedData() {
		return loopedData;
	}

	public ClassOne() {
		helpfulData = new LinkedHashMap<>();
		helpfulData.put("one", "1");
		helpfulData.put("two", "2");

		loopedData = new LinkedHashMap<>();
		loopedData.put("one", this);
		loopedData.put("two", "2");
	}

	@Override
	public JSONObject toJSONObject() {
		return new JSONObject(helpfulData);
	}
}
```

stleary:
Thanks for the pull request. The Javadocs for various library methods note that parsed data structures must be acyclical. This enhancement would allow users to bypass that restriction by controlling what gets parsed. This change cannot be accepted at the present time. It does not seem to be consistent with the library's goal of providing a simple reference implementation for how to parse JSON in Java.

PR #371, opened 2017-10-27
Update README to point users to alternate implementation 

slankas:
Many companies, organizations, and projects do not wish to use this project because of Doug Crockford's license. An alternative clean-room implementation was written. However, that version (https://github.com/tdunning/open-json) used the same package name as this work. That alternate version has created conflict when projects use this project as well as other projects that include the clean-room implementation. This message will help steer individual to a fork of that project that has an appropriate package name.


stleary:
Thanks for the pull request, but cannot accept this change at the present time.

PR #376, opened 2017-11-04
Compatability with jitpack.io

HoldYourWaffle:
I have no idea if this is something people want, but I needed it so I created a fork. It can't hurt to offer my (probably useless) changes to the greater public if I profited from it.

This PR adds gradle to the project, making it compatible with [jitpack.io](http://jitpack.io) for easy use in other gradle projects. I also reformatted the README to markdown, I really don't know/care if this is needed/wanted, if it isn't I can revert it right away.


MinnDevelopment:
See https://github.com/stleary/JSON-java/wiki/FAQ
>Why isn't support included for Maven, Gradle, or any other tools?

For this kinda support you should use your fork instead of this repository.

HoldYourWaffle:
For some reason I already expected some things like this, never thought I would've been able to find it in an FAQ.

So yeah you got a point there. I'll gladly use my own fork for this, I just created a PR because it's against my principles to "hide" your own (maybe useful for somebody) changes if you can just make a PR.

Issue #379, opened 2017-11-06
False positives in JSONObject.similar() and JSONArray.similar()?


enoes:
I'm afraid the for-loops in the above mentioned methods are breaking to soon if only one similarity was detected, see line 1487 in JSONObject.java and 1292 in JSONArray.java. Or did I overlook something?

johnjaylward:
Yes, you did. Those lines are determining if the 2 references in memory are the same. If they are, there is no way their child objects could differ.

* https://github.com/stleary/JSON-java/blob/master/JSONObject.java#L1847
* https://github.com/stleary/JSON-java/blob/master/JSONArray.java#L1292

johnjaylward:
wait, you may be correct. @stleary , let me make a test case just to confirm. it's possible that those `return`s should be `break`s instead

stleary:
Issue resolved.

PR #380, opened 2017-11-06
Fix for false positives in similar functions

johnjaylward:
fix for issue #379

**Risks**
None.

**Changes to the API?**
None.

Yes.

No.

Yes, but new unit tests were added that pass with this change. See https://github.com/stleary/JSON-Java-unit-test/pull/81

No.

**Review status**
**ACCEPTED**. Starting 3 day window for comments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+# ignore eclipse project files
+.project
+.classpath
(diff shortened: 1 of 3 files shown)


stleary:
Starting 3 day window for comments. 
Actually, the code change does not appear to break the unit tests, unless I am missing something. 
Deferring new release, since we are in a 3 month self-imposed moratorium for feedback/reaction to the Android support changes in #350. 
Thanks @enoes for catching the problem and @johnjaylward for fixing it.

wiiitek:
👍

PR #389, opened 2018-01-07
Added JSONObject.put for char

blakemcbride:
JSONObject.put("key", 'x');
   and
JSONObject.put("key", (Character) 'x');

used to give different results.  The first one would treat 'x' as in int.  This code makes them work the same (treat as a character not an int).

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -1589,6 +1589,22 @@ public class JSONObject {
     }
 
     /**
+     * Put a key/char pair in the JSONObject.
+     *
+     * @param key
+     *            A key string.
+     * @param value
+     *            A char which is the value.
+     * @return this.
+     * @throws JSONException
+     *             If the key is null.
+     */
+    public JSONObject put(String key, char value) throws JSONException {
+	    this.put(key, (Character)value);
+        return this;
+    }
+
+    /**
      * Put a key/value pair in the JSONObject. If the value is null, then the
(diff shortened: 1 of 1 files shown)


JSONObject.java
```diff
@@ -1588,6 +1588,22 @@ public JSONObject put(String key, Map<?, ?> value) throws JSONException {
...
+     * @throws JSONException
+     *             If the key is null.
+     */
+    public JSONObject put(String key, char value) throws JSONException {
+	    this.put(key, (Character)value);
+        return this;
+    }
+
     /**
      * Put a key/value pair in the JSONObject. If the value is null, then the
```
johnjaylward: I'd rather have this be `Character.valueOf(value)` instead of a cast.

blakemcbride:
I just noticed my editor messed up the method format a bit.  Easy to fix.  Just a few extra spaces.

JSONObject.java
blakemcbride: Either way...

johnjaylward: The valueOf method should offer caching on most jvm implementations while the cast never will. The valueOf should be preferable for any of the primitive to object boxing conversions.

blakemcbride: Okay.  Will do.

blakemcbride:
I just noticed my editor messed up the method format a bit.

PR #390, opened 2018-01-07
Added JSONObject.put for char

blakemcbride:
JSONObject.put("key", 'x');
and
JSONObject.put("key", (Character) 'x');

used to give different results. The first one would treat 'x' as in int. This code makes them work the same (treat as a character not an int).

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -1589,6 +1589,22 @@ public class JSONObject {
     }
 
     /**
+     * Put a key/char pair in the JSONObject.
+     *
+     * @param key
+     *            A key string.
+     * @param value
+     *            A char which is the value.
+     * @return this.
+     * @throws JSONException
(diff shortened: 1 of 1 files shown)


stleary:
Passes unit tests, but a new test will need to be added for coverage. Based on the text in #389 it sounds like this is a regression. Can anyone identify the pull request or release where it was introduced?

blakemcbride:
Should be using #390 instead of #389.  #390 includes a correction.

stleary:
Thanks for the pull request. It cannot be accepted at this time due to changing lib behavior. Please see the Wiki FAQ:  https://github.com/stleary/JSON-java/wiki/FAQ#how-do-you-decide-which-pull-requests-to-accept

Unless I missed something the current behavior goes back to the first commit of this repo.

PR #400, opened 2018-02-04
XML toJSONObject(Reader reader)

foldvari:
- Currently converting XML to JSON can be done using toJSONObject(String string) method only. In case of huge XML-s the need to convert them to string is very memory consuming. This limits usability. 
As JSONTokener is working with a Reader anyway, introducing a new method in the XML class which accepts a Reader is a very simple, risk free change.

**Risks**
- Low

**Changes to the API?**
- 3 additional methods.

- No

- Not necessary

- No. Existing unit tests validate this change as well.

- XML.java toJSONObject(String string, boolean keepStrings) now converts String to Reader

**Review status**
**ACCEPTED**
Starting 3 day comment window

### DIFF
--- a/XML.java
+++ b/XML.java
@@ -24,6 +24,8 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 */
 
+import java.io.Reader;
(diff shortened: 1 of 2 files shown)


stleary:
Thanks for the pull request. Accepted, starting comment window.

PR #404, opened 2018-03-01
Added stream function to JSONArray

aghasemi:
Added stream function to JSONArray in order to obtain similar functionality to Java collections

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -35,6 +35,7 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
 
 /**
  * A JSONArray is an ordered sequence of values. Its external text form is a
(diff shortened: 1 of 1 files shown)


erosb:
This would be a very nice improvement, but I'm afraid this can't be merged, because it breaks java6 compatibility.

aghasemi:
Oops. Ok. So there is no plan to drop Java6 support in the future? I'll be back then.

johnjaylward:
there is no plan at this time, but it is expected to happen eventually. Since JSONArray implements Iterable, you should be able to create a stream using the StreamSupport class:

```java
JSONArray myArray = new JSONArray();
// add some elements to the array somehow

java.util.stream.StreamSupport.stream(new JSONArray().spliterator(), false);
```

aghasemi:
How about adding a `toList()` method or something like that? Then we have streams as well.

johnjaylward:
Personally, I don't even like our Iterator support. It offers raw access to the underlying type without any of the checks/conversions to JSON Types. It should be used sparingly with the understanding that the underlying object may not come back as an expected type.

aghasemi:
But then integration with the "rest " of Java will be way easier.

johnjaylward:
I'm not sure what you mean by that. I also don't see why having a toList method would be any better than the first option I provided using StreamSupport.

MinnDevelopment:
Unless I'm mistaken there already is a toList: https://github.com/stleary/JSON-java/blob/d402a99fd87dc5c5d20f24342656237b18629838/JSONArray.java#L1477-L1500

aghasemi:
Yes that's true. My bad.

stleary:
Thanks for the pull request, but cannot be accepted at the present time. Please see https://github.com/stleary/JSON-java/wiki/FAQ#what-version-of-java-does-this-project-require

Issue #402, opened 2018-02-22
Minor - JSONObject.put(String Object) throws wrong Exception


krakenreleased:
The javadoc states, invoking JSONObject.put(String key, Object value) with a null key should throw a JSONException, however it is throwing a null pointer exception instead.

```**
     * Put a key/value pair in the JSONObject. If the value is null, then the
     * key will be removed from the JSONObject if it is present.
     *
     * @param key
     *            A key string.
     * @param value
     *            An object which is the value. It should be of one of these
     *            types: Boolean, Double, Integer, JSONArray, JSONObject, Long,
     *            String, or the JSONObject.NULL object.
     * @return this.
     * @throws JSONException
     *             If the value is non-finite number or if the key is null.
     */
    public JSONObject put(String key, Object value) throws JSONException {
        if (key == null) {
            throw new NullPointerException("Null key.");
        }
        if (value != null) {
            testValidity(value);
            this.map.put(key, value);
        } else {
            this.remove(key);
        }
        return this;
    }
```

johnjaylward:
I will have a patch for this soon.

johnjaylward:
Actually, @stleary, Douglas was the last one to touch both the JavaDoc and the code that throws the exception. He changed it way back in a73066f6bacd7c2fbb545dbbc2be3e569493b0b5 from throwing a JSONException to throwing the NPE.

I'd say we defer to the JavaDoc on this one, but perhaps a simple JavaDoc update is all that's required. Sadly the commit message only really mentions one of many changes that were applied (mostly formatting), so I'm not sure if the code change was intended or not,

stleary:
@johnjaylward Thanks for looking up the history. No preference. Leaving the NullPointerException alone is safer, changing to JSONException maintains consistency with the most of the rest of the lib, other than JSONPointer.

stleary:
I believe this has been addressed by #405. If you think otherwise, let me know.

PR #405, opened 2018-03-07
Updates javadoc to match actual exceptions thrown.

johnjaylward:
* Corrects Javadoc for #402 
* Some API methods are added/modified
* Updates some some boxing calls to use the newer `<Number>.valueOf` methods instead of `new <Number>`
* Updates calls to return sooner. i.e 
```java
this.put(value);
return this;
```
changes to:
 ```java
return this.put(value);
```

**Risks**
Low. Changes are either documentation only, or semantic to reduce stack by a single dword by reusing the return values. Changes to API methods are not expected to affect existing applications.

**Changes to the API?**
* Adds new API methods:
JSONArray put(float) throws JSONException
JSONArray put(int, float) throws JSONException
* Modifies behavior of existing API methods
JSONArray put(double) : Removes validity check (because put(Object) does)
JSONArray put(Object): Tests validity before assignment
JSONArray put(int, Object): tests validity if in-list, add instead of put NULL after end of current list.
JSONObject(Map<?,?>): Throws NPE if key is null

No

This PR updates the documentation. If some is missed, please note it in the comments or Code Review sections.

No. New unit tests were added to reinforce the API

Yes. As mentioned above some boxing calls were updated to the newer java5+ `<Number>.valueOf` instead of using the constructors as recommended by java code practices. Also, some returns were optimized which should not affect execution beyond decreasing stack size by a single dword.

**Review status**
**ACCEPTED**

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -36,6 +36,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
+
 /**
  * A JSONArray is an ordered sequence of values. Its external text form is a
  * string wrapped in square brackets with commas separating the values. The
@@ -182,7 +183,7 @@ public class JSONArray implements Iterable<Object> {
      * Construct a JSONArray from an array
      *
      * @throws JSONException
-     *             If not an array.
+     *             If not an array or if an array value is non-finite number.
      */
     public JSONArray(Object array) throws JSONException {
         this();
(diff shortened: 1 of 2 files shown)


stleary:
Looks good, starting 3 day comment window.

data-enabler:
> JSONObject(Map): Throws NPE if key is null

A bit late here, but I would consider throwing an exception for an input that used to be accepted a breaking API change. In my company's codebase we had some tests that verified this behavior that were broken as a result, and a currently unknown amount of code that could actually be depending on this behavior.

Obviously this ship has already sailed, but can we at least point out the breaking change in the release history documentation? Would be very useful as a heads-up for anyone trying to upgrade from an old version.

stleary:
Moving this comment to new issues

Issue #401, opened 2018-02-21
Customize the default key naming logic/process in generated json from getters


dhruv-niki:
Hi,
It seems the default logic to generate the json keys based on public getters is something like this:
1. Find public functions named: get[A-Z]+
2. In generated json, the key name is the part after 'get' with the first letter in lowercase - but case 3 below indicates that's not it.

The auto naming works great most of the time, but I have one particular case where it's an issue.
I need the json output to be: `{ "mCode": "xyz" }`

If my getter is named --> json generated is:
1. `getmCode()` --> not picked up at all, since lowercase letter after 'get'
2. `getmcode()` --> not picked up at all, since lowercase letter after 'get'
3. `getMCode()` --> `"MCode": "xyz"`
4. `getMcode()` --> `"mcode": "xyz"`
5. `getMcOde()` --> `"mcOde": "xyz"`

So I can see no way of automatically getting "mCode": "xyz" in the json. For now I am manually adding it to the JSONObject.

Can there be a way to custom define the key naming logic for a particular getter, or an entire class?

johnjaylward:
@stleary the issue seems to be here:
https://github.com/stleary/JSON-java/blob/master/JSONObject.java#L1449
```java
if (key.length() > 0
      && Character.isUpperCase(key.charAt(0))) {
```

I'm not sure why the uppercase check is in there. Looks like it's been there at least 5 years and was put in by Douglas.

It does seem like something we shouldn't be selecting on. It looks like if we remove the uppercase check there, it should not make the autonaming logic change, just be more inclusive.

If removed, @dhruv-niki 's examples would look more like this:
1. `getmCode()` --> `"mCode": "xyz"`
2. `getmcode()` --> `"mcode": "xyz"`
3. `getMCode()` --> `"MCode": "xyz"`
4. `getMcode()` --> `"mcode": "xyz"`
5. `getMcOde()` --> `"mcOde": "xyz"`

stleary:
I would guess that the original intent was to support Netbean getters. Not convinced it is it should be changed. Users who need this functionality should probably just fork the project. However, OK to go ahead if anyone:
* Feels strongly about supporting this
* Can write the code in a way that does not affect existing applications
* Can update the unit tests as needed

mouse07410:
> I'm not sure why the uppercase check is in there. Looks like it's been there at least 5 years and was put in by Douglas.

Not meaning to insult, but it looks like an example of shortsightedness in design.

> Not convinced it should be changed.

[...]

PR #406, opened 2018-03-07
Adds annotations to customize field names during Bean serialization

johnjaylward:
Allows users to customize key names when performing Java Bean conversion into a `JSONObject`. This is in reference to #401. This PR also has support to force the serializer to ignore bean getters if a specific annotation is provided.

**Risks**
LOW. Should not affect existing programs. If an existing program does require a field to be ignored, then the new annotation can be used to force it out of the resulting JSONObject result.

**Changes to the API?**
Two (2) new annotation classes were created that allow customization of field names when used on Bean getters.

Example:
```java
public class MyBean {
    @JSONPropertyName("myStringField")
    public String getSomeString() { return "someStringValue"; }
    @JSONPropertyIgnore
    public String getSomeString2() { return "someStringValue2"; }
}

new JSONObject(new MyBean());
```

the output of the above JSONObject should look like this:
```json
{
    "myStringField" : "someStringValue"
}
```

yes

No, This PR updates both the Javadoc and the README where needed.

No. New tests have been added to support the new annotations.

Yes. Some private functions were created to supplement the `populateMap` method and support the new Annotations.

**Review status**
**ACCEPTED**

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -29,6 +29,7 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
+import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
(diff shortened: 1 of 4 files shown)


stleary:
Nice work on the the project's first attempt with annotations! 
- Since this is only used for Bean getters, we might be able to come up with a better name than JSONPropertyName, which seems a little ambiguous, and overly similar to Jackson's JSONProperty. JSONBeanName? I don't know, that actually sounds worse...
- There is no use case for JSONPropertyIgnore. Needs to be removed.
- JSONObjectTest testGenericIntBean() failed because the new code considers 'able' to be a key. This is a regression caused I think by removing the uppercase check for key.charAt(0). Needs to be fixed.
- Method doc for getAnnotationDepth() does not match actual return type. Although I think this method  goes away if JSONPropertyIgnore is removed.

johnjaylward:
* For the annotation name, I just used the one that was mentioned in #401, but updated the case to match the rest of our class name. I was unaware of the property name from Jackson. I'm open to a better name though, I didn't think that hard about it and am not attached.
* For JSONPropertyIgnore, I currently use a similar feature to make sure that my user beans don't emit the password field. Even though they are hashed, I still don't want it transmitted to the clients.
* The regression was noted, and I updated the test case in https://github.com/stleary/JSON-Java-unit-test/pull/84. I also feel that we probably shouldn't case about the casing there. In my private fork of the project, I encode everthing that starts with "get", "has", or "is" regardless of casing.
* Fixed the "return" javadoc for the depth method.

stleary:
No preference on the name either, unless someone comes up with something better, and OK to proceed with JSONPropertyIgnore, since there are people who will use it.
No regressions please. Existing behavior has to be retained.

johnjaylward:
added in the skip over keys that start with a lowercase, but only in the event that the annotation is not present.

stleary:
Accepted, starting 3 day comment window.

PR #411, opened 2018-03-19
Fix for invalid processing of trailing / for JSON Pointer

### LINKED ISSUE #410: JSON pointer query, trailing slashes and empty keys
Hey,

I found an inconsistency, or potentially a bug, with the handling of JSON pointers in JSONObject.query().

Specifically, the handling of trailing slashes differs between the root element and elements further down.

From reading the RFC and a discussion about it, I got the impression that a trailing slash is supposed to refer to an empty key, but this only works with the root element.
For elements further down, a trailing slash will not select the empty key below the element, but the element itself instead.

example code
```
public void testPointerQuery() {
	JSONObject none = new JSONObject();
	JSONObject emptyKeyRoot = new JSONObject("{'':'empty key at root'}");
	JSONObject emptyKeyBelow = new JSONObject("{'foo':{'':'empty key below foo'}}");
	
	System.out.println("json=" + none + " ptr='' query -> " + none.query(""));
	System.out.println("json=" + none + " ptr='/' query -> " + none.query("/"));
	
	System.out.println("json=" + emptyKeyRoot + " ptr='' query -> " + emptyKeyRoot.query(""));
	System.out.println("json=" + emptyKeyRoot + " ptr='/' query -> " + emptyKeyRoot.query("/"));
	
	System.out.println("json=" + emptyKeyBelow + " ptr='' query -> " + emptyKeyBelow.query(""));
	System.out.println("json=" + emptyKeyBelow + " ptr='/' query -> " + emptyKeyBelow.query("/"));
	System.out.println("json=" + emptyKeyBelow + " ptr='/foo' query -> " + emptyKeyBelow.query("/foo"));
	System.out.println("json=" + emptyKeyBelow + " ptr='/foo/' query -> " + emptyKeyBelow.query("/foo/"));
}
```

output for example code above
```
json={} ptr='' query -> {}
json={} ptr='/' query -> null
json={"":"empty key at root"} ptr='' query -> {"":"empty key at root"}
json={"":"empty key at root"} ptr='/' query -> empty key at root
json={"foo":{"":"empty key below foo"}} ptr='' query -> {"foo":{"":"empty key below foo"}}
json={"foo":{"":"empty key below foo"}} ptr='/' query -> null
json={"foo":{"":"empty key below foo"}} ptr='/foo' query -> {"":"empty key below foo"}
json={"foo":{"":"empty key below foo"}} ptr='/foo/' query -> {"":"empty key below foo"}
```

expected output (last line)
```
json={"foo":{"":"empty key below foo"}} ptr='/foo/' query -> empty key below foo
```

This happens with both the string and URI fragment representations, using the 20180130 maven artifact.

Thanks,
M.

johnjaylward:
Fixes #410.  `String.split` does not take into account trailing nulls when splitting. It instead trims them off and they are "lost".

**Risks**
Moderate. This is a bug fix to account for faulty processing of the token string in a JSONPointer expression. There is some change to existing behavior.

**Changes to the API?**
No

Yes

No

Yes. One test case accidentally had a trailing `/` that causes an error with the corrected code. That test case has been corrected to use the correct JSONPointer expression. There were no unit tests to cover this type of input as expected input. New test cases have been created in https://github.com/stleary/JSON-Java-unit-test/pull/85

Yes. I replaced the String.split call with custom processing of the token string.

**Review status**
**ACCEPTED** 
Changes to existing behavior are not usually accepted, but in this case the new code changes an edge case of a new feature.  For  more information please see the updated Wiki page: https://github.com/stleary/JSON-java/wiki/FAQ#what-kind-of-changes-will-be-accepted-for-this-project

### DIFF
--- a/JSONPointer.java
+++ b/JSONPointer.java
@@ -158,9 +158,28 @@ public class JSONPointer {
             throw new IllegalArgumentException("a JSON pointer should start with '/' or '#/'");
         }
         this.refTokens = new ArrayList<String>();
-        for (String token : refs.split("/")) {
-            this.refTokens.add(unescape(token));
-        }
+        int slashIdx = -1;
+        int prevSlashIdx = 0;
+        do {
+            prevSlashIdx = slashIdx + 1;
+            slashIdx = refs.indexOf('/', prevSlashIdx);
+            if(prevSlashIdx == slashIdx || prevSlashIdx == refs.length()) {
+                // found 2 slashes in a row ( obj//next )
+                // or single slash at the end of a string ( obj/test/ )
+                this.refTokens.add("");
+            } else if (slashIdx >= 0) {
+                final String token = refs.substring(prevSlashIdx, slashIdx);
+                this.refTokens.add(unescape(token));
+            } else {
+                // last item after separator, or no separator at all.
+                final String token = refs.substring(prevSlashIdx);
+                this.refTokens.add(unescape(token));
+            }
+        } while (slashIdx >= 0);
+        // using split does not take into account consecutive separators or "ending nulls"
+        //for (String token : refs.split("/")) {
+        //    this.refTokens.add(unescape(token));
+        //}
     }
 
     public JSONPointer(List<String> refTokens) {


johnjaylward:
@erosb Please review since you are most intimate with JSONPointer.

erosb:
Thank you for fixing this @johnjaylward . I'm a bit sorry to see there was no simpler way to do it.

Do you also plan to add a unittest?

johnjaylward:
New unit tests are available. I think I covered the cases.

johnjaylward:
There may be an easier way, but I figured this should be fast enough given how modern JVM and string library implementations handle substrings.

stleary:
@johnjaylward  Breaks a unit test, can you investigate?
org.json.junit.JunitTestSuite > org.json.junit.JSONObjectTest.jsonObjectAppend FAILED
    org.json.JSONPointerException at JSONObjectTest.java:670
Using the committed unit test code, not pull request 410.

johnjaylward:
The test was using incorrect syntax for the unit test. To get the item in a JSONArray at index 1 it was using `/myArray/1/` but the trailing slash there is invalid for the data being expected. The Trailing / there indicates that we are looking for a JSONObject to be in the first index of the array and have a blank key `""`. However, the test actually wanted the value `false`.

I updated the test case and it should pass now.

mfriedemann:
Hey, as reporter of #410 I can confirm that this proposed change fixes #410 and the issue in the actual code I was putting together.

Thanks,
M.

erosb:
Hello, thank you for putting effort into fixing this.
I'm not at home this week and my current environment is not set up for looking into & testing the change locally, so feel free to proceed with opening the 3-day merge window.

stleary:
Accepted, starting 3 day comment window.

Issue #394, opened 2018-01-26
Issue with xml.java package


raviteja713:
Trying to convert JSON data into XML string using the below logic of code snippet. But, output xml has a JSONObject being displayed after "content" attribute as show below. Any suggestions.?

# **_# input.json:_**

```
{
  "id": "0001",
  "type": "donut",
  "name": "Cake",
  "payload": {
	"uname":"David"
    "content": {
      "batter": [
        {
          "id": "1001",
          "type": "Regular",
          "mode": [
            "rel",
            "don"
          ]
        }
      ]
    }
  }
}

```
# **_jsontoxml.java:_**

```
org.json.JSONObject jsonObject = new org.json.JSONObject(jsonData);
string xml="<?xml version=\"1.0\" encoding=\"ISO-8859-15\"?>\n<"+root+">" + org.json.XML.toString(jsonObject) + "</"+root+">
```

# _output.xml:_

```
<?xml version="1.0" encoding="ISO-8859-15"?>
<root>
    <id>0001</id>
    <type>donut</type>
    <name>Cake</name>  
	<payload>
        <uname>David</uname>
        {"batter": [{"id": "1001","type": "Regular","mode": ["rel","don"]}]}
    </payload>
</root>
```

stleary:
"content" is an unfortunately-named keyword in the XML <-> Java transformation. For the history of this issue, please see #344, #286, and #108. For more information about how content is used, see XMLTest.java contentOperations() in https://github.com/stleary/JSON-Java-unit-test. 

No objection if someone wants to propose a workaround along the lines of #108, or any other approach that does not break existing applications.

johnjaylward:
#108 looks like a decent implementation to me. I don't have time to work on it now, but may in a few weeks.

johnjaylward:
@stleary what do you think of having a spplemental configuration object for the XML parsing. something like this:

```java
public class XMLParserConfiguration {
    /**
     * When parsing the XML into JSON, specifies if values should be kept as strings (true), or if
     * they should try to be guessed into JSON values (numeric, boolean, string)
     */
    public final boolean keepStrings;
    /**
     * The name of the key in a JSON Ojbect that indicates a CDATA section. Historically this has
     * been the value "content" but can be changed. Use <code>null</code> to indicate no CDATA
     * processing.
     */
    public final String cDataTagName;

    /**
     * Default parser configuration. Does not keep strings, and the CDATA Tag Name is "content".
     */
[...]

PR #412, opened 2018-03-22
Initial implementation of XMLParserConfig object for flexible XML Parsing

johnjaylward:
DRAFT
Implements improvement requested in #394. Adds a new configuration class that can be used ot change how the XML Parser works.

**Risks**
Low. Current functionality works as expected.

**Changes to the API?**
3 new public APIs are available that take advantage of possible 
* `public static JSONObject toJSONObject(Reader reader, XMLParserConfiguration config) throws JSONException`
* `public static JSONObject toJSONObject(String string, XMLParserConfiguration config) throws JSONException`
* `public static String toString(final Object object, final String tagName, final String cDataTagName)`

A private method signature was also changed, but should not affect the API.

Yes

Yes

No. No new unit tests were created yet, but we probably should add some.

Not beyond extending the public APIs to use the new configuration object.

**Review status**
NOT REVIEWED


stleary:
@johnjaylward Looks great! Please add a complete set of unit tests to exercise the new functionality.

johnjaylward:
I'm not sure how soon I can get tests together. If someone else knows they have time, then please comment here and generate the pull request. Otherwise it may be some time before I get to them.

stleary:
If more XML config options are added in the future (e.g. #429), the XMLParserConfig constructor param list won't scale, and the current implementation using public final params can't be modified after creating the object. Recommend using fields with getters/setters for the parser config class. I think it is acceptable in this case to have a single empty constructor that creates an object no params set, and require users to set the config params individually before use.

stleary:
Approved, starting 3 day comment window.  
**Note:** The XMLParserConfiguration API is subject to change if additional config params are proposed.

srinivasprv:
Please update the maven jar. This change is not present in latest maven jar.

stleary:
@srinivasprv Thanks for the reminder, will get this started in the next few days.

PR #414, opened 2018-04-04
Removed use of deprecated function new Double()

moseslecce:


stleary:
Looks good to me also, but someone needs to make sure the unit tests either have coverage or add a test for it. I can look into it this weekend.

stleary:
See #417

PR #417, opened 2018-05-03
fix double ctor in JSONWriter

stleary:
This replaces #414 which had a coding standards error (tab instead of spaces). Also just noticed we don't have any coding standards published. Will update the Wiki.

### DIFF
--- a/JSONWriter.java
+++ b/JSONWriter.java
@@ -391,7 +391,7 @@ public class JSONWriter {
      * @throws JSONException If the number is not finite.
      */
     public JSONWriter value(double d) throws JSONException {
-        return this.value(new Double(d));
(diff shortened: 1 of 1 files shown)


stleary:
The Double ctor is deprecated, so changed to with the recommended replacement method. This appears to only be an OpenJDK issue, but no harm in maintaining compatibility. For more information, see:
[https://bugs.openjdk.java.net/browse/JDK-8176335](https://bugs.openjdk.java.net/browse/JDK-8176335).
This code was originally submitted  in #414 but that changed used a tab char, which has been corrected.

**Risks**
Low

**Changes to the API?**
No

No

No

No. Confirmed that the existing tests already provide coverage for the modified line.

No

**Review status**

PR #421, opened 2018-05-19
add isEmpty and isNotEmpty methods

strkkk:
It is clearer to check if object is empty or not by calling method rather then comparing length.
Also, it can be used as method reference, see code snippet below.

        JSONObject example = new JSONObject();
        example.length() == 0; // was 
        example.isEmpty(); // become 

        example.length() != 0; // was 
        example.isNotEmpty(); // become 

        List<JSONObject> list = ...;
        list.stream().filter(JSONObject::isEmpty).filter(JSONObject::isNotEmpty).count(); 

**Risks**
Low

**Changes to the API?**
Yes

No

Methods are added with javadocs

No. This methods are new.

~~No~~ Internal uses of length() == 0 and similar code were converted to isEmpty()

**Review status**
**APPROVED**
(isNotEmpty() is not included, JSONArray.isEmpty() was added)
Starting 3 day comment window

### DIFF
--- a/CDL.java
+++ b/CDL.java
@@ -224,7 +224,7 @@ public class CDL {
      */
     public static JSONArray toJSONArray(JSONArray names, JSONTokener x)
             throws JSONException {
-        if (names == null || names.length() == 0) {
+        if (names == null || names.isEmpty()) {
             return null;
         }
         JSONArray ja = new JSONArray();
@@ -235,7 +235,7 @@ public class CDL {
             }
             ja.put(jo);
         }
-        if (ja.length() == 0) {
+        if (ja.isEmpty()) {
             return null;
         }
         return ja;
@@ -272,7 +272,7 @@ public class CDL {
      */
     public static String toString(JSONArray names, JSONArray ja)
             throws JSONException {
-        if (names == null || names.length() == 0) {
+        if (names == null || names.isEmpty()) {
             return null;
         }
         StringBuffer sb = new StringBuffer();
--- a/CDL.java
+++ b/CDL.java
@@ -224,7 +224,7 @@ public class CDL {
      */
     public static JSONArray toJSONArray(JSONArray names, JSONTokener x)
             throws JSONException {
-        if (names == null || names.isEmpty()) {
+        if (names == null || names.length() == 0) {
             return null;
         }
         JSONArray ja = new JSONArray();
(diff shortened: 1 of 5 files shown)


johnjaylward:
No other collection type in Java has a `isNotEmpty` method that I know of. I'd rather just have the `isEmpty` check then in your example, it would look like:

```java
JSONObject example = new JSONObject();
    example.length() == 0; // was 
    example.isEmpty(); // become 

    example.length() != 0; // was 
    example.isNotEmpty(); // become 

    List<JSONObject> list = ...;
    list.stream().filter(JSONObject::isEmpty).filter(item -> !item.isEmpty()).count();
```

strkkk:
@johnjaylward 
Yep, java collections does not have it, but JSONObject != java collection, isn't it? :)
Some libs provide such method, e.g. [CollectionUtils](https://commons.apache.org/proper/commons-collections/javadocs/api-release/org/apache/commons/collections4/CollectionUtils.html#isNotEmpty(java.util.Collection)) to make it easier to read.

Arguably, but Scala collections/strings/option have nonEmpty method as part of standard API as well.
Comparing both approaches with lambda, I would prefer method reference way.
`list.stream().filter(JSONObject::isNotEmpty).count();`

May be we need more opinions.

stleary:
No objection to adding isEmpty(), but please do it the right way:
* Add the method to JSONArray as well, unless you have some reason why it should not be included.
* Update the code in the lib where JSONObject and JSONArray empty condition is checked. I think there are about 4 occurrences.
* Update those unit tests that check for JSONObject and JSONArray empty condition. There should be about 16 occurrences. 
* If you do the previous 2 steps it won't be necessary to add explicit unit tests.

I don't think there is a valid use case for isNotEmpty().

strkkk:
- Added isEmpty() to JSONArray
- Removed isNotEmpty() method
- Refactored some usages of length() method for JSONArray, JSONObject and String

Will update tests later.

stleary:
Starting 3 day comment window

johnjaylward:
@stleary If we want to ensure Android compatability, then the other classes should not be updated, only JSONArray and JSONObject.

i.e. if using the CDL class in android, you'll get a no method exception for isEmpty.

stleary:
My bad, thanks @johnjaylward. @strkkk please revert the changes to CDL, JSONML, and XML. I will need to update the Wiki regarding Android compatibility.

strkkk:
Ok, I will change it. 
by the way, why changes in this classes will break Android compatibility?

strkkk:
@stleary 
Tests are updated, changes for mentioned classes are partially reverted.

Issue #429, opened 2018-08-03
Handling of null-values in JSONObject and XML.toString()


tobihein:
We have a json-string and want to convert it to a corresponding xml-string. In case, that there are null-values in the json-string, we get an unexpected result

**JSon-String**
```
{
   "myKey": null
}
```

**What we do**
```
String jsonString="{\"myKey\": null}";
final JSONObject jsonObject = new JSONObject( jsonString );
finalString xmlString = XML.toString( jsonObject));
```

**What we get**
`<myKey>null</myKey>`

**What we expect**
`<myKey/>`

Is our expectation correct? If yes, I would do a pull-request and change line 594 in class `XML.java`

johnjaylward:
No, in this case it's working as designed. In Javascript and JSON, null is a value (as opposed to undefined). In order to maintain the JSON value of null when converting to XML we output that as  the string "null".

tobihein:
Thanks for your quick reply.

So in our case there is no difference in the xml for the input

```{
   "myKey": null,
   "myKey2": "null"
}```

Do you have a suggestion how we can get our expected output?

johnjaylward:
hmm, now that I'm looking at the code and the history of it, this may be a bug. At https://github.com/stleary/JSON-java/blob/master/XML.java#L594 you can see a null check for the value. This is the java `null` which this library uses to represent javascript `undefined`. However, there is no further check for the JSON `null`  (https://github.com/stleary/JSON-java/blob/master/JSONObject.java#L165) which has a `toString` method that returns`"null"`. (This is why null is output for your sample JSON. The library internally converts it to that `JSONObject.NULL` object).

In reality, the null check in the `XML` class should never happen since the `JSONObject.put` methods all remove keys with `null` values. Hence, the `keySet` call should never contain values that are JAVA `null`. I feel like the intent of that check was to actually check for JSON null and not JAVA null.

@stleary Let me know your opinion on this. It may be a small tweak we want to make.

stleary:
@johnjaylward thanks for investigating. Are you suggesting adding a check for JSON null or replacing the existing check?

johnjaylward:
Replacing the existing check with  `JSONObject.NULL.equals(value)`.

stleary:
Sounds reasonable, no objection on this end.

tobihein:
Thanks, you two. My pull-request would have looked exactly like yours, johnjaylward.

I'm looking forward to the upcomming release.

[...]

PR #430, opened 2018-08-03
Changes NULL handling in XML output

johnjaylward:
Adjusts null handling in XML output in reference to #429.

**Risks**
Due to the change in `null`  handling, release notes should be updated to reflect the change in behavior. Previously, `{ "key": null }` would output to `<key>null</key>`. It will now output to `<key/>`. Also,  empty string values `""` would output to `<key/>`, but will now output to `<key></key>`.

**Changes to the API?**
None.

Yes

Yes

Yes. New unit test are provided in a separate PR (https://github.com/stleary/JSON-Java-unit-test/pull/87)

no

**Review status**
NOT REVIEWED

### DIFF
--- a/XML.java
+++ b/XML.java
@@ -591,7 +591,7 @@ public class XML {
             jo = (JSONObject) object;
             for (final String key : jo.keySet()) {
                 Object value = jo.opt(key);
-                if (value == null) {
+                if (JSONObject.NULL.equals(value)) {
                     value = "";
                 } else if (value.getClass().isArray()) {
                     value = new JSONArray(value);
--- a/XML.java
+++ b/XML.java
@@ -120,6 +120,9 @@ public class XML {
      * @return The escaped string.
      */
     public static String escape(String string) {
+        if (string == null) {
+            return "";
+        }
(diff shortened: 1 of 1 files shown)


johnjaylward:
I'll take a look. At the moment work is pretty busy, but I'll try to get to this in the next week or so.

stleary:
Thanks!

johnjaylward:
I'm not sure I'd label it a bug per-say... before this PR, the 2 different tag types would generate empty strings:
```xml
<root>
<Tag1 />
<Tag2></Tag2>
<root>
```

The JSONML would generate JSON that indicated Tag1 and Tag2 are empty strings.

After the change, JSONML would indicate that Tag1 is null, and Tag2 is an empty string.

I still need to check the JSONML spec to verify if that is a proper interpretation.

stleary:
JSONML behavior should not be changed, or a case made as to why it is currently incorrect.

kmnalluri:
What happened to these changes? I don't see any of them in the latest releases.

johnjaylward:
they were abandoned as I did not have time to validate the JSONML portions

PR #188, opened 2016-01-27
Fix negative zero

johnjaylward:
Fixes #187 -0 now returns as a double.

PR #439, opened 2018-10-02
-0 get cast to a string.

mmilkin:
When serializing JSONObject {"key": -0.0} to string and then back again will result in a "-0" string.

Based on ECMA-404 -0 is a valid number.

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -2186,6 +2186,13 @@ public class JSONObject {
                     }
                 } else {
                     Long myLong = Long.valueOf(string);
+
+                    // If the value is equal to 0 then the prefix of negative should be removed.
+                    if (myLong.equals(0) && string.equals("-0"))
+                    {
+                        return Integer.valueOf(myLong.intValue());
+                    }
+
                     if (string.equals(myLong.toString())) {
                         if (myLong.longValue() == myLong.intValue()) {
                             return Integer.valueOf(myLong.intValue());


johnjaylward:
we specifically put in support for -0.
https://github.com/stleary/JSON-java/pull/188

Do you have a test case that shows the exact issue you are seeing?

johnjaylward:
they are linked in the readme.md

mmilkin:
Here is the short test for this in my project.
```
JSONObject jsonData = new JSONObject()
jsonData.put("key", -0.0);

String serialized = jsonData.toString();
Assert.assertEquals(-0.0, new JSONObject(serialized).get("key"));
```

johnjaylward:
If I remember correctly, You can't use assertEquals for the double comparison. the get method returns an Object, in this case a Double. For comparing Doubles, you should really use Double.compare instead.

for example in our test cases we have something like this:
```java
    @Test
    public void jsonObjectValues() {
        String str = 
            "{"+
                "\"negZeroKey\":-0.0,"+
                "\"negZeroStrKey\":\"-0.0\""+
            "}";
        JSONObject jsonObject = new JSONObject(str);
        assertTrue("opt negZeroKey should be a Double", 
                jsonObject.opt("negZeroKey") instanceof Double);
        assertTrue("get negZeroKey should be a Double", 
                jsonObject.get("negZeroKey") instanceof Double);
        assertTrue("opt negZeroKey should be double", 
                Double.compare(jsonObject.optDouble("negZeroKey"), -0.0d) == 0);
        assertTrue("opt negZeroStrKey with Default should be double", 
                Double.compare(jsonObject.optDouble("negZeroStrKey"), -0.0d) == 0);
        assertTrue("optNumber negZeroKey should return Double",
                jsonObject.optNumber("negZeroKey") instanceof Double);
        assertTrue("optNumber negZeroStrKey should return Double",
                jsonObject.optNumber("negZeroStrKey") instanceof Double);
        assertTrue("optNumber negZeroKey should be -0.0", 
                Double.compare(jsonObject.optNumber("negZeroKey").doubleValue(), -0.0d) == 0);
        assertTrue("optNumber negZeroStrKey should be -0.0", 
                Double.compare(jsonObject.optNumber("negZeroStrKey").doubleValue(), -0.0d) == 0);
```

I trimmed out stuff that's not relevant to this case.

All those tests pass with the current code.

Issue #441, opened 2018-10-04
Change the underflying storage for floating point numbers to BigDecimal


johnjaylward:
The current implementation of JSONObject/JSONArray will store most floating point numbers as double when parsing a JSON text. For example:
```json
{ "myFloat" : 1.234 }
```
Will parse into a JSONObject with a key of "myFloat" having a value of `Double.valueOf("1.234")`. When directly outputting this new JSONObject, no inconsistencies are noted as the `Double.toString` automatically performs some magic rounding to get us the same string representation of "1.234".

However, the same can't be said of conversion to BigDecimal. Using the same JSON text/JSONObject above, a subsequent call to `getBigDecimal("myFloat")` will instead return a BigDecimal with a value of `1.2339999675750732421875`. The main reason for using this type of conversion is to preserve any possible value that may have happened due to a purposeful calculation as opposed to assuming the value should be 1.234. For example: `JSONObject.put("myKey",someDouble+1.234f);` Where the value is purposefully set to a double value.

The goal of this issue is not to change the conversion from `double`->`BigDecimal`, but instead to change how we interpret the textual numbers upon parsing.

Numbers in the JSON specifications are always noted in decimal form; i.e. Hex, octal, and binary numbers are not allowed in the formal specification. Java however, uses binary notation for it's floating point numbers `float` and `double` (see [ieee 754](https://en.wikipedia.org/wiki/IEEE_754)). This causes an inconsistency between what is presented in the original data, and what we parse and store in the backend of JSONObject and JSONArray.

By changing the parser to always use BigDecimal, the backing data will better match the file data that is parsed. The get/opt methods themselves would return a different implementation type, but the get/optDouble methods would still return the proper values. Given how the raw get/opt methods work, I don't think this is a large issue as it would currently return any of 6 Number types (BigDecimal, BigInteger, Double, Float, Long, or Integer) if the value is a number (it changes based on number format and size).

stleary:
This would be an improvement.

stleary:
Please continue the discussion in #453.

PR #453, opened 2018-12-10
changes number parsing to use BigDecimal as the backing type

johnjaylward:
See issue #441 (use BigDecimal for underlying storage of decimal number types)

**Risks**
Medium

**Changes to the API?**
No, but it changes how certain API calls work for decimal/floating-point types work.

for example a JSON document list this:
```json
{ "float":1234 }
```
would parse like this:
```java
JSONObject jo = new JSONObject("{\"float\":1234}");
jo.get("float"); // used to return a Double, now returns a BigDecimal
jo.getDouble("float"); // still returns the value/object as in the old version (Double)
```
In general, the raw "get/opt" methods with no type information (like getDouble, getInt, optDouble, etc) are not guaranteed to be useful as the backing types were never guaranteed. However, people may be using them without the knowledge that the backing types are semi-dynamic (i.e. number types have always been backed differently if an `Integer` was parsed vs a `float`). So even a "standard" document like this:
```json
{
"items": [
 { "floatValue": 123.4 },
 { "floatValue": 321 }
]
}
```
Would have a `Double` stored for the first `floatValue` and an `Integer` for the second `floatValue` before the change. After the change they would be stored as `BigDecimal` and `Integer` respectively.

Yes.

Yes. we should probably mention that the backing type has changed and how that may affect existing code.

Yes, many of the tests depend on the data type stored for numbers. The tests have been updated to reflect the new backing type.

Yes, changes were made to the number parsing methods for XML and JSONObject (they need to be kept in sync for Android support)

**Review status**
**APPROVED**

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -2109,48 +2109,54 @@ public class JSONObject {
         if ((initial >= '0' && initial <= '9') || initial == '-') {
             // decimal representation
             if (isDecimalNotation(val)) {
-                // quick dirty way to see if we need a BigDecimal instead of a Double
-                // this only handles some cases of overflow or underflow
-                if (val.length()>14) {
-                    return new BigDecimal(val);
+                // Use a BigDecimal all the time so we keep the original
+                // representation. BigDecimal doesn't support -0.0, ensure we
+                // keep that by forcing a decimal.
+                try {
+                    BigDecimal bd = new BigDecimal(val);
+                    if(initial == '-' && BigDecimal.ZERO.compareTo(bd)==0) {
+                        return Double.valueOf(-0.0);
+                    }
+                    return bd;
+                } catch (NumberFormatException retryAsDouble) {
+                    // this is to support "Hex Floats" like this: 0x1.0P-1074
+                    try {
+                        Double d = Double.valueOf(val);
+                        if(d.isNaN() || d.isInfinite()) {
+                            throw new NumberFormatException("val ["+val+"] is not a valid number.");
+                        }
+                        return d;
+                    } catch (NumberFormatException ignore) {
+                        throw new NumberFormatException("val ["+val+"] is not a valid number.");
+                    }
                 }
-                final Double d = Double.valueOf(val);
-                if (d.isInfinite() || d.isNaN()) {
-                    // if we can't parse it as a double, go up to BigDecimal
-                    // this is probably due to underflow like 4.32e-678
-                    // or overflow like 4.65e5324. The size of the string is small
-                    // but can't be held in a Double.
-                    return new BigDecimal(val);
+            }
+            // block items like 00 01 etc. Java number parsers treat these as Octal.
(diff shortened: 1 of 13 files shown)


stleary:
Breaks a lot of unit tests. Aren't users likely to see the same breakages?

johnjaylward:
Yes, I'm not sure how impactful that would be though. The APIs that are affected are the Object get/opt and not the typed ones (i.e. getString and getDouble still work the same way). The only reason the raw ones are affected is because they type changed, not the value. A programmer calling the Object get/opt methods would hopefully not be depending on a certain type like we do in the test cases.

stleary:
Is there any way to restrict this so that just values that are added as BigDecimal get the new behavior?

PeerHeijnen:
The behavior for this conversion has changed from 20180813 to 20190722.

`
		System.out.println( new JSONObject( new JSONObject().put( "test", new BigDecimal( "348.8" ) ).toString() ).getBigDecimal( "test" ) );
`

Output for 20180813: `384.8`

Output for 20190722: `348.80000000000001136868377216160297393798828125`

Instead of chaging this behavior by default, would it be possible to choose a method for parsing that is suitable to the application?

johnjaylward:
@PeerHeijnen, the main issue here is that you do the following steps:
1. Create a JSONObject with a single `BigDecimal` field
1. Convert the JSONObject to a JSON Text
1. Convert the JSON Text back to a JSONObject. The parser will read the number as a `double`
1. Read a `double` value as a `BigDecimal`.

The conversion from `double` to `BigDecimal` uses the exact double value which is the output you see now. If you set the `BigDecimal` scale to 1 or 2, you would lose the extra info you don't want from the double.

This PR is meant to fix the round tripping that causes the issue you are seeing. Currently in `master` the backing store is always an `int`, `long`, or `double` which makes round trips more difficult as the `double` is inaccurate as a storage means for decimal floating point.

Another option would be to add another getter with a `scale` parameter so you can configure the scale you wish to read right away as one call.
```java
BigDecimal bd = JSONObject.getBigDecimal("test", 4);
System.out.println(bd); // would print 348.8000 abbreviated to 348.8
```

Using today's API, this would look like:
```java
BigDecimal bd = JSONObject.getBigDecimal("test").setScale(4);
System.out.println(bd); // would print 348.8000 abbreviated to 348.8
```

PeerHeijnen:
Sure, it was not ment as criticism. I just noted the change in behavior, be it intentional or not.

The code I shared was not actual production code, just a short version of how the conversion happens as you explained.

I was just wondering whether people may experience side effect from the PR. For me, it solves my problems I have now perfectly, so thanks for the effort! :)

johnjaylward:
rebased to support the new project structure

stleary:
Starting 3 day comment window

johnjaylward:
@TheMatthew Do you still have the benchmark tests you were running for the other improvements we made? I was wondering if you could spin them up for this change... I should have probably thought of doing that in the last 1.5 years this has been open, but it just occurred to me now

PR #467, opened 2019-04-09
add configuration for xsi:nil="true" conversion to null

meiskalt7:
Hello :-)

I add configuration for xsi:nil="true" conversion to null. In JSON spec null should be represented like
{ key: null }
link: https://stackoverflow.com/questions/21120999/representing-null-in-json

Please check it and merge, cause it is necessary for many dependent projects.

Representation of null in accordance with JSON spec

**Risks**
Low

**Changes to the API?**
No, only addition new constructor for XMLParserConfiguration

Yes.

Yes. Javadoc already updated.

No

Yes, changes were made to the number parsing methods for XML and added new configuration
(no changes in public methods)

**Backward compatibility**
Yes

**Review status**
APPROVED

### DIFF
--- a/XML.java
+++ b/XML.java
@@ -31,7 +31,7 @@ import java.util.Iterator;
 /**
  * This provides static methods to convert an XML text into a JSONObject, and to
  * covert a JSONObject into an XML text.
- * 
+ *
  * @author JSON.org
  * @version 2016-08-10
  */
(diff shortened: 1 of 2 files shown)


stleary:
HI @meiskalt7. Thanks for submitting this PR, it is appreciated, and sorry for not responding sooner. Looks like no one on the team has a strong opinion regarding this change. Your change does not break unit tests, so that is good, but you will need to add additional unit tests if you want to proceed. Please refer to the [FAQ](https://github.com/stleary/JSON-java/wiki/FAQ).  Also, please update your pull request, which should make a compelling case for why this change would be an improvement on the existing implementation. I did not find the stackoverflow link convincing, but if others have a different opinion, please let me know.

meiskalt7:
Thanks for your answer. I will add tests for this changes ASAP.

meiskalt7:
I added test for this pull request

stleary:
Approved. Starting 3 day comment window.

PR #470, opened 2019-07-01
JSONObject constructor with custom map supplier

isg-software:
Added a constructor for JSONObject allowing the user to choose the Map implementation (instead of the default HashMap).
By choice of map implementations like LinkedHashMap or some sorted map the user can influence the iteration order and thus especially the order of the toString-methods' output. While this is not important for machine-readable JSON, for a pretty-printed, human-readable JSON created by #toString(int) (probably meant for humans to edit the data), a custom order may increase readability and maintainability of the generated code.

# Background
I needed to create JSON files which should be easily readable and editable by humans. The JSON should meet two criteria:

* It should be pretty-printed, and
* the properties of certain JSON objects should be printed in a certain order.

While the JSONStringer outputs all properties in the order of the .key().value()-calls, it does not support pretty printing. The alternative is to create JSONObjects with the default constructor, add properties by its put() methods and then call the toString(int) method, which performs pretty-printing – but the default HashMap container of JSONObject does not support outputting the properties in insertion order (order of the put() calls).

That's why I added a possibility to create a new, empty JSONObject with a new constructor allowing to select a different Map implementation, like a SortedMap oder a LinkedMap (the latter preserving insertion order, which is precisely what I needed).

Instead of passing a newly created map as parameter I decided to use a supplier argument in order to reduce the risk that the caller passes a non-empty map or retains a reference to the created map, enabling him to manipulate it other than by the JSONObject's methods. Also, a new constructor with a Map argument would have clashed with the existing one taking a map and putting all it's content into a new internal HashMap.

Normally I'd have used `java.util.function.Supplier<Map<String, Object>>` as argument type, but since that would have meant to drop support for Java 1.6 and 1.7, I instead introduced a new (functional) interface.

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -98,7 +98,7 @@ import java.util.regex.Pattern;
  * </ul>
  *
  * @author JSON.org
- * @version 2016-08-15
+ * @version 2019-07-01
  */
 public class JSONObject {
     /**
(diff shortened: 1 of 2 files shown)


stleary:
@isg-software Thanks for the pull request; your work to improve this project is appreciated. Please see the Wiki FAQ section on JSONObject ordering:
https://github.com/stleary/JSON-java/wiki/FAQ#why-isnt-ordering-allowed-in-jsonobjects

Ordering of objects is a frequently raised topic when JSON is discussed. In keeping with the JSON definition, reference apps like this one should not support ordering of JSONObjects. If  your application needs this functionality, please consider forking the project and providing it there, or adding this capability outside of the app.

PR #474, opened 2019-07-23
Prefer unsynchronized StringBuilder

gaul:
`StringBuffer` provides unneeded synchronization.  Found via
error-prone.


stleary:
Replacing an internally used inefficient builtin class 

**Risks**
Low

**Changes to the API?**
No

No

No

No, and additional unit tests are not needed.

No

**Review status**
**Approved**, starting 3 day comment window

PR #475, opened 2019-07-26
Make private methods static where possible

gaul:
This avoids an unneeded object reference.  Found via error-prone.


stleary:
error-prone warnings

**Risks**
**Changes to the API?**
None/No

**Review status**
**Approved** Starting 3 day comment window

Issue #448, opened 2018-11-15
Creating JSONObject accept character after closing bracket.


jagdeep85:
Why is following statement is legal ? Shouldn't it be a parsing error (JSONException) ?

JSONObject obj = new JSONObject("{}whateverishere") ;
System.out.println(obj);

This is a valid statement and print on '{}' ?

johnjaylward:
yes, that is expected. Our parser does very little validation. They way the parser is working in this case is is as follows:
1. Is the first character (ignoring white space) an open brace '{'. If not, error, if yes, parse as an object
2. read all the key/value pairs
3. stop at the end brace '}'
4. ignore everything after the end brace.

The underlying JSONTokener that is parsing the text would theoretically let you parse the next valid JSON text in the original string, in this case it would be 'whateverishere'. The JSONObject constructor you use though doesn't give you access to the underlying tokener.

jagdeep85:
Thanks! 👍

PR #481, opened 2019-08-14
clarifies exception that the parser makes when reading JSON

johnjaylward:
Just README updates. I had this sitting in my local repo for ages and forgot to push it. The clarifications stem from #448


johnjaylward:
also #447

stleary:
**ACCEPTED** No comment window needed.

PR #482, opened 2019-09-06
Fix problem with precision when retrieving BigDecimal JSON properties.

sollecitom:
As it is now, retrieving a property with value "0.01" with `JSONObject.getBigDecimal(key)` will return "0.01000000000000000020816681711721685132943093776702880859375" instead "0.01".

### DIFF
--- a/JSONObject.java
+++ b/JSONObject.java
@@ -1139,7 +1139,7 @@ public class JSONObject {
             if(Double.isNaN(d)) {
                 return defaultValue;
             }
-            return new BigDecimal(((Number) val).doubleValue());
+            return BigDecimal.valueOf(((Number) val).doubleValue());
         }
         if (val instanceof Long || val instanceof Integer
                 || val instanceof Short || val instanceof Byte){


stleary:
Breaks unit tests. Please see https://github.com/stleary/JSON-Java-unit-test

stleary:
@sollecitom Thanks for the pull request, it is appreciated. This project is in maintenance mode, so most changes are user-driven.  We ask that code submitters follow the process from end to end, which includes ensuring the unit tests pass, as described in this wiki page: https://github.com/stleary/JSON-java/wiki/How-to-contribute-code-for-this-project

Breakages may indicate a change in behavior to existing code. When that happens, we have to determine if fixing the problem is more urgent than potentially breaking existing applications. You can find more information about this in the [FAQ](https://github.com/stleary/JSON-java/wiki/FAQ).

johnjaylward:
Also, see PR #453 and issue #441 which also address issues similar to this one

stleary:
Closed for inactivity.

Issue #484, opened 2019-09-17
Infinite Loop in org.json.JSONML.toJSONObject


wdblair:
The following input causes the `org.json.JSONML.toJSONObject` method to run in an infinite loop.

    JSONML.toJSONObject("??*^M??|?CglR^F??`??>?w??PIlr^E??D^X^]?$?-^R?o??O?*??{OD?^FY??`2a????NM?b^Tq?:O?>S$^K?J?^FB.gUK?m^H??zE??^??!v]?^A???^[^A??^U?c??????h???s???g^Z???`?q^Dbi??:^QZl?)?}1^??k?0??:$V?$?Ovs(}J??^V????2;^QgQ?^_^A?^D?^U?Tg?K?`?h%c?hmGA?<!C*^P^Y?^X9?~?t?)??,z^XA???S}?Q??.q?j????]");

If you trace the execution of JSONObject on this input, you see that it eventually makes it to the `JSONML.parse` method.

    856715 [INVOKE] org.json.JSONML.parse(Lorg/json/XMLTokener;ZLorg/json/JSONArray;Z)Ljava/lang/Object;(571585615 false 0 false );

But then it gets stuck inside the parse method. If I'm reading the trace correctly, it repeatedly calls JSONTokener.next() on the XMLTokener `x` object on line 93 in JSONML.java https://github.com/stleary/JSON-java/blob/2a6af29235ffe7eeff157bc280c11b664ad54a5e/JSONML.java#L93 It seems like there may be a missing case to throw a syntax error. Is the code searching for an end tag that doesn't exist?

You can find the full trace of the method running on this input in the following gist:

    https://gist.github.com/wdblair/e462ca73c791162aab14e19605b4ae14

The following Java program reproduces the issue when the org.json library is in the classpath.

    import java.util.Base64;
    import org.json.JSONML;

    public class JSONInfiniteLoop {

        public static void main(String argv[]) throws Throwable {
            String x = "Pz8qDT8/fD9DZ2xSBj8/YD8/Pj93Pz9QSWxyBT8/RBgdPyQ/LRI/bz8/Tz8qPz97T0Q/Blk/P2Ay" +
                       "YT8/Pz9OTT9iFHE/Ok8/PlMkCz9KPwZCLmdVSz9tCD8/ekU/P38/IXZdPwE/Pz8bAT8/FT9jPz8/" +
[...]

PR #485, opened 2019-09-17
Fix for #484

johnjaylward:
Fix for #484 that corrects EOF error when the Meta tag isn't closed properly and we reach the end of the input.

**Risks**
None. Bug fix

**Changes to the API?**
No

Yes

No

No. But a new unit test was created to prevent regressions. See: https://github.com/stleary/JSON-Java-unit-test/pull/94

No

**Review status**
**APPROVED**


stleary:
Approved, starting 3 day comment window.

PR #486, opened 2019-09-17
Standardize Exception Messages

johnjaylward:
Simply standardizes error messages when data types don't match expected value types. I.e. someone tries to convert the string "bob" to an int value.

**Risks**
None.

**Changes to the API?**
No.

Not immediately.

No.

Yes. There were corrections needed to match the new string values of the messages. See: https://github.com/stleary/JSON-Java-unit-test/pull/95

Yes. The exception code was pulled out to a common function similar to the `syntaxError` function used in the Tokeners.

**Review status**
**APPROVED**

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -249,7 +249,7 @@ public class JSONArray implements Iterable<Object> {
                         .equalsIgnoreCase("true"))) {
             return true;
         }
-        throw new JSONException("JSONArray[" + index + "] is not a boolean.");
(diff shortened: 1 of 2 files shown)


stleary:
Starting 7 day comment window.
Normally a 3 day window is sufficient, but this change could potentially break user applications. However, probably only if they are using test automation that expects exact wording in exceptions. Appropriate text should be included in the release text. 
Users who think this could cause them problems are welcome to comment here during the window.

PR #489, opened 2019-10-19
javadoc: update the javadoc files

sivaraam:
Generate the javadoc from the latest sources and also link
the to JDK javadocs.

The command used to generate the javadoc is the following:

```
javadoc -d ./ -link https://docs.oracle.com/javase/7/docs/api/ ./*.java
```


sivaraam:
The change can be found live in the following site: https://sivaraam.github.io/JSON-java/

stleary:
In review.

johnjaylward:
@stleary this may be a good candidate for something that gets automated through the new "github actions" support

stleary:
Thanks, I will look into it. The actions will need to be updated anyway when 1.6 test support is committed.

stleary:
Closed due to out of date changes. JavaDocs should still be updated from the latest code, and at least for every release. TBD...

PR #495, opened 2019-11-12
fix typo

harkue:
fix some typo and camel cases

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -1409,7 +1409,7 @@ public class JSONArray implements Iterable<Object> {
     public Writer write(Writer writer, int indentFactor, int indent)
             throws JSONException {
         try {
-            boolean commanate = false;
+            boolean hasComma = false;
             int length = this.length();
             writer.write('[');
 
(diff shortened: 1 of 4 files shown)


JSONArray.java
```diff
@@ -1409,7 +1409,7 @@ public Writer write(Writer writer) throws JSONException {
     public Writer write(Writer writer, int indentFactor, int indent)
             throws JSONException {
         try {
-            boolean commanate = false;
```
johnjaylward: `needsComma` I think would be a better name than `hasComma`

JSONObject.java
```diff
@@ -2496,7 +2496,7 @@ static final void indent(Writer writer, int indent) throws IOException {
     public Writer write(Writer writer, int indentFactor, int indent)
             throws JSONException {
         try {
-            boolean commanate = false;
```
johnjaylward: same here, `needsComma` seems like a better fit.

harkue: ok

stleary:
Refactoring typos from the code - no change to behavior.

**Risks**
Minimal

**Changes to the API?**
None


No

No

No

var names updated.

**Review status**
**APPROVED**

Starting 3 day comment window

PR #502, opened 2019-12-25
update JSONTokener's brief: parse simple json text

Alanscut:
JSONTokener can parse a simple json text to a simple json object.


stleary:
Readme update

**Risks**
None

**Changes to the API?**
No

No

N/A

N/A

No

**Review status**
**ACCEPTED** No comment window needed


Notes:

The "problem" may just be an enhancement, it does not have to be a bug fix.

In most cases a new release is not required.

After the code has been tested and reviewed, you can set the review status to something like, "ACCEPTED, Starting 3 day comment window", and then do the merge after 3 days. Minimum comment period is:
- None for emergency commits
- 3 days by default
- 7 days for complex, risky, or behavior-changing commits

PR #507, opened 2020-01-09
Creating Stream from JSONArray's spliterator

VitorMac10:
Added utility function to create stream from spliterator, there's also the possibility to create a stream from "myArrayList", but because of encapsulation, I would stick with JSONArray's spliterator instance instead

### DIFF
--- a/JSONArray.java
+++ b/JSONArray.java
@@ -35,6 +35,8 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


stleary:
@VitorMac10 thanks for the PR. At the present time, we are only accepting code changes that compile under Java 6. Please see [What version of java does this project require](https://github.com/stleary/JSON-java/wiki/FAQ#what-version-of-java-does-this-project-require). At some point this will probably change, at which time your PR can be re-submitted and considered.

Issue #517, opened 2020-05-21
Cookie needs an update to support HttpOnly (and also Secure)


Hubgut:
Hi,
I was happy to see your little Cookie-Implementation but it does not handle tokens without an '=' correctly!
There is only one exception for the "secure" token in the code, even "Secure" or other case-insensitive versions won't work. The standard request field "HttpOnly" will throw an exception like everything else too.
The best solution would be to handle all tokens without an '=' as a Boolean.TRUE.

johnjaylward:
Is there an RFC or other specification that notes this behavior?

johnjaylward:
found RFC Here:
https://tools.ietf.org/html/rfc6265

johnjaylward:
Looks like "User Agents" should ignore unknown attributes:

> Unless the cookie's attributes indicate otherwise, the cookie is
>    returned only to the origin server (and not, for example, to any
>    subdomains), and it expires at the end of the current session (as
>    defined by the user agent).  User agents ignore unrecognized cookie
>    attributes (but not the entire cookie).

So it looks like instead of throwing an error, we just set the "true" flag as suggested by the OP and move on, then the caller of our Cookie class is responsible for validating actual attributes.

johnjaylward:
Also:

> A user agent MUST use an algorithm equivalent to the following
>    algorithm to parse a "set-cookie-string":
> 
>    1.  If the set-cookie-string contains a %x3B (";") character:
> 
>           The name-value-pair string consists of the characters up to,
>           but not including, the first %x3B (";"), and the unparsed-
>           attributes consist of the remainder of the set-cookie-string
>           (including the %x3B (";") in question).
> 
>        Otherwise:
> 
>           The name-value-pair string consists of all the characters
>           contained in the set-cookie-string, and the unparsed-
>           attributes is the empty string.
> 
>    2.  If the name-value-pair string lacks a %x3D ("=") character,
>        ignore the set-cookie-string entirely.
> 
>    3.  The (possibly empty) name string consists of the characters up
>        to, but not including, the first %x3D ("=") character, and the
>        (possibly empty) value string consists of the characters after
>        the first %x3D ("=") character.
> 
>    4.  Remove any leading or trailing WSP characters from the name
>        string and the value string.
> 
>    5.  If the name string is empty, ignore the set-cookie-string
>        entirely.
> 
>    6.  The cookie-name is the name string, and the cookie-value is the
>        value string.
> 
[...]

PR #521, opened 2020-05-22
Updates Cookie class to be more generic in attribute parsing and emit.

johnjaylward:
See #517 
Brings the Cookie implementation closer to the RFC specification (https://tools.ietf.org/html/rfc6265)
This is so the library can age better as new attributes are added or removed in future RFC revisions.

This also fixes bugs in the implementation:
* Allowed empty cookie names which should be ignored
* Attribute names were case sensitive when they should not be
* Name and Value of the cookie were not ensured trimmed
* Attributes were not ensured to have white space removed

**Risks**
Low to Medium. I doubt many people were using this since the problems brought up in #517 were not found until now. They seem to be pretty glaring gaps in the parsing and emit logic.

**Changes to the API?**
No

Yes

I don't think it's needed

Yes, corrections were made

No, bug fixes and support for new attributes only.

**Review status**
**APPROVED**

Starting 3 day comment window

### DIFF
--- a/src/main/java/org/json/Cookie.java
+++ b/src/main/java/org/json/Cookie.java
@@ -27,6 +27,7 @@ SOFTWARE.
 /**
  * Convert a web browser cookie specification to a JSONObject and back.
  * JSON and Cookies are both notations for name/value pairs.
+ * See also: <a href="https://tools.ietf.org/html/rfc6265">https://tools.ietf.org/html/rfc6265</a>
  * @author JSON.org
  * @version 2015-12-09
  */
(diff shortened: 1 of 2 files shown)


stleary:
More discussion can be found in #517, which is now closed. Please continue the discussion here.

johnjaylward:
@hubgut, please review the latest changes. I believe I corrected the escaping issue and also a few other minor tweeks

stleary:
Unit tests fail for Maven with Java 7 but pass for Gradle with Java 8. The failures are in unrelated modules, due to the unit tests not accounting for possible differences in JSONObject ordering.
JSONObjectTest, XMLConfigurationTest, XMLTest. 
These failures will be fixed in a separate commit.

Hubgut:
@johnjaylward. Now, I found some time for my little project and changed to the most current version.
I had no time for thorough testing yet; but it seems to be working!
I have also looked into the Cookies class. It looks much better now. My former complaints are gone.

Issue #523, opened 2020-06-03
Unit tests appear to be broken - recent merge?


stleary:
**Top priority  - get the unit tests working again. Needs to be working by EOD 6/3/2020**
Unit tests now working for gradle 6.3 with Java 1.8, and maven 3.2.3 with Java 1.8.

**Before any more PRs can be accepted:**
* Code and unit tests have to compile and execute using Java 1.6 - 1.8 **1.8 completed**
* Gradle builds need to work, with documentation suitable for new developers on how to get it running. **Completed sufficently for now, docs pending**
* Maven builds need to work, with documentation suitable for new developers on how to get it running. **Completed sufficently for now, docs pending**
* Eclipse builds need to work, with documentation. Optionally, if anyone has gotten this to work with IntelliJ, that would be welcome as well.
* Documentation for how a person will little prior experience can manually find, compile, and use the lib files.
* Restore unit tests in JSONObjectTest: 
* * testSingletonBean()
* * testSingletonEnumBean()

johnjaylward:
@stleary the changes made here https://github.com/viveksacademia4git/JSON-java/blob/master/.github/workflows/maven.yml

provides an interface like this: https://github.com/viveksacademia4git/JSON-java/actions/runs/123544252

For every PR, it's automated and we can confirm if test cases are failing before merging them.

I think taking the changes from @viveksacademia4git would be a good idea for PR requests going forward.

viveksacademia4git:
Hello @stleary and @johnjaylward,

I was just going through this issue and I discovered a strange phenomenon, and I guess this issue is focusing on that. 
If you execute the *mvn test* command it exhibits that there are total number of 315 tests, as displayed in the pipeline/action stage: [test:line9053](https://github.com/viveksacademia4git/JSON-java/runs/734229352?check_suite_focus=true#step:4:9053).

But upon executing the tests from Eclipse, be it JUnit-Test or Coverage, the total number of unit-tests executed is exactly double i.e. 630.
 
![image](https://user-images.githubusercontent.com/45398326/83648481-64ae5900-a5b6-11ea-9ea0-10210368d3ca.png)

[...]

PR #524, opened 2020-06-03
Development for stleary#516 completed with rebased repository

viveksacademia4git:
- Introduced JSONObject(int) constructor.
   * param - int > Initial capacity of the underlying data structure

- Test for new introduced JSONArray(int) constructor.
   1. Checked with input parameter: 0
   2. Checked with input parameter: positive number
   3. Checked with positive number input parameter and compared length
   4. If the input parameter is negative number JSONException is thrown:
         JSONArray's initial capacity cannot be negative.

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -205,6 +205,22 @@ public class JSONArray implements Iterable<Object> {
         }
     }
 
+    /**
+     * Construct a JSONArray with the specified initial capacity.
+     *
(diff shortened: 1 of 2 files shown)


stleary:
PRs are temporarily on hold, see #523. I will review when the issues are resolved.

stleary:
This is an enhancement to allow users to allocate very large JSONArrays in the constructor

**Risks**
Low

**Changes to the API?**
No

No

No

No

No

**Review status**
**APPROVED**

PR #532, opened 2020-06-25
Verifies BigDecimal support does not have a regression

### LINKED ISSUE #531: JSONObject.objectToBigDecimal precision errors when parsing doubles
When parsing doubles the method uses the BigDecimal(double) constructor, which can have unpredictable results as stated in the docs: https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html#BigDecimal(double)

It is instead preferred to use BigDecimal.valueOf(double)
https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html#valueOf(double)

This lead to a few precision errors on my end, when updating from an older version of this library.

johnjaylward:
Fixes #531
Add test result to confirm that #531 is working in latest version. No code changes needed.

A new release should be scheduled so the fix can be used by people using the MAVEN artifacts.

### DIFF
--- a/src/test/java/org/json/junit/JSONObjectTest.java
+++ b/src/test/java/org/json/junit/JSONObjectTest.java
@@ -3081,6 +3081,10 @@ public class JSONObjectTest {
         assertNotNull(jo.get("ALL"));
     }
     
+    /**
+     * Sample test case from https://github.com/stleary/JSON-java/issues/531
+     * which verifies that no regression in double/BigDecimal support is present.
+     */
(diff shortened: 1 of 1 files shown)


stleary:
There was a question about whether BigDecimal support had introduced an unexpected regression. Added unit tests to confirm this was not the case.

**Risks**
None

**Changes to the API?**
No

No

No

No

No

**Review status**
**APPROVED**
No comment window is necessary in this case.

PR #540, opened 2020-07-19
Added type conversion support

kumar529:
The lib doesn't support type conversion for a specific value.
I have added the feature to support type conversion for a specific tag.
Example:

//XML String
`<root><id1 xsi:type="java.lang.String">1234</id1><id2 xsi:type="java.lang.Integer">1234</id2></root>`

//Corresponding JSON String
`{"root":{"id2":1234,"id1":"1234"}}`

### DIFF
--- a/src/main/java/org/json/XMLParserConfiguration.java
+++ b/src/main/java/org/json/XMLParserConfiguration.java
@@ -23,6 +23,9 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 */
 
+import java.util.Map;
+
+
 /**
  * Configuration object for the XML parser. The configuration is immutable.
  * @author AylwardJ
@@ -57,10 +60,9 @@ public class XMLParserConfiguration {
     private boolean convertNilAttributeToNull;
 
     /**
-     * When parsing the XML into JSON, specifies if values with attribute xsi:type="java.lang.Integer"
-     * should be kept as attribute(false), or they should be converted to the given type
+     * This will allow type conversion for values in XML if xsi:type attribute is defined
      */
-    public boolean useValueTypeCast;
+    public Map<String, XMLXsiTypeConverter<?>> xsiTypeMap;
 
     /**
      * Default parser configuration. Does not keep strings (tries to implicitly convert
@@ -129,7 +131,7 @@ public class XMLParserConfiguration {
      */
     @Deprecated
     public XMLParserConfiguration (final boolean keepStrings, final String cDataTagName, final boolean convertNilAttributeToNull) {
-        this(keepStrings, cDataTagName, convertNilAttributeToNull, false);
+        this(keepStrings, cDataTagName, convertNilAttributeToNull, null);
     }
 
     /**
@@ -140,16 +142,16 @@ public class XMLParserConfiguration {
      *      to use that value as the JSONObject key name to process as CDATA.
      * @param convertNilAttributeToNull <code>true</code> to parse values with attribute xsi:nil="true" as null.
      *                                  <code>false</code> to parse values with attribute xsi:nil="true" as {"xsi:nil":true}.
-     * @param useValueTypeCast  <code>true</code> to parse values with attribute xsi:type="java.lang.Integer" as
-     *                          integer,  xsi:type="java.lang.String" as string
-     *                          <code>false</code> to parse values with attribute xsi:type="java.lang.Integer" as {"xsi:type":"java.lang.Integer"}.
+     * @param xsiTypeMap  <code>new HashMap<String, XMLXsiTypeConverter<?>>()</code> to parse values with attribute
+     *                   xsi:type="integer" as integer,  xsi:type="string" as string
+     *                    <code>null</code> to use default behaviour.
      */
     public XMLParserConfiguration (final boolean keepStrings, final String cDataTagName,
-            final boolean convertNilAttributeToNull, final boolean useValueTypeCast ) {
+            final boolean convertNilAttributeToNull, final Map<String, XMLXsiTypeConverter<?>> xsiTypeMap ) {
         this.keepStrings = keepStrings;
         this.cDataTagName = cDataTagName;
         this.convertNilAttributeToNull = convertNilAttributeToNull;
-        this.useValueTypeCast = useValueTypeCast;
+        this.xsiTypeMap = xsiTypeMap;
     }
     
     /**
--- a/src/main/java/org/json/XMLParserConfiguration.java
+++ b/src/main/java/org/json/XMLParserConfiguration.java
@@ -114,7 +114,9 @@ public class XMLParserConfiguration {
      */
     @Deprecated
     public XMLParserConfiguration (final boolean keepStrings, final String cDataTagName) {
-        this(keepStrings, cDataTagName, false);
+        this.keepStrings = keepStrings;
+        this.cDataTagName = cDataTagName;
+        this.convertNilAttributeToNull = false;
     }
 
     /**
@@ -131,7 +133,9 @@ public class XMLParserConfiguration {
      */
     @Deprecated
     public XMLParserConfiguration (final boolean keepStrings, final String cDataTagName, final boolean convertNilAttributeToNull) {
-        this(keepStrings, cDataTagName, convertNilAttributeToNull, null);
+        this.keepStrings = keepStrings;
+        this.cDataTagName = cDataTagName;
+        this.convertNilAttributeToNull = convertNilAttributeToNull;
     }
 
     /**
--- a/src/main/java/org/json/XMLParserConfiguration.java
+++ b/src/main/java/org/json/XMLParserConfiguration.java
@@ -24,6 +24,7 @@ SOFTWARE.
 */
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 
 
@@ -278,7 +279,8 @@ public class XMLParserConfiguration {
      */
     public XMLParserConfiguration withXsiTypeMap(final Map<String, XMLXsiTypeConverter<?>> xsiTypeMap) {
         XMLParserConfiguration newConfig = this.clone();
-        newConfig.xsiTypeMap = Collections.unmodifiableMap(xsiTypeMap);
+        Map<String, XMLXsiTypeConverter<?>> cloneXsiTypeMap = new HashMap<String, XMLXsiTypeConverter<?>>(xsiTypeMap);
+        newConfig.xsiTypeMap = Collections.unmodifiableMap(cloneXsiTypeMap);
         return newConfig;
     }
 }
--- a/src/main/java/org/json/XMLParserConfiguration.java
+++ b/src/main/java/org/json/XMLParserConfiguration.java
@@ -57,6 +57,12 @@ public class XMLParserConfiguration {
     private boolean convertNilAttributeToNull;
 
     /**
+     * When parsing the XML into JSON, specifies if values with attribute xsi:type="java.lang.Integer"
+     * should be kept as attribute(false), or they should be converted to the given type
+     */
+    public boolean useValueTypeCast;
+
+    /**
      * Default parser configuration. Does not keep strings (tries to implicitly convert
      * values), and the CDATA Tag Name is "content".
      */
@@ -106,9 +112,7 @@ public class XMLParserConfiguration {
      */
     @Deprecated
     public XMLParserConfiguration (final boolean keepStrings, final String cDataTagName) {
-        this.keepStrings = keepStrings;
-        this.cDataTagName = cDataTagName;
-        this.convertNilAttributeToNull = false;
+        this(keepStrings, cDataTagName, false);
     }
 
     /**
@@ -125,9 +129,27 @@ public class XMLParserConfiguration {
      */
     @Deprecated
     public XMLParserConfiguration (final boolean keepStrings, final String cDataTagName, final boolean convertNilAttributeToNull) {
+        this(keepStrings, cDataTagName, convertNilAttributeToNull, false);
+    }
+
+    /**
+     * Configure the parser to use custom settings.
+     * @param keepStrings <code>true</code> to parse all values as string.
+     *      <code>false</code> to try and convert XML string values into a JSON value.
+     * @param cDataTagName <code>null</code> to disable CDATA processing. Any other value
+     *      to use that value as the JSONObject key name to process as CDATA.
+     * @param convertNilAttributeToNull <code>true</code> to parse values with attribute xsi:nil="true" as null.
+     *                                  <code>false</code> to parse values with attribute xsi:nil="true" as {"xsi:nil":true}.
+     * @param useValueTypeCast  <code>true</code> to parse values with attribute xsi:type="java.lang.Integer" as
+     *                          integer,  xsi:type="java.lang.String" as string
+     *                          <code>false</code> to parse values with attribute xsi:type="java.lang.Integer" as {"xsi:type":"java.lang.Integer"}.
+     */
+    public XMLParserConfiguration (final boolean keepStrings, final String cDataTagName,
+            final boolean convertNilAttributeToNull, final boolean useValueTypeCast ) {
         this.keepStrings = keepStrings;
         this.cDataTagName = cDataTagName;
         this.convertNilAttributeToNull = convertNilAttributeToNull;
+        this.useValueTypeCast = useValueTypeCast;
     }
     
     /**
--- a/src/main/java/org/json/XMLParserConfiguration.java
+++ b/src/main/java/org/json/XMLParserConfiguration.java
@@ -23,6 +23,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 */
 
+import java.util.Collections;
 import java.util.Map;
 
 
@@ -62,7 +63,7 @@ public class XMLParserConfiguration {
     /**
      * This will allow type conversion for values in XML if xsi:type attribute is defined
      */
-    public Map<String, XMLXsiTypeConverter<?>> xsiTypeMap;
+    private Map<String, XMLXsiTypeConverter<?>> xsiTypeMap;
 
     /**
      * Default parser configuration. Does not keep strings (tries to implicitly convert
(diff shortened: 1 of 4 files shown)


stleary:
Sounds like you are trying to make the case that Java-specific XML xsi:type strings should be supported. If so, is this commonly used by projects other than yours?  Otherwise, no real objection since we allow a bit more latitude for supporting XML to JSON transformations, which will always be imperfect. 

We already have too many fixed parameters for XMLParserConfiguration. This needs to be fixed first. The existing constructors need to be deprecated, the properties made non-final, and setters added to set up the config manually. I will add an issue to address this.

johnjaylward:
@kumar529 would it make sense to have mappings for `xsi:type` as a `HashMap` instead of a boolean?

i.e.
```java
/**
Interface to support conversions of XML xsl:type values to Java objects
*/
public interface XMLXsiTypeConverter<T> {
    public T convert(String value);
}

public void someFunc() {
Map<string, XsiTypeConverter<?>> XsiTypeMap = new HashMap<>();
XsiTypeMap.put("string", new XMLXsiTypeConverter<String>(){ public String convert(String value){return value;});
XsiTypeMap.put("int", new XMLXsiTypeConverter<Long>(){ public Long convert(String value){return Long.valueOf(value);});
XsiTypeMap.put("bool", new XMLXsiTypeConverter<Long>(){ public Long convert(String value){return Boolean.FALSE;});

XMLParserConfiguration config = new XMLParserConfiguration(false, "content", true, XsiTypeMap);

JSONObject jo = XML.toJSONObject(myInput, config);

}
```

stleary:
Let's try the simplest solution first, and see if it works to everyone's satisfaction.

kumar529:
@johnjaylward I think HashMap will not serve the purpose.
Because I want to add data type support on an element.

@stleary I am facing the problem of data conversion. Currently, the lib supports two types of config for data conversion:

1. **Convert all values to the respective data types:**  If the value is an integer in XML, it will convert it to an integer in JSON. So sometimes the conversion changes the values. It will convert the value to scientific notation.

2. **Convert all values to a string:** In this case, if want some integer as an actual integer in JSON. Then using this config, we will not get it.

So I want to add type support on specific elements. 

For example:
`<root><id1 xsi:type="java.lang.String">1234</id1><id2 xsi:type="java.lang.Integer">1234</id2><id3>1234</id3></root>` 

Here the value for id1 will be converted to a string, id2 will be converted to an integer, id3 has no type information so default behavior will be applicable.

We can do without adding any configuration change. If xsi:type is present, it will try to convert only. @stleary Please suggest what is your opinion.

We can also use "string", "number", "boolean" for type representation and we can handle these cases from code.

johnjaylward:
@kumar529, that was why I suggested the hash map. The hashmap is to map the xsi:type to the java type.

The xsi:type would then be configurable per application. You don't need to pick how the type is represented in the xml now. Instead co figure it at runtime with converters. I don't think we should presume that all providers of XML will have the same xsi:type schema, nor that all consumers will want conversions done the same way.

stleary:
Hi @johnjaylward 

\>>>If the value is an integer in XML, it will convert it to an integer in JSON. **So sometimes the conversion changes the values.** 
\>>>It will convert the value to scientific notation.

Can it be confirmed that this is not a regression introduced in #453 or one of the subsequent BigNumber/Decimal/Integer commits?

kumar529:
@johnjaylward Yes, HashMap can work. The client will provide the HashMap at the time of initialization.

stleary:
@kumar529 This PR has several issues - 
- XMLTest has changed due to a recent commit, merge is needed.
- Additional XMLParserConfiguration, XML, and XMLTest changes are pending in #543, merge will probably be needed in a few days.
- We still don't know if this observation is due to an underlying bug: `sometimes the conversion changes the values.`

johnjaylward:
@stleary I think the "changing values" is due to not using the "keep strings" option (default parsing). The XSI processing as proposed here would basically be another flavor of "keep strings" but lets generators of XML specify the actual types instead us the tokener always using a string value.

kumar529:
@stleary Sure, I will make the change.
I think the idea of passing HashMap as configurations suggested by @johnjaylward looks good.
So I will try to do this.

src/main/java/org/json/XMLXsiTypeConverter.java
```diff
@@ -0,0 +1,5 @@
+package org.json;
```
johnjaylward, line 1: Please add the standard header here as well as some JavaDoc to describe the use

src/main/java/org/json/XMLParserConfiguration.java
```diff
@@ -106,9 +114,7 @@ public XMLParserConfiguration (final String cDataTagName) {
      */
     @Deprecated
     public XMLParserConfiguration (final boolean keepStrings, final String cDataTagName) {
-        this.keepStrings = keepStrings;
-        this.cDataTagName = cDataTagName;
-        this.convertNilAttributeToNull = false;
+        this(keepStrings, cDataTagName, false);
```
stleary, line 117: Please do not modify deprecated functions, or use them in your tests.

src/main/java/org/json/XMLParserConfiguration.java
```diff
@@ -56,6 +59,11 @@ of this software and associated documentation files (the "Software"), to deal
      */
     private boolean convertNilAttributeToNull;
 
+    /**
+     * This will allow type conversion for values in XML if xsi:type attribute is defined
+     */
+    public Map<String, XMLXsiTypeConverter<?>> xsiTypeMap;
```
johnjaylward, line 65: Sorry, I finally had time for a closer look at the configuration changes. With the new changes in #543, this should be private with a Getter and no Setter. Preferably the getter should return an "[Unmodifiable Map](https://docs.oracle.com/javase/6/docs/api/java/util/Collections.html#unmodifiableMap(java.util.Map))". To prevent wrapping the collection every time the getter is called, storing the unmodifiable map in this variable would be acceptable.

Note in the Getter javadoc that the map is "unmodifiable"

src/main/java/org/json/XMLParserConfiguration.java
```diff
@@ -56,6 +59,11 @@ of this software and associated documentation files (the "Software"), to deal
...
+
     /**
      * Default parser configuration. Does not keep strings (tries to implicitly convert
      * values), and the CDATA Tag Name is "content".
```
johnjaylward, line 69: In the default constructor here, please add a default [emptyMap](https://docs.oracle.com/javase/6/docs/api/java/util/Collections.html#emptyMap())

src/main/java/org/json/XMLParserConfiguration.java
```diff
@@ -129,6 +137,26 @@ public XMLParserConfiguration (final boolean keepStrings, final String cDataTagN
...
+     *      <code>false</code> to try and convert XML string values into a JSON value.
+     * @param cDataTagName <code>null</code> to disable CDATA processing. Any other value
+     *      to use that value as the JSONObject key name to process as CDATA.
+     * @param convertNilAttributeToNull <code>true</code> to parse values with attribute xsi:nil="true" as null.
+     *                                  <code>false</code> to parse values with attribute xsi:nil="true" as {"xsi:nil":true}.
+     * @param xsiTypeMap  <code>new HashMap<String, XMLXsiTypeConverter<?>>()</code> to parse values with attribute
+     *                   xsi:type="integer" as integer,  xsi:type="string" as string
+     *                    <code>null</code> to use default behaviour.
+     */
+    public XMLParserConfiguration (final boolean keepStrings, final String cDataTagName,
```
johnjaylward, line 153: This can be marked private instead of public.
Also, please change:
```java
this.xsiTypeMap = xsiTypeMap
```
to:
```java
this.xsiTypeMap = Collections.unmodifiableMap(new HashMap<String, XMLXsiTypeConverter<?>>(xsiTypeMap));
```
Our expected configuration should look something like this:
```java
XMLParserConfiguration xmlConfig = new XMLParserConfiguration().withKeepStrings(false).WithXsiTypeMap(myXsiMap);

// use the config
```

After marking this private, also update the "clone" method (currently line 165). Be sure to follow the comment that is included in the "clone" method. If changing the constructor above to do the shallow clone/unmodifiable map wrapping, you can probably just update the clone to call this private constructor.

Lastly, be sure to create the new "with" method that will take the XSI:Type conversion map.

stleary:
Manual gradlew build is failing due to javadoc problems, please fix.

@johnjaylward Do you have any more un-addressed issues?
Otherwise, after this PR is merged, I would like to follow up with a PR that removes clone() and the new XMLParserConfiguration ctor that @kumar529 just added. Any objections?

stleary:
**5 Sept 2020 Remaining issues (as I see them)**

Add test case:

image


Change to unmodifiableMap (or comment here if that is not possible):

image

johnjaylward:
yes, the test case is required as well as any code modifications needed to make the test case pass.

An example test case is here:
```java
    @Test
    public void testToJsonWithPartialXSITypeWhenTypeConversionEnabled() {
        String originalXml = "<root><asString xsi:type=\"string\">12345</asString><asInt>54321</asInt></root>";
        String expectedJsonString = "{\"root\":{\"asString\":\"12345\",\"asInt\":54321}}";
        JSONObject expectedJson = new JSONObject(expectedJsonString);
        Map<String, XMLXsiTypeConverter<?>> xsiTypeMap = new HashMap<String, XMLXsiTypeConverter<?>>();
        xsiTypeMap.put("string", new XMLXsiTypeConverter<String>() {
            @Override public String convert(final String value) {
                return value;
            }
        });
        JSONObject actualJson = XML.toJSONObject(originalXml, new XMLParserConfiguration().withXsiTypeMap(xsiTypeMap));
        Util.compareActualVsExpectedJsonObjects(actualJson,expectedJson);
    }
```

Also, I'd like at least a shallow copy of the maps used (not just the unmodifiable wrapper).

Here is a test case that I would like to pass:
```java
    @Test
    public void testXSITypeMapNotModifiable() {
        Map<String, XMLXsiTypeConverter<?>> xsiTypeMap = new HashMap<String, XMLXsiTypeConverter<?>>();
        xsiTypeMap.put("string", new XMLXsiTypeConverter<String>() {
            @Override public String convert(final String value) {
                return value;
            }
        });
       XMLParserConfiguration config = new XMLParserConfiguration().withXsiTypeMap(xsiTypeMap));
        xsiTypeMap.put("string", new XMLXsiTypeConverter<String>() {
            @Override public String convert(final String value) {
                return value;
            }
        });
        assertEquals("Config Conversion Map size is expected to be 0", 0, config.getXsiTypeMap().size());

        try {
            config.getXsiTypeMap().put("string", new XMLXsiTypeConverter<String>() {
                @Override public String convert(final String value) {
                    return value;
                }
            });
            fail("Expected to be unable to modify the config");
        } catch (Exception ex) { }
    }
```

kumar529:
@stleary 
The test case already there.
`@Test
    public void testToJsonWithPartialXSITypeWhenTypeConversionEnabled() {
        String originalXml = "<root><asString xsi:type=\"string\">12345</asString><asInt>54321</asInt></root>";
        String expectedJsonString = "{\"root\":{\"asString\":\"12345\",\"asInt\":54321}}";
        JSONObject expectedJson = new JSONObject(expectedJsonString);
        Map<String, XMLXsiTypeConverter<?>> xsiTypeMap = new HashMap<String, XMLXsiTypeConverter<?>>();
        xsiTypeMap.put("string", new XMLXsiTypeConverter<String>() {
            @Override public String convert(final String value) {
                return value;
            }
        });
        JSONObject actualJson = XML.toJSONObject(originalXml, new XMLParserConfiguration().withXsiTypeMap(xsiTypeMap));
        Util.compareActualVsExpectedJsonObjects(actualJson,expectedJson);
    }`

For the getXsiTypeMap method: Since the map has already been converted to an unmodifiable map during initialization, this getter will always return an unmodifiable map.

@johnjaylward If we want to create a shallow copy of xsiTypeMap before changing it to an unmodifiable map, then I think I have to convert this to a specific Map like TreeMap or HashMap. If we convert this to HashMap or TreeMap, then we will be forcing the user to supply only a specific type of Map.

We need to change the withXsiTypeMap method implementation like this.

`public XMLParserConfiguration withXsiTypeMap(final Map<String, XMLXsiTypeConverter<?>> xsiTypeMap) {
        XMLParserConfiguration newConfig = this.clone();
        Map<String, XMLXsiTypeConverter<?>> cloneXsiTypeMap = new HashMap<String, XMLXsiTypeConverter<?>>(xsiTypeMap);
        newConfig.xsiTypeMap = Collections.unmodifiableMap(cloneXsiTypeMap);
        return newConfig;
    }`

kumar529:
I can add the below test case.

```java
    @Test
    public void testXSITypeMapNotModifiable() {
        Map<String, XMLXsiTypeConverter<?>> xsiTypeMap = new HashMap<String, XMLXsiTypeConverter<?>>();
        XMLParserConfiguration config = new XMLParserConfiguration().withXsiTypeMap(xsiTypeMap);
        xsiTypeMap.put("string", new XMLXsiTypeConverter<String>() {
            @Override public String convert(final String value) {
                return value;
            }
        });
        assertEquals("Config Conversion Map size is expected to be 1", 1, config.getXsiTypeMap().size());

        try {
            config.getXsiTypeMap().put("boolean", new XMLXsiTypeConverter<Boolean>() {
                @Override public Boolean convert(final String value) {
                    return Boolean.valueOf(value);
                }
            });
            fail("Expected to be unable to modify the config");
        } catch (Exception ignored) { }
    }
```

johnjaylward:
> @stleary
> The test case already there.
> 
> ```java
>     @Test
>     public void testToJsonWithPartialXSITypeWhenTypeConversionEnabled() {
>         String originalXml = "<root><asString xsi:type=\"string\">12345</asString><asInt>54321</asInt></root>";
>         String expectedJsonString = "{\"root\":{\"asString\":\"12345\",\"asInt\":54321}}";
>         JSONObject expectedJson = new JSONObject(expectedJsonString);
>         Map<String, XMLXsiTypeConverter<?>> xsiTypeMap = new HashMap<String, XMLXsiTypeConverter<?>>();
>         xsiTypeMap.put("string", new XMLXsiTypeConverter<String>() {
>             @Override public String convert(final String value) {
>                 return value;
>             }
>         });
>         JSONObject actualJson = XML.toJSONObject(originalXml, new XMLParserConfiguration().withXsiTypeMap(xsiTypeMap));
>         Util.compareActualVsExpectedJsonObjects(actualJson,expectedJson);
>     }
> ```
> 
> For the getXsiTypeMap method: Since the map has already been converted to an unmodifiable map during initialization, this getter will always return an unmodifiable map.
> 
> @johnjaylward If we want to create a shallow copy of xsiTypeMap before changing it to an unmodifiable map, then I think I have to convert this to a specific Map like TreeMap or HashMap. If we convert this to HashMap or TreeMap, then we will be forcing the user to supply only a specific type of Map.
> 
> We need to change the withXsiTypeMap method implementation like this.
> 
> ```java
>     public XMLParserConfiguration withXsiTypeMap(final Map<String, XMLXsiTypeConverter<?>> xsiTypeMap) {
>         XMLParserConfiguration newConfig = this.clone();
>         Map<String, XMLXsiTypeConverter<?>> cloneXsiTypeMap = new HashMap<String, XMLXsiTypeConverter<?>>(xsiTypeMap);
>         newConfig.xsiTypeMap = Collections.unmodifiableMap(cloneXsiTypeMap);
>         return newConfig;
>     }
> ```
> 
> Please let me know if this is a good idea to change the map to HashMap.

that should work. 

> I can add the below test case.
> 
> ```java
>     @Test
>     public void testXSITypeMapNotModifiable() {
>         Map<String, XMLXsiTypeConverter<?>> xsiTypeMap = new HashMap<String, XMLXsiTypeConverter<?>>();
>         XMLParserConfiguration config = new XMLParserConfiguration().withXsiTypeMap(xsiTypeMap);
>         xsiTypeMap.put("string", new XMLXsiTypeConverter<String>() {
>             @Override public String convert(final String value) {
>                 return value;
>             }
>         });
>         assertEquals("Config Conversion Map size is expected to be 1", 1, config.getXsiTypeMap().size());
> 
>         try {
>             config.getXsiTypeMap().put("boolean", new XMLXsiTypeConverter<Boolean>() {
>                 @Override public Boolean convert(final String value) {
>                     return Boolean.valueOf(value);
>                 }
>             });
>             fail("Expected to be unable to modify the config");
>         } catch (Exception ignored) { }
>     }
> ```

The size check there should be 0 (zero) not 1 (one). The "Put" on line 4 of the test should be ineffective due to the shallow copy, which is what the test is for.

johnjaylward:
I updated my original `testXSITypeMapNotModifiable` example which now has the correct verification as well as comments explaining how I expect the test to work.

johnjaylward:
> @stleary
> The test case already there.
> 
> ```java
>     @Test
>     public void testToJsonWithPartialXSITypeWhenTypeConversionEnabled() {
>         String originalXml = "<root><asString xsi:type=\"string\">12345</asString><asInt>54321</asInt></root>";
>         String expectedJsonString = "{\"root\":{\"asString\":\"12345\",\"asInt\":54321}}";
>         JSONObject expectedJson = new JSONObject(expectedJsonString);
>         Map<String, XMLXsiTypeConverter<?>> xsiTypeMap = new HashMap<String, XMLXsiTypeConverter<?>>();
>         xsiTypeMap.put("string", new XMLXsiTypeConverter<String>() {
>             @Override public String convert(final String value) {
>                 return value;
>             }
>         });
>         JSONObject actualJson = XML.toJSONObject(originalXml, new XMLParserConfiguration().withXsiTypeMap(xsiTypeMap));
>         Util.compareActualVsExpectedJsonObjects(actualJson,expectedJson);
>     }
> ```

I do not see this test case in the change set. The XML I see tested is:
``` xml
<root><asString xsi:type="string">12345</asString><asInt xsi:type="integer">54321</asInt></root>
```
NOT
```xml
<root><asString xsi:type="string">12345</asString><asInt>54321</asInt></root>
```

kumar529:
@stleary @johnjaylward Please let me know if there is anything pending in this PR.

johnjaylward:
@kumar529, I believe that my concerns have been addressed. Sorry I did not notice your most recent commit until you commented.

johnjaylward (approved):

stleary:
Provides additional support for XML / JSON transformation via xsitype, by adding an xsiTypeMap to XMLParserConfiguration. This is an enhancement, not a bug fix. Allowed because additional flexibility is allowed for XML / JSON code.

**Risks**
Low

**Changes to the API?**
No

No

No

No

No

**Review status**
**APPROVED** 

PR #543, opened 2020-07-21
Refactor XMLConfiguration to use Builder Pattern

### LINKED ISSUE #541: XMLParserConfiguration needs to be opened up.
High priority - at least one pull request is waiting on this.
See comments in #540 and #412 
Existing XMLParserConfiguration constructors to be deprecated.
Add property setters.
Add default constructor.

johnjaylward:
Fixes #541 

Refactors the XML Configuration to use a builder pattern instead of public constructors with many parameters. This makes viewing the configuration easier to read and maintain.

**Risks**

The fields on the configuration have changed visibility from public to private and are now accessible via public getters. This means that a new release will not be a drop-in replacement and consumers of this library will need to correct code that is directly referencing those fields. This should hopefully be minimal impact as the fields were readonly (i.e. hopefully no one was actually using them).

**Changes to the API?**

Yes, public fields were marked private and constructors were deprecated. New methods were put in place for managing the XML Configuration.

Yes

I don't believe we have any documentation outside of the Java-docs referencing this API. 

No, but unit tests were updated to use the new API.

Yes, as mentioned above.

**Review status**

**APPROVED**

### DIFF
--- a/src/main/java/org/json/XML.java
+++ b/src/main/java/org/json/XML.java
@@ -286,7 +286,7 @@ public class XML {
                     if (x.next() == '[') {
                         string = x.nextCDATA();
                         if (string.length() > 0) {
-                            context.accumulate(config.cDataTagName, string);
+                            context.accumulate(config.getcDataTagName(), string);
                         }
                         return false;
                     }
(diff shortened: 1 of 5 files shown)


stleary:

johnjaylward:
@stleary, would you mind using tags to make it easier to see which PR are up for review and which are ready to commit? Not that we usually have a lot, but it may make it easier to see at a glance from the main PR screen.

PR #529, opened 2020-06-16
Added putAll(Collection) and putAll(Array) methods.

ethauvin:
I've added two methods to add/put all of the elements of an array/collection to the `JSONArray`.

Mostly to avoid inefficient code like:

```java
JSONArray jsonArray;
...
if (myArray != null) {
    jsonArray = new JSONArray(myArray);
} else {
    jsonArray = new JSONArray();
}
```
You can now do:
```java
JSONArray jsonArray = new JSONArray();
...
if (myArray != null && !myArray.isEmpty()) {
    jsonArray.putAll(myArray);
}
```
It is also very useful if you need to add elements from multiple arrays/collections. For example:
```java
int[] myInts = { 1, 2, 3, 4, 5 };
List<String> myList = Arrays.asList("one", "two", "three", "four", "five");
```
Instead of:
```java
JSONArray jsonArray = new JSONArray(myInts);

for (String s : myList) {
    jsonArray.put(s);
}
```
Simply do:
```java
JSONArray jsonArray = new JSONArray(myInts);
jsonArray.putAll(myList);
```

There's no real code change, just things moved around. I did not implement tests since the constructors are already tested.

Let me know if you have any questions.

PR #552, opened 2020-07-27
Updates for JSONArray.putAll methods

johnjaylward:
The new `putAll` methods did not implement a way to concatenate JSONArray objects. It was an oversight during the review process, but should probably be allowed. This also adds in a shallow copy constructor for JSONArrays so the solution is a bit more holistic. I also added support for adding Generic `Iterable` objects which don't have a `size` function.

* Adds a copy constructor for JSONArray
* Updates the JSONArray.addAll(Object) method to be more lenient
* Adds support for JSONArray.putAll of generic Iterables
* Adds support for JSONArray.putAll of another JSONArray
* Updates some JavaDoc for clarification.
* Updates `putAll` methods to match other `put` methods and not wrap the items in the collection/array

**Risks**
Low. Just closes some gaps that were missed in the initial PR

**Changes to the API?**
New methods added, but no breaking changes

No

Yes. Notes should be added about the new `putAll` methods and information on expected use cases

No. New tests were added to verify existing functionality and show new functions

Yes. The internal `addAll` methods created in #529 were modified to optionally wrap the contents of the collection/array instead of always wrapping them.

**Review status**
**APPROVED**

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -1242,12 +1242,12 @@ public class JSONArray implements Iterable<Object> {
      * Put an array's elements in to the JSONArray.
      *
      * @param array
-     *            Array. If the parameter passed is null, or not an array, an
+     *            Array. If the parameter passed is null, or not an array or Iterable, an
      *            exception will be thrown.
      * @return this. 
      *
      * @throws JSONException
-     *            If not an array or if an array value is non-finite number.
+     *            If not an array, JSONArray, Iterable or if an value is non-finite number.
      * @throws NullPointerException
      *            Thrown if the array parameter is null.
      */
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -194,15 +194,16 @@ public class JSONArray implements Iterable<Object> {
     /**
      * Construct a JSONArray from another JSONArray. This is a shallow copy.
      *
-     * @param collection
-     *            A Collection.
+     * @param array
+     *            A array.
      */
     public JSONArray(JSONArray array) {
         if (array == null) {
             this.myArrayList = new ArrayList<Object>();
         } else {
-            this.myArrayList = new ArrayList<Object>(array.length());
-            this.addAll(array.myArrayList);
+            // shallow copy directly the internal array lists as any wrapping
+            // should have been done already in the original JSONArray
+            this.myArrayList = new ArrayList<Object>(array.myArrayList);
         }
     }
 
@@ -1213,7 +1214,7 @@ public class JSONArray implements Iterable<Object> {
      * Put an Iterable's elements in to the JSONArray.
      *
      * @param iter
-     *            A Collection.
+     *            An Iterable.
      * @return this. 
      */
     public JSONArray putAll(Iterable<?> iter) {
@@ -1229,7 +1230,9 @@ public class JSONArray implements Iterable<Object> {
      * @return this. 
      */
     public JSONArray putAll(JSONArray array) {
-        this.addAll(array.myArrayList);
+        // directly copy the elements from the source array to this one
+        // as all wrapping should have been done already in the source.
+        this.myArrayList.addAll(array.myArrayList);
         return this;
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/json/JSONArray.java
```diff
@@ -177,6 +177,35 @@ public JSONArray(Collection<?> collection) {
...
+        this.addAll(iter);
+    }
+
+    /**
+     * Construct a JSONArray from another JSONArray. This is a shallow copy.
+     *
+     * @param collection
+     *            A Collection.
+     */
+    public JSONArray(JSONArray array) {
```
stleary, line 200: Do we support this for JSONObject? If not, we should not support it for JSONArray either.

src/main/java/org/json/JSONArray.java
```diff
@@ -210,6 +243,11 @@ public JSONArray(int initialCapacity) throws JSONException {
     	this.myArrayList = new ArrayList<Object>(initialCapacity);
     }
 
+    @Override
```
stleary, line 246: Please remove clone() method.

src/main/java/org/json/JSONArray.java
```diff
@@ -177,6 +177,35 @@ public JSONArray(Collection<?> collection) {
         }
     }
 
+    /**
+     * Construct a JSONArray from an Iterable. This is a shallow copy.
+     *
+     * @param iter
+     *            A Iterable collection.
+     */
+    public JSONArray(Iterable<?> iter) {
```
stleary, line 186: I don't think there is a use case for this, but let me know if you disagree.

stleary:
No objection to enhancing add_all() and put_all() a bit.
A strong case is needed to justify new constructors. 
Clone() looks like it was added by accident, code seems to work fine without it.

src/main/java/org/json/JSONArray.java
stleary: After considering, withdrawing the constructor comments. Should be fine to add these methods.

src/main/java/org/json/JSONArray.java
stleary: Resolving, not a critical issue.

stleary:
If clone() is not removed during that time, we can remove it next time the code is updated, unless there is some reason it needs to stay.

johnjaylward:
do you still want `clone` removed?

stleary:
> do you still want clone removed?

Yes, it does not seem to be related to the rest of the PR.

johnjaylward:
@stleary in reference to your note about parity with JSONObject constructors, we currently have this one that does a partial copy: https://github.com/stleary/JSON-java/blob/master/src/main/java/org/json/JSONObject.java#L197

Should I add another one that does a full copy?

johnjaylward:
thinking the constructor would look like this:
```java

    /**
     * Construct a JSONObject from another JSONObject.
     *
     * @param jo
     *            A JSONObject.
     * @throws NullPointerException
     *            If the {@link jo} JSONObject parameter is <code>null</code>
     */
    public JSONObject(JSONObject jo) {
        this(jo.map.size());
        this.map.putAll(jo.map);
    }
```

src/main/java/org/json/JSONArray.java
```diff
@@ -1534,12 +1593,25 @@ private void addAll(Collection<?> collection) {
         }
```
johnjaylward, line 1593: I just noticed the JSONObject.wrap method here (and in the other `addAll` method), and I think we may have an issue with consistency. If you notice both the `JSONObject.put(String, Object)` and `JSONArray.put(Object)` methods do not do any "wrap" calls to the values. However, the constructors (before the new `addAll` mothods) both wrap values. I feel like this may have been an oversight at some point. Changing the wrap to happen consistently in the `put` methods breaks a number of tests.

@stleary some guidance on what to do in this case would be appreciated.

johnjaylward: so, to this effect, I actually worked around it in my own copy by adding the wrap calls to the `put` methods

stleary:
@johnjaylward Thanks for catching these items, I added 2 new issues to track them away from this PR.

johnjaylward:
@stleary , I changed the way the `allAll` and hence the `putAll` methods work in the latest commit. I did this so functionality of `putAll` would be consistent with `put`. You may want to reset the review timer so interested people can check it out.

stleary:
Thanks, we still have 2 days left on the comment window, should be fine.

PR #558, opened 2020-08-17
Improve JSONObject's equals function

ace-zhang18:
Now compares classtype and data instead of object ID.

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -128,8 +128,13 @@ public class JSONObject {
          *         null.
          */
         @Override
-        public boolean equals(Object object) {
-            return object == null || object == this;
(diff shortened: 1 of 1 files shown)


ace-zhang18:
I was working with this library, very useful by the way, and discovered that it was failing equality checks on JSONObjects with the same data. To my knowledge Java objects should not base equality on having the same ID, lest they only ever pass an equality check when compared with itself. Thus I devised a simple change that will instead check for matching class identity and internal data instead of just leaving it at seeing if the object matches itself.

johnjaylward:
The method you modified is not for the JSONObject class, but for the NULL class.

To compare json objects with this library, please use the `similar` method

stleary:
Closed due to lack of activity.

PR #562, opened 2020-09-03
Create module-info.java

jaghaimo:
When using this library in a modular app you need to add to `module-info.java`:
```
requires hjson;
```
Which will raise a warning, e.g. `Name of automatic module 'hjson' is unstable, it is derived from the module's file name`, as obviously it is an automatic module. By adding a `module-info.java` we formalize the name as `hjson` (which makes it backward compatible with current automatic module resolution).

More on automatic modules: https://stackoverflow.com/a/46742802 correctly, this should be all that's needed in order for hjson

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ repositories {
     }
 
     maven {
-        url = uri('http://repo.maven.apache.org/maven2')
+        url = uri('https://repo.maven.apache.org/maven2')
     }
 }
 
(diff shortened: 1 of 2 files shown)


jaghaimo:
This is not possible at the moment, need -source 9 or higher. Feel free to close the MR, sorry :(

jaghaimo:
Pretend you saw nothing, wrong project.

PR #575, opened 2020-11-19
Fix similar compare numbers

### LINKED ISSUE #573: Support equality testing
It would be great if the library would support `equals` for `JSONObject`.  

An important use case would be equality/regression testing. See [here ](https://github.com/holgerbrandl/kalasim/blob/b5d9416d48ce959b4a3e637db54cd93fe24968ce/src/test/kotlin/org/github/holgerbrandl/kalasim/examples/kalasim/SalabimExampleTests.kt#L65) for an example which hacks around the problem by using toString (which somehow works in this particular case but is certainly not good practise).

Another use case would be the use of `JSONObject` as map keys.

Since there seems to be just a single data fields (`map`) in `JSONObject`, it may be possible to simple delegate to `HashMap.equals`. Clearly, also JSONArray would need to be equipped with an `equals` (see  https://stackoverflow.com/questions/4082416/equality-between-2-hashmap or https://www.baeldung.com/java-compare-hashmaps).

Alternatively, the docs https://github.com/stleary/JSON-java#notes shall detail out better why an `equals` implementation can not be provided.

johnjaylward:

Fixes #573. Similar currently requires exact types match for numeric comparisons, but in JSON there are just "numbers", not specific `int`, `long`, `float`, etc.

**Risks**

Low

**Changes to the API?**

No.

Yes.

No.

No, we did not have a unit test for this case. A new unit test was added to demonstrate the issue.

New methods were added to accommodate the new check. They currently are `private` or have `default` visibility. Some areas of unrelated code were modified to take advantage of the new methods where applicable

**Review status**

**APPROVED**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ repositories {
 }
 
 dependencies {
-    testImplementation 'junit:junit:4.12'
+    testImplementation 'junit:junit:4.13.1'
     testImplementation 'com.jayway.jsonpath:json-path:2.1.0'
     testImplementation 'org.mockito:mockito-core:1.9.5'
 }
--- a/pom.xml
+++ b/pom.xml
@@ -80,7 +80,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.12</version>
+            <version>4.13.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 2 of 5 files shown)


johnjaylward:
Actually, let me make one more change. The BigDecimal compare should probably only be used if the types are different and doesn't implement compare

stleary:

PR #577, opened 2020-12-04
Added clear() methods to JSONObject and JSONArray

stranck:
I noticed that there wasn't this easy way to clear all the childs of a JSONObject or a JSONArray, so I added it.
Changes in code are really simple

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -568,6 +568,14 @@ public class JSONArray implements Iterable<Object> {
     }
 
     /**
+     * Removes all of the elements from this JSONArray.
+     * The JSONArray will be empty after this call returns.
+     */
+    public void clear() {
+        return this.map.clear();
+    }
+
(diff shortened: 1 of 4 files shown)


stleary:
@stranck Thanks for submitting this PR. The submitter is usually asked to provide a use case or justification, but I guess in this case it is self-evident. No objection to proceeding after you add the appropriate unit tests.

stranck:
I should've added them :)
Tell me if they're alright and thank you for being so fast in replying ahahahhaha

stleary:
This is an enhancement to include a missed clear() method, not a bug fix. The API includes JSONObject and JSONArray methods that recognize they are containers, so adding this method makes sense. 

**Risks**
Low

**Changes to the API?**
New public methods are added to JSONObject and JSONArray

No

No

No, new unit tests were added

No

**Review status**
**APPROVED**

stleary:
Starting 3 day comment window

PR #583, opened 2021-01-27
Checked the length of key for checker framework

ek08:
Checked the length of key IN JSONObject  for Checker Framework

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -1558,7 +1558,7 @@ public class JSONObject {
         // if the first letter in the key is not uppercase, then skip.
         // This is to maintain backwards compatibility before PR406
         // (https://github.com/stleary/JSON-java/pull/406/)
-        if (Character.isLowerCase(key.charAt(0))) {
+        if (key.length() > 0 && Character.isLowerCase(key.charAt(0))) {
             return null;
         }
         if (key.length() == 1) {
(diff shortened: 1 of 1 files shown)


stleary:
JavaCheck fails due to not being able to detect that zero-length keys are disallowed. I don't think there is any change to functionality.

**Risks**
Low

**Changes to the API?**
No

No

No

No

No

**Review status**
**APPROVED**

johnjaylward:
I don't see what this change is adding to the code. The key can not be 0 length at that spot. Also, I would argue that the logic chosen for the check is wrong.

I'd argue it should look like this if you feel the check needs to be added

```java
if (key.length() == 0 || Character.isLowerCase(key.charAt(0))) {
    return null;
}
```

johnjaylward (changes requested):
Check is not valid

ek08:
> I don't see what this change is adding to the code. The key can not be 0 length at that spot. Also, I would argue that the logic chosen for the check is wrong.
> 
> I'd argue it should look like this if you feel the check needs to be added
> 
> ```java
> if (key.length() == 0 || Character.isLowerCase(key.charAt(0))) {
>     return null;
> }
> ```

Thanks for the correction. I agree with you that the length of key cannot be 0 length at that spot. So the logic of the check should be changed to 
```
if (key.length() == 0 || Character.isLowerCase(key.charAt(0)))

stleary:
Approval restored.

Issue #563, opened 2020-09-19
JSONPointer: Encoding quotes not required?


jimblackler:
JSONPointer encodes quotes as \\"

https://github.com/stleary/JSON-java/blob/7abd89b4bcc321152c178af5504da7ea4ac66693/src/main/java/org/json/JSONPointer.java#L271

However I can't find a requirement for that in https://tools.ietf.org/html/rfc6901#section-3 or elsewhere.

It causes a conflict in an application that expects the canonical form of the URL not to be escaped in this way. For use in fragment identifiers RFC 3986 allows %22 to encode ".

stleary:
Please see  [https://tools.ietf.org/html/rfc6901#section-5](https://tools.ietf.org/html/rfc6901#section-5). 
Does this address your concern?

jimblackler:
Is that saying that those chracters should be escaped when represented as a JSON string literal? (That would be something handled by clients not by JsonPointer()). Does the requirement apply otherwise? 

If the code at https://github.com/stleary/JSON-java/blob/7abd89b4bcc321152c178af5504da7ea4ac66693/src/main/java/org/json/JSONPointer.java#L271  is executed and I JSON encode the resulting string I'll get two sets of encoding on the string characters.

jimblackler:
The motivation for the query is a clash between JSON Schema offical test suite requirements and org.json.JSONPointer implementation.

https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/master/tests/draft7/ref.json#L363

The test "refs with quote" creates a reference to the value at definitions/foo"bar; in JSON Schema references are required to by JSON Pointers, first escaped to the JSON Pointer requirements and then escaped to URI requirements to represent them as fragments (after the #). The fact then that the reference is stored as a JSON String will cover the requirements in https://tools.ietf.org/html/rfc6901#section-5.

https://tools.ietf.org/html/draft-handrews-json-schema-01#section-8.3
https://tools.ietf.org/html/draft-handrews-json-schema-01#section-8.3.2

If org.json.JSONPointer is used to contruct this reference it first will escape the " to \" which will then result in a reference foo\"bar, then URI encoded to foo%5C%22bar

stleary:
Are you requesting a change to the code?

jimblackler:
Yes; if the exising maintainers agree with me that the current behavior is incorrect.

johnjaylward:
[...]

Issue #561, opened 2020-08-25
Queries involving keys containing four backslashes ("\\\\") failing. 


jimblackler:
I encountered a problem with an application when JSON keys contained four backslashes (\\\\).

After a lot of debugging I found that JSONPointer (used in my app) was returning null when queried. Stepping through the code it seems that the token used for the key may be double unescaped; I saw the same token unescaped() on line 173 when the JSONPointer was built, then on line 213 when the pointer was used in a query.

https://github.com/stleary/JSON-java/blob/master/src/main/java/org/json/JSONPointer.java#L173
https://github.com/stleary/JSON-java/blob/master/src/main/java/org/json/JSONPointer.java#L213

This example test will fail on the trunk version of the library:

https://github.com/jimblackler/JSON-java/commit/8d6e1d85ff581e0934b839df04554b10f6f9ab29#diff-b4df778ff5909ef85877e003ed15b3beR279

(Note there are eight double slashes in the literal in the test, because it's escaped for the Java string literal)

I'm not 100% sure of my understanding of the issue or the correct behavior. I'd be grateful if someone who is familiar with the practices handling these strings could comment or confirm, and also advise if the fix in the patch above is something that I should submit.

johnjaylward:
@erosb do you have time to look at this one?

jimblackler:
Thanks guys. I found that the pattern ~~0 fails for the same reason.

stleary:
Will wait a few more days in case @erosb is available to look into this.

erosb:
Guys, unfortunately I can't see any chance of looking into this in detail.

stleary:
@erosb No worries, thanks for responding

stleary:
Looks like this will need to be addressed without the benefit of expert advice. 

It seems likely to me that the implementation is only intended to handle Java strings with quotes and double backslashes, both of which happen to be examples in [RFC6901](https://tools.ietf.org/html/rfc6901).  Arbitrary sequences of backslashes in keys would appear to be unsupported. There may also be an issue with the code calling `JSONPointer.unescape()` twice in succession, from the JSONPointer() constructor, and again from queryFrom(). This does not seem correct.  The escape() method on the other hand is only called from toString(). 

If anyone wants to see if they can address this without affecting existing functionality, they are welcome to go ahead.

niteshbheda:
I am facing a similar issue in Values when we a backslash escape character supporting " inside the value then .op function fails to return the escape character. For instance: 

[...]

Issue #551, opened 2020-07-27
Implement opt/get methods for accumulated values


johnjaylward:
Currently we have an `accumulate` method on `JSONObject`, however, we don't have an easy way to get back out a consistent value (i.e. JSONArray).

One area where this is most painful is in XML to JSON conversion where some XML documents may contain a single element, while other documents may contain many.

Example 1:
```xml
<Student>
  <name>jack</name>
  <subjects>
    <class>english</class><grade>98</grade>
  </subjects>
  <subjects>
    <class>english</class><grade>98</grade>
  </subjects>
</Student>
```
Example 2:
```xml:
<Student>
  <name>jack</name>
  <subjects>
    <class>english</class><grade>98</grade>
  </subjects>
</Student>
```

This currently leads to very messy code that has to check the type of value parsed:
```java
void someFunction() {
  JSONObject jo = XML.toJSONObject(someXmlSource);
  JSONObject student = jo.getJSONObject("Student");
  if ( student.optJSONArray("subjects") ==null ) {
      processSubject(student.optJSONObject("subjects"));
  } else {
      processSubjects(student.optJSONArray("subjects"));
  }
}

void processSubjects(JSONArray subjects) {
  if (subjects == null || subjects.length() == 0) return;
  for(int i=0; i<subjects.length; i++) {
    processSubject(subjects.getJSONObject(i));
  }
}

void processSubject(JSONObject subject) {
  // some implementation
}
```

It would be nice if there was a convenience function in the JSONObject to get values that are accumulated:
```java
// In JSONObject

/** Always returns a JSONArray, If the key doesn't exist or the value is `null`, the array will be empty.
* If the key holds a non-array/non-collection value, then it is placed in a JSONArray
* before returning.
* If the value is a JSONArray, then it is directly returned.
* If the value is directly convertible to a JSONArray (i.e. a JAVA Array or Collection), then it is
* converted to a JSONArray.
* @param key The key in the JSONObject to fetch
* @returns a JSONArray regardless of the value of the key.
*/
public JSONArray optAccumulated(String key) {
  Object o = opt(key);
  if (JSONObject.NULL.equals(o)) {
    return new JSONArray();
  }
  if (o instanceof JSONArray) {
    return o;
  }
  if (o instanceof Collection || o.getClass().isArray()) {
    return new JSONArray(o);
  }
  JSONArray ret = new JSONArray();
  ret.put(o);
  return ret;
}


[...]

PR #585, opened 2021-02-15
Added JSONObject.expand()

fossterer:
- Adds _JSONObject.expand()_ implementation
- **Reference:** https://blog.fossterer.com/Technical/contributing-jsonutils-getexpandedjson-org-json.html

**Risks**
Low

**Changes to the API?**
Adds new

No

Yes

No

No

**Review status**


src/main/java/org/json/JSONUtils.java
```diff
@@ -0,0 +1,178 @@
...
+	 * 
+	 * <pre>
+	 * {		{		{
+	 *   a: b,	  a: b,	  	  a: b,
+	 *   c: d1,	  c: d2,  	  c: d3,
+	 *   e: f1	  e: f2	  	  e: null
+	 * }		}		}
+	 * </pre>
+	 */
+	public static List<JSONObject> getExpandedJSON(JSONObject incomingJSON) {
```
fossterer, line 108: You may start reading from here.

fossterer:
@stleary You may start reading from [Line 108](https://github.com/stleary/JSON-java/pull/585/files#diff-d37b32c2eb8e100341e9cf7ea26352950c70874d5071112abaa81e28d98fccd4R108)

The rest are tests and test resources

johnjaylward:
When I think of a "utility" class, I generally think of 2 use cases.

1. A class with commonly used functions that are not part of the existing API but would be useful under multiple use-cases or across multiple domains.
2. A class with functions in it that don't really fit anywhere else because their scope of use is too narrow to fit into the above "general use" class, however they may need to be used in a few key, distinct points of an application.

In general, I highly dislike utility classes of any kind as they feel like an anti-pattern (Why wasn't the function just added to the API in the first place). The exception being when creating a utility class to work with APIs that are not managed by your code.

An instance of a good use of a utility class is something like `StringUtil` where you are adding helping functions to a class that you can not modify.

I've never personally felt the need for the functions like the one this PR uses, and I've never seen a request for the feature here before. While it looks like a lot of effort and good testing went into the feature, this feels more like the second use of "utility" above. But again, why is this a "utility class" instead of just being part of the API?

Maybe I'm wrong, and this is a feature hotly wanted, but my gut based on the history of this project tells me that this is more of a special case and doesn't fit the "general use case" that this library tries to adhere to.

Now, our JSONObject and JSONArray classes certainly fit into "God Class" anti-pattern as there are definitely functions in there that would be more appropriate to be elsewhere (in the tokener for instance, or perhaps a package-visible class). However, given the current design, If these new functions were to be accepted, I'd probably look into making these instance methods on their respective classes instead. i.e. Move `JSONUtils.getExpandedJSON(JSONObject incomingJSON)` to `JSONObject.expand()` and move `JSONUtils.getExpandedJSON(List<JSONObject> incomingJSONList)` to be a static function on `JSONObject` or just drop it altogether since it's no more than a wrapper around the first function.

fossterer:
Hello,

Just notifying that I attended to the suggestions given above. All changes have been pushed

stleary:
Hi @fossterer, Agreed, nice coding and tests, but in order to progress this PR, the description needs to include text explaining how this feature is used and how it aligns with the project goals. I can't come up with a use case.  If it only serves some purpose in your project, then it should probably be implemented as part of your project.

fossterer:
Glad to have your comments @stleary. I'm a fan of this nimble library and always looked forward to give back something to improve.
As I solved this challenge I faced one day in a generic form so close to underlying library (`JSONObject`) I took that as an opportunity to contribute.

So this project may continue to be light-weight, this code may not be welcome here. Do any JSONObject-friendly projects come to your mind that can be a new home for this implementation? Please respond and then you may close this PR.

Thank you

stleary:
@fossterer  Although expand() was not the right match, issues with the label "Coding Issue" need someone to fix/enhance the code:
#563 
#561 
#551
The JSONPointer issues are interesting because the original author is not available and no one else has come up to speed yet. Consider taking on one of these issues if you want.

PR #587, opened 2021-02-20
JSONTokener implements Closeable

nguyensach:
What problem does this code solve?

- When create JSONTokener's object by String (Ex: new JSONTokener("String")), JSONTokener's reader is not closed 

Risks
High

Changes to the API?
Yes

Will this require a new release?
Yes

Should the documentation be updated?
Yes

Does it break the unit tests?
No. I run unit tests in my local.

Was any code refactored in this commit?
No

Review status

### DIFF
--- a/src/main/java/org/json/JSONTokener.java
+++ b/src/main/java/org/json/JSONTokener.java
@@ -1,11 +1,6 @@
 package org.json;
 
-import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.Reader;
-import java.io.StringReader;
+import java.io.*;
 
 /*
 Copyright (c) 2002 JSON.org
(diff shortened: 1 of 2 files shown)


stleary:
Please add or update a unit test that exercises the new JSONTokener.close() method. Otherwise, no objection on my part to including this change.

nguyensach:
@stleary
I added a unit test for the new JSONTokener.close() method by bellow commit.
https://github.com/stleary/JSON-java/pull/587/commits/3e381bf6c20de6080a0b5b83c32d215af0cd9d40

johnjaylward:
In my opinion, The string reader close method not being called doesn't pose a risk as it's a no-op.

The changes to the public apis worry me more as the tokens may now close a reader unexpectedly by a caller.

nguyensach:
@johnjaylward
I got it! Thank you!

PR #588, opened 2021-02-28
JSONPointer should not process reverse solidus or double-quote chars in tokens

fossterer:
1. As never defined in https://tools.ietf.org/html/rfc6901#section-3 and
2. as discussed in https://github.com/stleary/JSON-java/issues/563,

backslashes(\\) and quotes(") are not to be treated any special by JSONPointer

### DIFF
--- a/src/main/java/org/json/JSONPointer.java
+++ b/src/main/java/org/json/JSONPointer.java
@@ -187,10 +187,11 @@ public class JSONPointer {
         this.refTokens = new ArrayList<String>(refTokens);
     }
 
+    /**
+     * @see https://tools.ietf.org/html/rfc6901#section-3
+     */
     private static String unescape(String token) {
-        return token.replace("~1", "/").replace("~0", "~")
-                .replace("\\\"", "\"")
-                .replace("\\\\", "\\");
+        return token.replace("~1", "/").replace("~0", "~");
     }
 
     /**
@@ -263,16 +264,15 @@ public class JSONPointer {
     /**
      * Escapes path segment values to an unambiguous form.
      * The escape char to be inserted is '~'. The chars to be escaped 
-     * are ~, which maps to ~0, and /, which maps to ~1. Backslashes
-     * and double quote chars are also escaped.
+     * are ~, which maps to ~0, and /, which maps to ~1.
      * @param token the JSONPointer segment value to be escaped
      * @return the escaped value for the token
+     * 
+     * @see https://tools.ietf.org/html/rfc6901#section-3
      */
     private static String escape(String token) {
         return token.replace("~", "~0")
-                .replace("/", "~1")
-                .replace("\\", "\\\\")
-                .replace("\"", "\\\"");
+                .replace("/", "~1");
     }
 
     /**
--- a/src/test/java/org/json/junit/JSONPointerTest.java
+++ b/src/test/java/org/json/junit/JSONPointerTest.java
@@ -117,14 +117,24 @@ public class JSONPointerTest {
         assertSame(document.get("m~n"), query("/m~0n"));
     }
 
+    /**
+     * We pass backslashes as-is
+     * 
+     * @see https://tools.ietf.org/html/rfc6901#section-3
+     */
     @Test
-    public void backslashEscaping() {
-        assertSame(document.get("i\\j"), query("/i\\\\j"));
+    public void backslashHandling() {
+        assertSame(document.get("i\\j"), query("/i\\j"));
     }
 
+    /**
+     * We pass quotations as-is
+     * 
+     * @see https://tools.ietf.org/html/rfc6901#section-3
+     */
     @Test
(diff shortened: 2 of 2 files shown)


stleary:
This PR claims that the JSONPointer implementation does not follow the spec [RFC-6901](https://tools.ietf.org/html/rfc6901)

**Risks**
**High**  This code has been in place for several years

**Changes to the API?**
API behavior is changed

No

No

Yes

No

**Review status**
Pending **30 day window** for comments

stleary:
@erosb @jimblackler @johnjaylward Please comment if you have any thoughts or concerns.

johnjaylward:
Section 5 of the RFC states:

https://tools.ietf.org/html/rfc6901#section-5
> 
>  5.  JSON String Representation
> 
>    A JSON Pointer can be represented in a JSON string value.  Per
>    [RFC4627], Section 2.5, all instances of quotation mark '"' (%x22),
>    reverse solidus '\' (%x5C), and control (%x00-1F) characters MUST be
>    escaped.
> 
>    Note that before processing a JSON string as a JSON Pointer,
>    backslash escape sequences must be unescaped.
> 

So, I take this to mean that the current implementation may be correct.

See the continuation of section 5 for a test document and test cases.

johnjaylward:
@fossterer , can you add an explicit test case that matches the ones in the RFC?

| Test JSON Pointer | Expected Result |
| ----------------- | ---------------- |
| `""`              | `// the whole document` |
| `"/foo"`          | `["bar", "baz"]` |
| `"/foo/0"`        | `"bar"` |
| `"/"`             | `0` |
| `"/a~1b"`         | `1` |
| `"/c%d"`          | `2` |
| `"/e^f"`          | `3` |
| `"/i\\j"`         | `5` |
| `"/k\"l"`         | `6` |
| `"/ "`            | `7` |
| `"/m~0n"`         | `8` |

and this one (because GitHub doesn't like the `|` in table values)
* `"/g|h"` -> `4` 

The paths are expressed as JSON strings, not Java strings. so  `"/k\"l"`  expressed as a java string should look like  `"/k\\\"l"`

johnjaylward:
Update my previous comment.

Also @fossterer, if some of these test cases already exist, can you confirm that they are working as expected?

johnjaylward:
linking original implementation discussion: #218, #222 and #292

johnjaylward (approved):
OK, after reading through the comments by @jimblackler in #563 and re-reading the RFC section 5 a few times, I'm reaching the conclusion that the `escape` and `unescape` functions in `JSONPointer` are wrong, and this PR corrects them.

Test cases like this one:
```java
@Test
    public void quotationEscaping() {
        assertSame(document.get("k\"l"), query("/k\\\\\\\"l"));
    }
```

are essentially double escaping the "query" string. The RFC states that the query only needs to be escaped when stored as JSON. For the purpose of the test, the query is not a JSON String and is instead a Java String, and should not have the escaping applied.

The above test should be changed to (as provided in this PR):
```java
@Test
    public void quotationEscaping() {
        assertSame(document.get("k\"l"), query("/k\"l"));
    }
```


As such, I'm giving my stamp of approval for this PR and I believe it matches the "the bug is worse than the fix" requirement.

johnjaylward:
@stleary can we try to get this in the next release too? @jimblackler has been waiting on it for a while now.

stleary:
We can include it in the release at the end of this week. I would still like to give @erosb a few more days to respond, as he is the original JSONPointer author.

erosb (approved):
Agreed, based on section 3 and 5 of RFC 6901, this is the correct behavior. Thanks for the fix.

stleary:
Merging the code but someone needs to complete the unit tests per the comments in this PR, before the next release.

fossterer:
@stleary All the tests as indicated by @johnjaylward exist already in the test suite!

Reading form the [highlighted line](https://github.com/stleary/JSON-java/blob/d6ccc64c794bd54bfa70bd510e217cdb72c10268/src/test/java/org/json/junit/JSONPointerTest.java#L110) you can spot all the `m~`, `k\"l`, `foo` etc. cases are already accounted for

johnjaylward:
Thanks for taking the time to double check.

stleary:
> @stleary All the tests indicated by @johnjaylward exist already in the test suite!
> 
> Reading from the [highlighted line](https://github.com/stleary/JSON-java/blob/d6ccc64c794bd54bfa70bd510e217cdb72c10268/src/test/java/org/json/junit/JSONPointerTest.java#L110) you can spot all the `m~`, `k\"l`, `foo` etc. cases

Thanks for pointing this out, I will confirm.

stleary:
Some of the tests exist, some do not. Most of the tests do not clearly call out the expected result. For example, if both document.get() and query() return the same incorrect result, the test will not catch it. Will open an issue to track this to completion.

Issue #537, opened 2020-07-18
Issue while converting xml to json


venkat-natsoft:
xml: <tag>Hypersegmented &#X7C; Bld-Ser-Pla</tag>

Caller method:
JSONObject xmlJSONObj = XML.toJSONObject(xml, true);

Error:
Caused by: java.lang.NumberFormatException: For input string: "X7C"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65) ~[na:1.8.0_201]
	at java.lang.Integer.parseInt(Integer.java:580) ~[na:1.8.0_201]
	at java.lang.Integer.parseInt(Integer.java:615) ~[na:1.8.0_201]
	at org.json.XML.unescape(XML.java:200) ~[json-20170516.jar:na]
	at org.json.XML.parse(XML.java:398) ~[json-20170516.jar:na]
	at org.json.XML.parse(XML.java:403) ~[json-20170516.jar:na]
	at org.json.XML.toJSONObject(XML.java:485) ~[json-20170516.jar:na]

Error at calle:
at org.json.XML.  public static String unescape(String string) 

Error prone code piece
 if (entity.charAt(1) == 'x') {
                            // hex encoded unicode
                            cp = Integer.parseInt(entity.substring(2), 16);
                        } else {
                            // decimal encoded unicode
                            cp = Integer.parseInt(entity.substring(1));
                        }

stleary:
Can you provide some context? Perhaps a small XML doc that demonstrates the problem?

venkat-natsoft:
"<tag>Neutrophils.Hypersegmented &#X7C; Bld-Ser-Plas</tag>"
Full document url:
https://clinicaltrials.gov/ct2/show/NCT03874338?term=NCT02652975&rank=1&displayxml=true

stleary:
Is the problem that the XML-to-JSON parser does not properly handle  hex-encoded XML escape sequences?

johnjaylward:
@stleary , I believe that @venkat-natsoft is pointing out that our parser is case sensitive when it should not be.

https://github.com/stleary/JSON-java/blob/6ddaa13c1e2cd870de90d96dc139cf64ea3c583f/src/main/java/org/json/XMLTokener.java#L170

Current code:
```java
if (e.charAt(1) == 'x') {
```

Should be:
```java
if (e.charAt(1) == 'x' || e.charAt(1) == 'X') {
```

johnjaylward:
[...]

PR #596, opened 2021-03-30
Update Issue537.xml

IsmaelP19:
There was a Java error in line 74: "A decimal representation must immediately follow the “&#” in a character reference".
We have changed the letter X uppercase to lowercase and it seems to be solved.


johnjaylward:
I am not clear what issue this is fixes, can you explain?

johnjaylward:
Please ignore that error from Eclipse. The test file is there to explicitly test the uppercase `X` in the entity token as you can see if you reference issue #537 which generated that test file.

stleary:
Closing because the code seems to work as intended and there have been no further responses. If you disagree, drop a comment here and I will reopen.

PR #597, opened 2021-03-30
Update README.md

IsmaelP19:
There was a mistake in line 66. The command we must type to execute the test is "javac -cp .;json-java.jar Test.java" instead of "java -cp .;json-java.jar Test".

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ public class Test {
 
 *Execute the Test file*
 ```` 
-java -cp .;json-java.jar Test
+java -cp .;json-java.jar Test.java
 ````
 
 *Expected output*


stleary:
Are you sure the line change is correct? On a Windows command line, the existing line works, but your changed line does not. Can you clarify?

IsmaelP19:
I tried the same line code again and the existing line now it works for me. However, my changed line also works.

IsmaelP19:
> Are you sure the line change is correct? On a Windows command line, the existing line works, but your changed line does not. Can you clarify?

I tried the same line code again and the existing line now it works for me. However, my changed line also works. 
image

stleary:
Closing because the code seems to work as intended and there have been no further responses. If you disagree, drop a comment here and I will reopen.

PR #604, opened 2021-04-28
Fixed incorrect cast getting float from array

ianlovejoy:
Hello,

I noticed that if a double is added to a JSONArray and then retrieved with getFloat(), a class cast exception was raised. This case is handled correctly by JSONObject (as opposed to array).

The root cause turned out to be an incorrect cast. I corrected the cast and also added a test for this case. I tried to closely follow the existing coding conventions as well as the corresponding code in JSONObject.

Please let me know if you need any more information. Thanks for maintaining this project!

Sincerely,
Ian Lovejoy

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -326,7 +326,7 @@ public class JSONArray implements Iterable<Object> {
     public float getFloat(int index) throws JSONException {
         final Object object = this.get(index);
         if(object instanceof Number) {
-            return ((Float)object).floatValue();
+            return ((Number)object).floatValue();
         }
         try {
             return Float.parseFloat(object.toString());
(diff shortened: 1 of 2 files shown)


johnjaylward (approved):

ianlovejoy:
@johnjaylward thanks for approving these changes, it appears that there is a workflow awaiting approval as well since I am a first-time contributor, not sure if you can unstick that as well? Thanks!

stleary:
Exception when calling getFloat() for a valid value.

**Risks**
Low

**Changes to the API?**
No

No

No

No, but it looks like a unit test has not been added for this change. This should be corrected in a future commit.

No

**Review status**
**APPROVED**

stleary:

johnjaylward:
There is a new test @stleary

stleary:
Thanks @johnjaylward I saw the test but just didn't read it closely. Review updated.

#readthecode

PR #606, opened 2021-05-31
Fix javadoc

artem-smotrakov:
This pull request addresses several issues in Javadoc formatting that were reported on LGTM

https://lgtm.com/projects/g/stleary/JSON-java/alerts?mode=tree

This also suppresses the "Equals method does not inspect argument type" alert by adding a `SuppressWarning` annotation. This alert looks like a false positive. In fact, `object == this` implicitly check argument type -- if `object`'s and `this`'s types are different, the condition would be false.

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -128,6 +128,7 @@ public class JSONObject {
          *         null.
          */
         @Override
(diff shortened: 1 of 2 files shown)


stleary:
Fixes some warnings in case we decide to use LGTM

**Risks**
Minimal

**Changes to the API?**
No

No

No

No

No

**Review status**
**APPROVED**

stleary:
Starting 3 day comment window

PR #610, opened 2021-06-23
New JSONObject.optJSONObject method with defaultValue parameter

tilds:
I found myself often needing a version of **optJSONObject** where I could define my own default value as the 2nd argument (as you can on most of the other 'opt' methods).

The use case is when I need to access something that is nested deep in a json object, I currently have to check for null for every level of nesting, but if I could just set the default value to be and empty jsonobject instead of null, I could do something like this:
`String nestedvalue = jsonObject.optJSONObject("key", new JSONObject()).optJSONObject("key", new JSONObject()).optJSONObject("key", new JSONObject()).optString("nestedvalue")`

Let me know what you think :)

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -1370,9 +1370,21 @@ public class JSONObject {
      *            A key string.
      * @return A JSONObject which is the value.
      */
-    public JSONObject optJSONObject(String key) {
+    public JSONObject optJSONObject(String key) { return this.optJSONObject(key, null); }
+
+    /**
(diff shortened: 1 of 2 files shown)


johnjaylward (approved):
Looks good to me. Thanks for the contribution.

stleary:
One of the opt() functions does not have a default return value param

**Risks**
Low

**Changes to the API?**
Adding a new API method


No

No, but the JavaDocs will be slightly out of date


No, but some unit tests were repurposed. The originals will be restored later.

No

**Review status**
**APPROVED**
Starting 3 day comment window

Issue #611, opened 2021-07-12
Bugfix in JSONObject.Similar()


eyalb40:
I JSONObject.Similar() you return true\false after comparing a number property. Should only return if false.

Should replace 
                    return isNumberSimilar((Number)valueThis, (Number)valueOther);
with
                    if (!isNumberSimilar((Number)valueThis, (Number)valueOther))
                        return false;

stleary:
Agreed, good catch @eyalb40, and thanks for the heads-up. Feel free to submit a pull request to fix this, if you want.

stleary:
Fixed in #613

johnjaylward:
@stleary, you should also correct the `similar` in `JSONArray`

PR #613, opened 2021-07-18
Fixes Issue #611 JsonObject.similar() returns after number entry check

stleary:
#611 reported that similar returns too soon on success when a Number entry type is checked. Fixed by returning only if the Number check fails. Added unit test to exercise the new code.

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -1383,7 +1383,9 @@ public class JSONArray implements Iterable<Object> {
                     return false;
                 }
             } else if (valueThis instanceof Number && valueOther instanceof Number) {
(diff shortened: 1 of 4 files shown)


stleary:
Fixes #611 similar() returns too soon when checking Number entries

**Risks**
Low

**Changes to the API?**
No

No

No

No, but a unit test was added 

No

**Review status**
**APPROVED** - By myself

Starting 3 day comment window

johnjaylward (changes requested):
Please also correct the implementation in `JSONArray`

johnjaylward:
Also, due to the nature of the bug fix, this probably should get a new release.

stleary:
Forgot to check JsonArray, thanks for the heads-up @johnjaylward.
Agreed this is a good release candidate.

PR #617, opened 2021-07-23
Fixes #616 similar() problem comparing double vs BigDecimal

### LINKED ISSUE #616: Issue with JSONObject::similar
After upgrading from 20200518 to 20210307 some of our tests that are using JSONObject::similar started to fail.

This case reproduces the problem:

```
        JSONObject a = new JSONObject();
        a.put("a", 1.1);
        JSONObject b = new JSONObject("{\"a\":1.1}");
        System.out.println("a: " + a.toString(0));
        System.out.println("b: " + b.toString(0));
        System.out.println(a.similar(b));
```

This creates two JSONObjects whose value is just `{"a":1.1}`.

The output of this snippet on 20200518 is
```
a: {"a":1.1}
b: {"a":1.1}
true
```

but on version 20210307 the similar call suddenly returns false, even though the string representations of the objects are the same:
```
a: {"a":1.1}
b: {"a":1.1}
false
```

johnjaylward:
JSONObject isSimilar() may fail when comparing decimal values. Fixes #616 

**Risks**
Low

**Changes to the API?**
No change to the API. Internal behavior in the way that double values get created has changed, but in this case, it is considered worth the possible risk.

Yes

No

No, New unit tests were added to show the bug.

Several unrelated files in this commit contain minor documentation updates.

**Review status**
**APPROVED**

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -1159,6 +1159,18 @@ public class JSONObject {
      *          to convert.
      */
     static BigDecimal objectToBigDecimal(Object val, BigDecimal defaultValue) {
+        return objectToBigDecimal(val, defaultValue, true);
+    }
+    
+    /**
+     * @param val value to convert
+     * @param defaultValue default value to return is the conversion doesn't work or is null.
+     * @param exact When <code>true</code>, then {@link Double} and {@link Float} values will be converted exactly.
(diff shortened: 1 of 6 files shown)


stleary:
Thanks, @johnjaylward for the quick fix!  
I am going to merge this commit ahead of #613,  which changes one of the same files.

PR #623, opened 2021-10-04
Formatted source codes with coveo:fmt-maven-plugin.

javadev:


stleary:
@javadev Thanks for the PR, but the project is no longer accepting formatting-only code changes. This is mentioned in the Wiki FAQ, but it may be easy to miss, so I added a separate topic for it:

[Why don't you accept pull requests with extensive formatting changes?](https://github.com/stleary/JSON-java/wiki/FAQ#why-dont-you-accept-pull-requests-with-extensive-formatting-changes)

If you disagree strongly and think you can make a compelling case why these particular changes need to be included, feel free to post here.

PR #631, opened 2021-10-11
Refactor JSONPointerTest

sreekesh93:
Refactor the existing JSONPointer unit test with explicitly mentioning the expected values.


sreekesh93:
> ```java
> assertTrue(new JSONObject(EXPECTED_COMPLETE_DOCUMENT).similar(actualObject));
> ```

Thanks will accommodate this change

PR #634, opened 2021-10-14
Update README with Unix examples

JanitSri:
The command for **Compile a program that uses the jar** and **Execute the Test file** in the build instructions requires a colon instead of a semi-colon for CLASSPATH separator in Unix systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,8 @@ jar cf json-java.jar org/json/*.class
 
 *Compile a program that uses the jar (see example code below)*
 ````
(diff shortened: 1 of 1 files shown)


stleary:
Path setting did not have a UNIX version in the README

**Risks**
Low

**Changes to the API?**
N/A

No

This is the documentation

N/A

N/A

**Review status**
**APPROVED**
No comment window is needed, the PR has already been up for 2 days, and no code files were changed

PR #645, opened 2021-11-18
Detect and handle circular references when parsing Java beans

Zetmas:
Try to achieve the recursive check within one pass
Added some related test cases

### DIFF
--- a/src/test/java/org/json/junit/JSONObjectTest.java
+++ b/src/test/java/org/json/junit/JSONObjectTest.java
@@ -3221,22 +3221,59 @@ public class JSONObjectTest {
     }
     @Test(expected=JSONException.class)
     public void testSimpleRecursiveObject() {
+        // B -> A -> B ...
         RecursiveBean ObjA = new RecursiveBean("ObjA");
-        RecursiveBean ObjB = new RecursiveBean("ObjB", ObjA);
+        RecursiveBean ObjB = new RecursiveBean("ObjB");
+        ObjB.setRef(ObjA);
         ObjA.setRef(ObjB);
         new JSONObject(ObjA);
         fail("Expected an exception");
     }
     @Test(expected=JSONException.class)
     public void testLongRecursiveObject() {
+        // D -> C -> B -> A -> D ...
         RecursiveBean ObjA = new RecursiveBean("ObjA");
-        RecursiveBean ObjB = new RecursiveBean("ObjB", ObjA);
-        RecursiveBean ObjC = new RecursiveBean("ObjB", ObjB);
-        RecursiveBean ObjD = new RecursiveBean("ObjB", ObjC);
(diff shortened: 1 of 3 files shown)


src/test/java/org/json/junit/JSONObjectTest.java
```diff
@@ -3218,6 +3219,61 @@ public void testPutNullObject() {
...
+        RecursiveBean ObjC = new RecursiveBean("ObjC");
+        ObjC.setRef(ObjA);
+        ObjB.setRef(ObjA);
+        ObjB.setRef2(ObjA);
+        new JSONObject(ObjC);
+        new JSONObject(ObjB);
+        new JSONObject(ObjA);
+    }
+    @Test(expected=JSONException.class)
+    public void testRepeatObjectRecursive() {
```
johnjaylward, line 3261: This test should not throw an exception. There is no recursion here.

johnjaylward (changes requested):

johnjaylward:
This looks like a much better approach for implementation than the first go-around. Just needs a a few corrections to make that one test not throw an exception.

src/test/java/org/json/junit/JSONObjectTest.java
Zetmas: Thanks for your reply! Could you please explain a little more about why this test case is not recursive? I assumed this test case was recursive because it starts with C and ends with C in both children branches, creating loops of C-B-A-D-C... I also ran the test and it seems there would be an Exception thrown. Thanks for your patience!

johnjaylward: Ah, you are correct! My eyes glazed over the first C for some reason.

As additional tests, can you make these:

Self recursion, should throw exception:
```
A -> A
```

```
B -> A -> A
```


Complex graph, duplicate object(s), but no recursion (no exception thrown):
```
E -> B -> A -> D -> C
       -> D -> C
```

Zetmas: Additional test cases added

johnjaylward:
@stleary can you approve the workflow?

johnjaylward (approved):

stleary:
Circular references when parsing Java beans to JSON result in a StackOverflowException. This code change detects the cycle and throws a JSONException instead.

**Risks**
Low. 
Functional risk: Low. The implementation looks correct.
Performance risk: Moderate. All Java bean parsing will have to pay the overhead for cycle checking. This is considered an acceptable tradeoff for not having StackOverflowExceptions.

**Changes to the API?**
No. A new private API function is added, 

No

Probably not (TBD - check for references to handling of circular references)

No. 
New unit tests were added

The existing code was modified to implement the cycle checking.

**Review status**
**APPROVED**

stleary:
Starting 3 day comment window

PR #646, opened 2021-11-20
XMLParserConfiguration support for xml to json arrays

Zetmas:
- Added a member variable named `Set<String> forceList` to `XMLParserConfiguration` with corresponding constructor/getter/setter
- After configuring `forceList` in `XMLParserConfiguration`, `XML.toJSONObject` will now set the corresponding values as arrays, regardless of their number of occurrences
- If the returning value is originally `""`, it will be replaced by `[]`

### DIFF
--- a/src/main/java/org/json/XML.java
+++ b/src/main/java/org/json/XML.java
@@ -413,14 +413,26 @@ public class XML {
                         } else if (token == LT) {
                             // Nested element
                             if (parse(x, jsonObject, tagName, config)) {
-                                if (jsonObject.length() == 0) {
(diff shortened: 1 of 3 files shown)


stleary:
Enhancement for unconditional parsing of XML elements to array

**Risks**
Low

**Changes to the API?**
Added a new XML config option

No

Yes (Do we even have an XMLConfiguration topic?)

No

No

**Review status**
**APPROVED**

stleary:
Starting a 7 day comment window (won't be able to complete a code review rght away)

PR #655, opened 2021-12-14
Adding ordering of items

lvca:
Even if the JSON standard does not mandate the ordering of properties, in some cases it is requested by the specific use case. For example, the ordering of the items is requested by <a href="https://spec.graphql.org/June2018/#sec-Serialized-Map-Ordering">GraphQL specification</a>. The default implementation should use the HashMap because if the order is not requested, it is more efficient using the HashMap in terms of memory consumption and runtime cost for most operations.

ArcadeDB project depends heavily on JSON-java project and without this feature, we are forced to use an alternative library.

At the end of the day, there is no added code, just a different implementation of Map and **only** on explicit request.

WDYT?

Thanks in advance,
Luca Garulli
(Author of ArcadeDB and OrientDB Multi-Model DBMS)

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -40,6 +40,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.Locale;
@@ -185,6 +186,18 @@ public class JSONObject {
         // retrieval based on associative access.
         // Therefore, an implementation mustn't rely on the order of the item.
         this.map = new HashMap<String, Object>();
+    }
+
+    /**
+     * Construct an empty JSONObject specifying if the items are ordered based on the order of insertion.
+     */
+    public JSONObject(boolean ordered) {
+        // If ordered, a LinkedHashMap is used instead of a HashMap. Even if the JSON standard does not mandate the ordering of properties,
+        // in some cases it is requested by the specific use case. For example the ordering of the items is requested by
+        // <a href="https://spec.graphql.org/June2018/#sec-Serialized-Map-Ordering">GraphQL specification</a>. The default implementation
+        // should use the HashMap because if the order is not requested, it is more efficient using the HashMap in terms of memory
+        // consumption and runtime cost for most operations.
+        this.map = ordered ? new LinkedHashMap<String, Object>() : new HashMap<String, Object>();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


stleary:
Successful real-world projects like ArcadeDB carry a lot of weight, and I respect that.
But how can I accept this PR while adhering to the JSON specification? WDYT?

lvca:
Hi, @stleary thanks for evaluating this PR.

The change still adhere to the standard, it's not required to be ordered, but if they are it's not against it :-)

Or if you don't want to change the API, what about making JSONObject extensible? Like plugging in the Map implementation you like. In my case LinkedHashMap, but it could be anything else. What about a new constructor `JSONObject( Map<String,Object> map )` that accepts any Map implementation. WDYT?

johnjaylward:
> but if they are it's not against it :-)

The spec specifically says that JSON Object keys are "unordered"

lvca:
Unordered alphabetically, but not in the meaning of insertion order.

johnjaylward:
Looking over the PR itself, I'd say this implementation is incomplete. 99% of our requests for ordered keys in JSON objects stems from the string parser side (i.e. reading a JSON value stored in a DB or on disk and maintaining the original order), and this PR doesn't accomplish it.

Overall, I think the "optional" aspect of it is a good one, but I think this feature is misleading if we add it in as-is.

If you could make something similar that's optional all the way from the parser to the JSON Object, I think that would solve all the use cases of allowing ordered keys, but not having it be the default.

johnjaylward:
> Unordered alphabetically, but not in the meaning of insertion order.

The spec doesn't say anything about alphabetical. The spec says "unordered"

lvca:
I agree the spec is pretty vague, and it could be interpreted in either way. 

But on Wikipedia (for JSON), I found an interesting link to the paper from ECMA adds some more salt on it "The JSON syntax does not impose any restrictions on the strings used as names, does not require that name strings be unique, and does not assign any significance to the ordering of name/value pairs.": (source: https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf). 

So the original meaning, I guess, is more "don't expect them ordered in a particular way as we require with arrays", but if they are it's still ok.

WDYT?

lvca:
What other parsers are doing? Many parsers keep the insertion order for properties or give you the option. V8, probably the most used JS engine out there (behind Chrome and Node.js) keeps the insertion order.

johnjaylward:
I've read over all the specs multiple times, and a strictly conforming parser has no order to the Object keys:

https://datatracker.ietf.org/doc/html/rfc8259#section-1
> An object is an unordered collection of zero or more name/value
> pairs, where a name is a string and a value is a string, number,
> boolean, null, object, or array.

The ECMA docs are rewrites/interpretations of the RFC docs. They are not the baseline. Also, the original author of JSON and the RFCs was the author of this library, though he doesn't maintain it or give input anymore.

As I [stated above](https://github.com/stleary/JSON-java/pull/655#issuecomment-1000339760) though, I'm not against the ordering as long as it's not the default, and I think the solution should be more holistic, not just a boolean flag strapped onto a constructor in JSONObject.

lvca:
Ok, closing this PR and opening a totally different implementation: https://github.com/stleary/JSON-java/pull/658.

Issue #571, opened 2020-11-14
Preserve insertion order when stringifing JSONObject


holgerbrandl:
Sure, the specs don't specify an ordering (as correctly pointed out in the javadocs), but a particular implementation - such as this one - _can_  provide some ordering semantics. It is not wrong if an implementation is opinionated, there is just no definition in the json spec about this should be done.

So why not using LinkedHashMap instead of HashMap in https://github.com/stleary/JSON-java/blob/master/src/main/java/org/json/JSONObject.java#L178? This would improve (human) readability by preserving insertion order when printing json.

stleary:
@holgerbrandl Thanks for offering the suggestion. Keeping this issue open for a few days in case anyone else wants to post.

janicki:
Sounds like a good idea to me, but maybe implement it as a subclass that programmers can opt for when appropriate, e.g. `JSONOrderedObject`, to avoid unexpected compatibility or performance issues in existing apps?

johnjaylward:
I'm still not a fan of an implicit order internally. I'm not opposed to a subclass that makes it explicit though. The `JsonTokener` would need the most modifications in that case in order to know what subclass of JSONObject to instantiate.

If we were to bump the minimum required JVM from 6-8, this would be fairly easy to do by adding a lamba parameter to the tokener which returns a JSONObject.

```java
// Original JSONObject class
public class JSONObject {
// ...
   protected JSONObject(Map<String, object> backingMapStore) {
        this.map = backingMapStore;
   }

    public JSONObject(String source) throws JSONException {
        this(new JSONTokener(source, (JSONTokener t)->new JSONObject(t)));
    }

    // also extract the "tokener" code out into a new `protected final` method from the constructor
// ...
}

// new sub-class
public class JsonObjectInsertionOrder : JSONObject {

    public JsonObjectInsertionOrder() {
        super(new LinkedHashMap());
    }

    public JsonObjectInsertionOrder(String source) throws JSONException {
        super(new LinkedHashMap());
        this.populateFromTokener(new JSONTokener(source, (JSONTokener t)->new JsonObjectInsertionOrder(t)));
    }
}

// then would need to modify/add-to the tokener constructor(s) to take the new lambda
public class JSONTokener {
    /** constructor to use **/
    private Function<JSONTokener, JSONObject> jsonObjectConstructor;
    public JSONTokener(Reader reader, Function<JSONTokener,JSONObject> constructor) {
        // ... setup internal fields
[...]

PR #658, opened 2021-12-23
Supported using custom map implementation for JSONObject

lvca:
Supported using custom map implementation for JSONObject, like a LinkedHashMap to keep the insertion order to the map fields.

This is a more flexible implementation than https://github.com/stleary/JSON-java/pull/655.

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -164,7 +164,7 @@ public class JSONObject {
     /**
      * The map where the JSONObject's properties are kept.
      */
-    private final Map<String, Object> map;
+    protected final Map<String, Object> map;
 
     /**
      * It is sometimes more convenient and less ambiguous to have a
@@ -365,6 +365,17 @@ public class JSONObject {
     public JSONObject(Object bean) {
         this();
         this.populateMap(bean);
+    }
+
+    protected JSONObject(Class<? extends Map> mapImplementation) {
+        if( mapImplementation == null )
+            throw new JSONException("Map implementation is null");
+
+        try {
+            this.map = mapImplementation.getConstructor().newInstance();
+        } catch (Exception e) {
+            throw new JSONException("Error on instantiating default constructor of map implementation from class '"+mapImplementation+"'");
+        }
     }
 
     private JSONObject(Object bean, Set<Object> objectsRecord) {
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -208,16 +208,16 @@ public class JSONObject {
     }
 
     /**
-     * Construct a JSONObject from a JSONTokener.
-     *
-     * @param x
-     *            A JSONTokener object containing the source string.
-     * @throws JSONException
-     *             If there is a syntax error in the source string or a
-     *             duplicated key.
-     */
+         * Construct a JSONObject from a JSONTokener.
+         *
+         * @param x
+         *            A JSONTokener object containing the source string.
+         * @throws JSONException
+         *             If there is a syntax error in the source string or a
+         *             duplicated key.
+         */
     public JSONObject(JSONTokener x) throws JSONException {
-        this();
+        this(x.getMapImplementation());
         char c;
         String key;
 
@@ -369,12 +369,14 @@ public class JSONObject {
 
     protected JSONObject(Class<? extends Map> mapImplementation) {
         if( mapImplementation == null )
-            throw new JSONException("Map implementation is null");
-
-        try {
-            this.map = mapImplementation.getConstructor().newInstance();
-        } catch (Exception e) {
-            throw new JSONException("Error on instantiating default constructor of map implementation from class '"+mapImplementation+"'");
+            // Use the default implementation.
+            this.map = new HashMap<String,Object>();
+        else {
+            try {
+                this.map = mapImplementation.getConstructor().newInstance();
+            } catch (Exception e) {
+                throw new JSONException("Error on instantiating default constructor of map implementation from class '" + mapImplementation + "'");
+            }
         }
     }
 
--- a/src/main/java/org/json/JSONTokener.java
+++ b/src/main/java/org/json/JSONTokener.java
@@ -6,6 +6,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.StringReader;
+import java.util.Map;
 
 /*
 Copyright (c) 2002 JSON.org
@@ -56,6 +57,8 @@ public class JSONTokener {
     /** the number of characters read in the previous line. */
     private long characterPreviousLine;
 
+    /** custom map implementation to use. */
+    private Class<? extends Map> mapImplementation = null;
 
     /**
      * Construct a JSONTokener from a Reader. The caller must close the Reader.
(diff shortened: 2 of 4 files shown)


lvca:
The rationale behind using this approach was:
- I like to keep `map` property final for a lot of reasons, and replacing the internal map after the default instantiation of the HashMap was overkill
- I couldn't use `JSONObject(Map)` constructor because it's already used to copy a map into the object
- Having a protected constructor mandate extending the JSONObject class
- This allows also using hyper optimized map implementation (like Trove collections or Guava) if needed

johnjaylward:
I think this is a good start, but it still doesn't have a solution for JSONTokener to use either a custom map property itself, or the new optional extended class like you have in your test case.

Would you be willing to look into what changes would be needed to extend JSONTokener to optionally use the extended JSONObject (i.e. `OrderedJSONObject`) as part of some kind of configuration on the Tokener?

Basically this line would be the one needing to change based on how you set up the configuration: https://github.com/stleary/JSON-java/blob/master/src/main/java/org/json/JSONTokener.java#L431

stleary:
It might be better to get a consensus on the design before submitting a pull request. Issue #571 has been kept open and is available for this purpose.

lvca:
Updated the PR with the support for JSONTokener and wrote on https://github.com/stleary/JSON-java/issues/571.

src/main/java/org/json/JSONObject.java
```diff
@@ -208,16 +208,16 @@ public JSONObject(JSONObject jo, String ... names) {
...
+         *
+         * @param x
+         *            A JSONTokener object containing the source string.
+         * @throws JSONException
+         *             If there is a syntax error in the source string or a
+         *             duplicated key.
+         */
     public JSONObject(JSONTokener x) throws JSONException {
-        this();
+        this(x.getMapImplementation());
```
abimarank, line 220: Better do a null check before invoking on that instance

abimarank (changes requested):

src/main/java/org/json/JSONObject.java
johnjaylward: The tokener is expected to be non-null always there, the the called constructor checks for a null parameter already. This looks fine as-is, but possibly ensure the java doc calls out the expectation.

johnjaylward (commented):
Overall, this looks like a fairly simple change that matches my expectations. I will need more time to review more thoroughly to ensure we have consistent API across the classes due to the holidays.

@stleary can we leave this up for review through mid January?

stleary:
This code will remain in review for a minimum of 3 weeks before a decision is made.

lvca:
Any news on this? Just checking.

johnjaylward (approved):
I think this looks good.

aaron-613:
Are we finally getting ordered keys?  Nice!

atkawa7:
@stleary Thank you for the response. I really wanted to use this for a special case. I was trying build a custom jsonschema2pojo generator. The issue when I am coming across when using this lib to parse the json is order of generated fields. I am expecting the order of fields to consistent with the order of jsonschema properties. However this is not the case. The order changes frequently. The generated code is not ignored when pushed to git. So frequent code generations result in different code arrangements which result in large diffs. 

```
{
  "$id": "https://example.com/person.schema.json",
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "title": "Person",
  "type": "object",
  "properties": {
    "firstName": {
      "type": "string",
      "description": "The person's first name."
    },
    "lastName": {
      "type": "string",
      "description": "The person's last name."
    },
    "age": {
      "description": "Age in years which must be equal to or greater than zero.",
      "type": "integer",
      "minimum": 0
    }
  }
}
```

```java
public interface Person{
  Integer age();
  String firstName();
  String lastName();
}
```

johnjaylward:
@stleary , while I'm not saying this is the best implementation, it does provide a few benefits:

1. It's 100% backwards compatible with existing code seeing no change
2. We don't actually have an "ordered" implementation in this library.
3. It allows downstream users to create their own sub-classes that has their own order rules. i.e. some may want insertion order, others may want alphabetical.

The one thing I'm not in love of is the new public getter/setter on the JSONTokener (`getMapImplementation` and `setMapImplementation`). We really don't have an API similar to that used elsewhere in the library. I think I would prefer a constructor override there instead and make the value `final`.

What I'd also say we need to do with this PR if it's excepted is update the [Examples](https://github.com/stleary/JSON-java/blob/master/Examples.md) documentation to show a simple implementation.

zjns:
In my opinion, it's not best solution, `JSONObject(int initialCapacity)` and `JSONObject(Map<?, ?> m)` constructor still can't be customized.

stleary:
@zjns Any suggestions on how to improve this?

zjns:
@stleary I'd prefer some codes like this.

MapFactory
```java
public interface MapFactory {
    MapFactory DEFAULT = new MapFactory() {
        @Override
        public Map<String, Object> newMap(int initialCapacity) {
            if (initialCapacity == -1) {
                return new HashMap<>();
            }
            return new HashMap<>(initialCapacity);
        }
    };

    Map<String, Object> newMap(int initialCapacity);
}
```

In JSONObject
```java
private MapFactory mapFactory = MapFactory.DEFAULT;

public JSONObject() {
        this.map = mapFactory.newMap(-1);
}

public JSONObject(Map<?, ?> m) {
    if (m == null) {
        this.map = mapFactory.newMap(-1);
    } else {
        this.map = mapFactory.newMap(m.size());
        # other codes
    }
}

protected JSONObject(int initialCapacity){
    this.map = mapFactory.newMap(initialCapacity);
}

public void setMapFactory(MapFactory mapFactory){
    this.mapFactory = mapFactory;
}

public MapFactory getMapFactory(){
    return mapFactory;
}
```

stleary:
Does anyone have an opinion about whether something like @zjns' proposed implementation is better or worse than the original?

lvca:
Guys, there is zero commitment to the project and zero support for any effort coming from the community. I wrote this PR almost 6 months ago, and nobody knows what to do with it. I got your feedback, changed the code, and sent it back. Still nothing. Who is in charge? Please, this is painful...

stleary:
@lvca Sorry, my fault for letting this linger so long. The fact that this PR is still open means I haven't given up on it. 

An alternative map implementation goes against the design by the application author, and changes to preserve ordering have been rejected by him in the past.  This raises a high bar for a change to be accepted now.

I am willing to consider an update that meets these criteria:
- Does not have any impact on existing users
- Easy to use
- Generally useful for a range of different users
- Well designed, documented, written, and tested.

javadev:
It will be better to replace HashMap with LinkedHashMap.

stleary:
@javadev as a senior engineer, can you please explain your reasoning for recommending a LinkedHashMap, given that this is a reference application for parsing JSON in a way that follows the spec?

javadev:
> @javadev as a senior engineer, can you please explain your reasoning for recommending a LinkedHashMap, given that this is a reference application for parsing JSON in a way that follows the spec?

I know about spec. It may be optional, I agree.

stleary:
Please continue this discussion in #681. As soon as we have a PR that meets the requirements, it will be approved and merged. Apologies for taking so long to come to the decision to break with the direction set by the original author, who is also the person who wrote the JSON spec.

The requirements will be that the app must use the existing HashMap by default but can be configured or executed to run in a mode that uses a LinkedHashMap instead. The new mode should only use LinkedHashMap, and never use HashMap. I will come up with a suitable test suite for validation.

I am closing this PR because I don't think it complies. If you think differently, post here, and I will reopen PR and run the test suite against it (as soon as it's written). Otherwise, I think the approach outlined by @zjns may be a good place to start.

lvca:
Hi guys, it's been more than 1y... @stleary I don't understand why my solution wasn't good enough. Could you please elaborate on it? You said you'd prefer a better solution, but how?

Issue #653, opened 2021-12-09
optLong vs getLong inconsitencies


ejoncas-rivalbet:
Hi there,

We found a problem today where we stopped using `getLong(String)` and started using `optLong(String)` as it doesn't throw exception. 

The problem is that there are inconsistencies between the two methods in regards on how they treat "numbers as strings". I know this is not the intended case but our mobile-app was sending strings when we expected numbers and everything was working well with `getLong` and suddenly some strange scenarios broke. 

Here's is a sample scenario:

```
 @Test
    public void testJSONProblem() {
        JSONObject json = new JSONObject("{\"number_1\":\"01234\", \"number_2\": \"332211\"}");

        assertThat(json.getLong("number_1"), is(1234L));
        assertThat(json.optLong("number_1"), is(1234L)); //THIS FAILS it results 0
        assertThat(json.getLong("number_2"), is(332211L));
        assertThat(json.optLong("number_2"), is(332211L));
    }
```

As you can see optLong and getLong almost behave the same for numbers as strings but the problem arises when the number starts with `0`. 

The root of this problem is on how getLong parse numbers: 
https://github.com/stleary/JSON-java/blob/master/src/main/java/org/json/JSONObject.java#L807
 (it uses instanceof and `Number#longValue()`) where as `optLong` ends up calling this function 
https://github.com/stleary/JSON-java/blob/f1b0210b8acdbc5f5316bc5507e2f3630506a9a3/src/main/java/org/json/JSONObject.java#L2218

stleary:
Thanks, @ejoncas-rivalbet for bringing up this issue, investigating.

johnjaylward:
@ejoncas-rivalbet , just to be sure I understand you correctly, you believe that `optLong` is correct, while `getLong` is wrong, correct?

To be honest, I'm a little surprised that your first test passes. I would have expected the parser to process that as octal, which would be `668` in decimal.

johnjaylward:
hmm, I'm reading the javadocs on the parsers, and I'm not seeing a reference to the "leading 0 means octal" as noted in our comments. From what I can tell, all the string -> number parsers seem to process as decimal unless using the overload with the radix.

@stleary , if we don't have it already, we should include tests which process various number formats (hex, scientific/exponent, floating point hex, octal, standard decimal) and make sure we have coverage.

johnjaylward:
The JSON spec does not allow leading zeros on a number:
[...]

PR #661, opened 2022-02-02
Fixed opt/get Int/Long/Double inconsistencies

abimarank:
As discussed [here](https://github.com/stleary/JSON-java/issues/653), this PR fixed the inconsistencies between get/opt Numbers


johnjaylward (approved):
This looks good to me. I think removal of the hexfloat is probably OK. It is not a standard in JSON anyway.

stleary:
Possible problem with the unit tests?

abimarank:
@stleary The test failures aren't from these changes, they must be there earlier too. 

For example, consider the following test case

`    @Test
    public void verifyConstructor() {
        
        final JSONArray expected = new JSONArray("[10]");
        
        @SuppressWarnings("rawtypes")
        Collection myRawC = Collections.singleton("10");
        JSONArray jaRaw = new JSONArray(myRawC);

        Collection<Integer> myCInt = Collections.singleton(Integer.valueOf(10));
        JSONArray jaInt = new JSONArray(myCInt);

        Collection<Object> myCObj = Collections.singleton((Object) 10);
        JSONArray jaObj = new JSONArray(myCObj);

        System.out.println(jaRaw.toString());

        assertTrue(
                "The RAW Collection should give me the same as the Typed Collection",
                expected.similar(jaRaw));
        assertTrue(
                "The RAW Collection should give me the same as the Typed Collection",
                expected.similar(jaInt));
        assertTrue(
                "The RAW Collection should give me the same as the Typed Collection",
                expected.similar(jaObj));
    }`

The line `JSONArray expected = new JSONArray("[10]");` outputs `"["10"]"` not `"[10]"`. Hence the test case is failed. 

What should be the behaviour with numbers in String format when providing through the constructor?

johnjaylward:
Something must be incorrect with the changes in this PR. That test case is correct.

abimarank:
@johnjaylward Testcase is correct.

The line `JSONArray expected = new JSONArray("[10]");` outputs `"["10"]"` not `"[10]"`

It's like instead of producing an array of `int`, it produces an array of `String` numeric literals. Any hint why this behaviour?

abimarank:
I have fixed the issue with validating the String for numeric literals.

stleary:
Modifying stringToNumber() may not be the best way to address this issue. The goal is for the get* and opt* methods to behave the same, while minimizing changes to existing behavior.

PR #665, opened 2022-02-21
checking .length() is faster than .equals("")

unlucku:
Very minor change, when checking if a string has no characters or not then string.length() == 0 is always faster than string.equals("")
This only effects the stringToValue method

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -2295,7 +2295,7 @@ public class JSONObject {
     // Changes to this method must be copied to the corresponding method in
     // the XML class to keep full support for Android
     public static Object stringToValue(String string) {
-        if ("".equals(string)) {
+        if (string.length() == 0) {
             return string;
         }
 
--- a/src/main/java/org/json/XML.java
+++ b/src/main/java/org/json/XML.java
@@ -835,7 +835,7 @@ public class XML {
                             sb.append(toString(val, key, config));
                         }
                     }
-                } else if ("".equals(value)) {
+                } else if (value instanceof String && ((String) value).length() == 0) {
                     sb.append('<');
                     sb.append(key);
                     sb.append("/>");
--- a/src/main/java/org/json/XML.java
+++ b/src/main/java/org/json/XML.java
@@ -479,7 +479,7 @@ public class XML {
     // the one in JSONObject. Changes made here should be reflected there.
     // This method should not make calls out of the XML object.
     public static Object stringToValue(String string) {
-        if ("".equals(string)) {
+        if (string.length() == 0) {
             return string;
         }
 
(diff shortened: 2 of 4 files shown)


stleary:
Have you considered that the existing code provides protection in case the string being compared has somehow been set to null?

unlucku:
I did actually, I was going to add null checks originally till I remembered that instanceof returns false if the value checked is null. I checked the places stringToValue was called in commit 291e04485ca2ed35462dbf5d5ec70ac3577a8fba: 
For XML.java line 370 has a preceending check that throws an error if the value passed does not pass the instanceof check and in lines 417 and 420 dont need either check as it already has a instanceof and length check in the if blocks, so the value cannot be null when passed to stringToValue.
For JSONObject.java the function is only called by JSONTokener.java on line 473 which also already has a length check right before it (More on this at the end) and is returned from a trim function meaning in this case it cant be null either. We can actually remove the check preceding it if we rearrange the code a bit. I did not do this as I was unsure of the exact use of that exception.

In commit 71488150bafad6ee2b6e1449242e8127ecb0c2f9 I added the instanceof checks along with length != 0 (since .equals calls instanceof anyways), I omitted it in Cookie.java as the name is a result from trim.

Let me know if you have anymore questions!

unlucku:
Hope that I can get an opinion on this @stleary

stleary:
The criteria that I try to follow for code reviews is that the change should satisfy one or more of:

- A compelling improvement to the existing code.
- Fixes an egregious bug.
- Does not change existing behavior.

This change does not qualify, so it is a lower priority. Optimizing the code for speed is reasonable, so no problem there. But it seems to me this change makes the code less safe. Protecting code from NullPointerExceptions is a core principle, for me at least. Try finding a safer way to speed up the code.

src/main/java/org/json/JSONObject.java
```diff
@@ -2295,7 +2295,7 @@ protected static Number stringToNumber(final String val) throws NumberFormatExce
     // Changes to this method must be copied to the corresponding method in
     // the XML class to keep full support for Android
     public static Object stringToValue(String string) {
```
johnjaylward, line 2297: As this is a public method, the contract has to remain the same. This means supporting a `null` parameter value.

src/main/java/org/json/XML.java
```diff
@@ -479,7 +479,7 @@ public static Object stringToValue(String string, XMLXsiTypeConverter<?> typeCon
     // the one in JSONObject. Changes made here should be reflected there.
     // This method should not make calls out of the XML object.
     public static Object stringToValue(String string) {
-        if ("".equals(string)) {
```
johnjaylward, line 482: As this is a public method, the contract has to remain the same. This means supporting a `null` parameter value.

johnjaylward (changes requested):

stleary:
Closed due to lack of activity.

PR #668, opened 2022-02-22
Joe remote - added toStream to JSONObject

josepylee:
JSONObject.toStream() uses JAVA spliterator class, which breaks down each node in a give JSONObject and streams out.
There is an existing issue where broken down nodes may have duplicate keys especially for the leaf JSONObject nodes with its content key/values not in a form of JSONObject. In those cases, a new JSONObject gets created with the corresponding key/value pair, but this may introduce duplicate instances of JSONObject with certain keys.

for example:
                        "<book>" +
                            "<title>BBB</title>" +
                            "<author>BSmith</author>" +
                        "</book>" +

book node will have title,
and title node will also have title.


johnjaylward (changes requested):
This isn't really the best place to submit school work

PR #675, opened 2022-03-21
Updates value error messages to be consistent.

### LINKED ISSUE #649: Suggestion: more helpful wrong-type errors
Currently the exception just states that the value was the wrong type, for example:
```
org.json.JSONException: JSONObject["layers"] is not a JSONArray.
```

It would be helpful if the exception told us what the type actually was as well. Was it null? JSONObject? something else?

johnjaylward:
Fixes #649 

Updates value error messages to be consistent. Some errors would output value information, but others not.

Provide both the type and value that failed conversion. Tries not to
"toString" large value types like Arrays or Maps. For those types it
will just output the type and not a value.

**Risks**

Low

**Changes to the API?**

No

Yes

No

Yes. Some tests that verify error messages have been updated.

Private methods were removed that allowed generation of the error messages with no value information.

**Review status**
**APPROVED**

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -288,7 +288,7 @@ public class JSONArray implements Iterable<Object> {
                         .equalsIgnoreCase("true"))) {
             return true;
         }
-        throw wrongValueFormatException(index, "boolean", null);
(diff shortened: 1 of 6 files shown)


stleary:
Thanks @johnjaylward starting 3 day comment window.

PR #685, opened 2022-08-20
JSONObject map type unit tests

stleary:
Set up unit tests for user-managed JSONObject map type. Ensure that all methods result in a JSONObject map type that matches the one created in the default constructor.

**Risks**
Low

**Changes to the API?**
Added a getMapType() method to JSONObject

No

No

No

A few unit tests were adjusted to make it easier to test multiple JSONObject and JSONArray instances.

**Review status**
**APPROVED** - by myself


stleary:
Starting a 3 day comment window

Issue #657, opened 2021-12-21
Unit tests should check various number formats 


stleary:
Unit tests should check various number formats (hex, scientific/exponent, floating point hex, octal, standard decimal) and make sure we have coverage. Checks should be performed when parsing JSON documents and when calling functions like getLong() and optLong().

stleary:
Merged in #691

PR #691, opened 2022-10-04
create unit tests for various number formats

hendrixjoseph:
For issue #657 Unit tests should check various number formats

This adds tests for various number formats (namely: decimal with and without a decimal point, exponent with both capital and lowercase E, both in the positive and negative) for the various `get*` and `opt*` number methods of JSONObject.

It does NOT add hex, floating point hex, or octal, as these are not supported JSON features (see https://stackoverflow.com/a/52671839/6323312). JSONObject also does not support hex, floating point hex, or octal.

I can add these additional number format tests if wanted, though.

### DIFF
--- a/src/test/java/org/json/junit/JSONObjectNumberTest.java
+++ b/src/test/java/org/json/junit/JSONObjectNumberTest.java
@@ -1,0 +1,126 @@
+package org.json.junit;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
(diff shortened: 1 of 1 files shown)


johnjaylward (approved):

stleary:
Adds some missing unit tests.

**Risks**
Low

**Changes to the API?**
N/A

No

No

No

N/A

**Review status**
**APPROVED**

Starting 3 day comment window

stleary (approved):

PR #693, opened 2022-10-06
Fix flaky tests caused by JSON ordering

yannizhou05:
The tests below were found flaky because of the ordering of JSON objects. The orders of elements in the JSON objects are not the same every time being called. In this case, `JsonParser` from the package `com.google.gson` is applied to parse the JSON objects in order to ensure the orders are not changing. 

- `org.json.junit.JSONObjectTest.valueToString`
- `org.json.junit.JSONMLTest.testToJSONObject_reversibility`

It matters to fix these flaky tests as it may cause potential issues due to the uncertainty of orders from JSON objects rather than the program itself.

The test failures could be reproduced by

1. `mvn install -pl . -am -DskipTests`
2.  run tests with NonDex 
    `mvn -pl . edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=org.json.junit.JSONObjectTest#valueToString`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,6 +72,11 @@
             <version>1.9.5</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.8.6</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/pom.xml
+++ b/pom.xml
@@ -72,11 +72,6 @@
             <version>1.9.5</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>com.google.code.gson</groupId>
-            <artifactId>gson</artifactId>
-            <version>2.8.6</version>
-        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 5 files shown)


stleary:
@yannizhou05 Thanks for the PR. Can you please explain exactly what problem this code change is trying to solve?

yannizhou05:
> @yannizhou05 Thanks for the PR. Can you please explain exactly what problem this code change is trying to solve?

Sorry for the confusion. This code change is trying to apply JsonParser to fix the [flaky tests](https://docs.gitlab.com/ee/development/testing_guide/flaky_tests.html) mentioned above, because they sometimes fail (as the picture showed) and sometimes pass when comparing the JSON strings. The failure could be reproduced by the commands above by using the tool of [NonDex](https://github.com/TestingResearchIllinois/NonDex). The code change is to make sure the tests will always pass in this case.

stleary:
Is there a way to fix the tests without importing a new library?

yannizhou05:
@stleary I have updated the fixes without a new library, but in this case I have to change `HashMap` to `LinkedHashMap` in `JSONObject.java` instead of only changing the test codes. But the flakiness is also fixed by this way. Please let me know if there is any concern regarding this change.

stleary:
Only the test code should be updated.

yannizhou05:
Sorry that I don't have other ways, and thank you for your time.

PR #694, opened 2022-10-06
Pretty print XML

### LINKED ISSUE #593: how to make a pretty-printed XML text from JSONObject?
org.json.JSONObject#toString(int) has an indentFactor parameter, but org.json.XML.toString does not have an indentFactor parameter

DeaneOC:
This feature was requested some time ago on #593, there are now methods that allow users to pass an `indentFactor`to the `toString` methods in the XML class and this will return a Pretty Printer XML String with spaces based on `indentFactor`. 

This is done similarly to the JSONObject model.

Previous XML result with no `indentFactor` specified:
`<employee><name>sonoo</name><salary>56000</salary><married>true</married></employee>`

New result with `indentFactor` set as 2:
```
<employee>
  <name>sonoo</name>
  <salary>56000</salary>
  <married>true</married>
</employee>
```

Contributed on behalf of the [@opencastsoftware](https://github.com/opencastsoftware) open source team. :wave:


stleary:
@DeaneOC Nice work. Can you please make the following changes?

* New methods should have JavaDocs, especially for new parameters.
* Make sure new methods actually need to be public, or else use more restrictive access.
* Unit test very large strings should be stored in src/test/resources. You can search for 'issue537' for an example of how to do this.

DeaneOC:
> @DeaneOC Nice work. Can you please make the following changes?
> 
> * New methods should have JavaDocs, especially for new parameters.
> * Make sure new methods actually need to be public, or else use more restrictive access.
> * Unit test very large strings should be stored in src/test/resources. You can search for 'issue537' for an example of how to do this.

Thanks @stleary 

In terms of point one and two, I have JavaDoc comments on the three new methods that are public:
```
toString(object, indentFactor)
toString(object, tagLine, indentFactor)
toString(object, tagline, config, indentFactor)
```
You can currently call toString three different ways so I thought three new public methods that can also take an indent factor were appropriate.

The two methods below are private as users should not be able to call them directly, so I have not added JavaDocs, if you would like me to do so let me know:
```
toString(object, tagline, config, indentFactor, indent)
indent(indent)
```
I believe the current restrictions on method access are correct, three public and two private as mentioned above but if you believe otherwise let me know.

Point 3: I have altered test three to use the resource, I was aware it was very long so very happy to make that change. 

Thanks for getting back to me so quickly and let me know what you think about the above comments so I can make and required changes.

stleary:
@DeaneOC Thanks for the update. Yes, please add Javadocs to all new methods, not just the public ones.

DeaneOC:
> @DeaneOC Thanks for the update. Yes, please add Javadocs to all new methods, not just the public ones.

Javadocs added. Thanks @stleary

stleary:
Changes look good, but one of the new unit tests is failing when run from a Windows box:
`org.json.junit.XMLTest > testIndentComplicatedJsonObjectWithArrayAndWithConfig FAILED
    org.junit.ComparisonFailure at XMLTest.java:1235`

DeaneOC:
> Changes look good, but one of the new unit tests is failing when run from a Windows box: `org.json.junit.XMLTest > testIndentComplicatedJsonObjectWithArrayAndWithConfig FAILED org.junit.ComparisonFailure at XMLTest.java:1235`

Hi, it was to do with the System (mac vs windows) line separators being different. It has now been updated to work for both. Thanks @stleary

stleary:
Add support for XML pretty print 

**Risks**
Low

**Changes to the API?**
New public API methods

No

No

No. Unit tests were added.

No

**Review status**
**APPROVED**

Starting 3-day comment window

src/test/java/org/json/junit/XMLTest.java
```diff
@@ -1049,4 +1049,82 @@ public void testXSITypeMapNotModifiable() {
...
+                "          \"age\": 9.41,\n" +
+                "          \"angle\": 0.55\n" +
+                "        }\n" +
+                "      }\n" +
+                "    }\n" +
+                "  ]\n" +
+                "}" ;
+        JSONObject jsonObject = new JSONObject(str);
+        String xmlString = XML.toString(jsonObject, "Outer", 1);
+        System.out.println(xmlString);
```
javadev: Why not assert?

stleary: @javadev Can you clarify where you saw this in the code? I cannot find it.

javadev: You are right, the code is outdated. The current implementation has `assertEquals`.

stleary (approved):

Issue #652, opened 2021-11-27
Cycle detection doesn't catch cycles through collections


cushon:
The cycle detection introduced in #632 / #645 doesn't handle cycles that involved collections. The following example still results in a `StackOverflowError`:

```java
import java.util.Collections;
import java.util.List;
import org.json.JSONObject;

class Json {

  public static void main(String[] args) {
    new JSONObject(new Foo());
  }

  public static class Foo {

    public List<Foo> getFoos() {
      return Collections.singletonList(this);
    }
  }
}
```

```
Exception in thread "main" java.lang.StackOverflowError
	at org.json.JSONObject.getKeyNameFromMethod(JSONObject.java:1591)
	at org.json.JSONObject.populateMap(JSONObject.java:1548)
	at org.json.JSONObject.populateMap(JSONObject.java:1529)
	at org.json.JSONObject.<init>(JSONObject.java:366)
	at org.json.JSONObject.wrap(JSONObject.java:2499)
	at org.json.JSONObject.wrap(JSONObject.java:2457)
	at org.json.JSONArray.addAll(JSONArray.java:1609)
	at org.json.JSONArray.<init>(JSONArray.java:176)
	at org.json.JSONObject.wrap(JSONObject.java:2478)
	at org.json.JSONObject.populateMap(JSONObject.java:1563)
	at org.json.JSONObject.populateMap(JSONObject.java:1529)
	at org.json.JSONObject.<init>(JSONObject.java:366)
```

stleary:
To be fixed in #651

johnjaylward:
@stleary this is a different issue than the one fixed in #651

hendrixjoseph:
I have a fix for this (plus a unit test), as well as a couple other locations this might still be happening. I'll try to make a PR tomorrow.

Can you assign me please? Thank you.

stleary:
I don't think it will be possible to check for cycles in collections without impacting performance. If you disagree, post here to reopen this ticket.

Issue #632, opened 2021-10-13
Consider detecting cycles when constructing JSONObjects


cushon:
`JSONObject.populateMap` mentions that "the bean can not be recursive", and attempting to pass a recursive bean to `JSONObject` results in a `StackOverflowError`. The `StackOverflowError` can be hard to debug, because the stack trace is usually truncated before the frames that show where `new JSONObject(bean)` was called.

It might be worth detecting cycles when constructing `JSONObjects` to report a better error message for beans containing cycles.

```java
import org.json.JSONObject;

class Json {

  public static void main(String[] args) {
    new JSONObject(new Foo());
  }

  public static class Foo {

    public Bar getBar() {
      return new Bar(this);
    }
  }

  public static class Bar {

    private final Foo foo;

    Bar(Foo foo) {
      this.foo = foo;
    }

    public Foo getFoo() {
      return foo;
    }
  }
}
```

```
Exception in thread "main" java.lang.StackOverflowError
	at java.base/java.lang.reflect.Method.getRoot(Method.java:68)
	at java.base/java.lang.reflect.Executable.declaredAnnotations(Executable.java:600)
	at java.base/java.lang.reflect.Executable.getAnnotation(Executable.java:572)
	at java.base/java.lang.reflect.Method.getAnnotation(Method.java:695)
	at java.base/java.lang.reflect.AnnotatedElement.isAnnotationPresent(AnnotatedElement.java:274)
	at java.base/java.lang.reflect.AccessibleObject.isAnnotationPresent(AccessibleObject.java:525)
	at org.json.JSONObject.getAnnotationDepth(JSONObject.java:1674)
	at org.json.JSONObject.getKeyNameFromMethod(JSONObject.java:1568)
	at org.json.JSONObject.populateMap(JSONObject.java:1538)
	at org.json.JSONObject.<init>(JSONObject.java:365)
	at org.json.JSONObject.wrap(JSONObject.java:2475)
	at org.json.JSONObject.populateMap(JSONObject.java:1543)
	at org.json.JSONObject.<init>(JSONObject.java:365)
	at org.json.JSONObject.wrap(JSONObject.java:2475)
	at org.json.JSONObject.populateMap(JSONObject.java:1543)
	at org.json.JSONObject.<init>(JSONObject.java:365)
	at org.json.JSONObject.wrap(JSONObject.java:2475)
	at org.json.JSONObject.populateMap(JSONObject.java:1543)
[...]

PR #699, opened 2022-10-20
Issue 652 cycle detection doesnt catch cycles through collections

hendrixjoseph:
For issue #652:

This PR detects cycles through collections. It also adds a unit test to verify that this cycle is detected.

The PR also adds three additional unit tests - one of which passes, two of which fail. The passing one verifies that cycles are detected for simple Objects - this was competed with issue #632 / PR #645.

The two failing ones attempt to verify that a cycle is detected in arrays and maps. This is currently not implemented, and a felt it was outside the scope of this issue and PR. If the tests need to pass before the code is merged, I can add an `@Ignore` annotation to the two failing tests pending another issue & PR.

### DIFF
--- a/src/test/java/org/json/junit/JSONObjectRecursiveExceptionTest.java
+++ b/src/test/java/org/json/junit/JSONObjectRecursiveExceptionTest.java
@@ -9,6 +9,7 @@ import org.hamcrest.core.StringContains;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
@@ -49,18 +50,20 @@ public class JSONObjectRecursiveExceptionTest {
         exceptionRule.expectMessage(StringContains.containsString("JavaBean object contains recursively defined member variable of key"));
     }
    
-    @Test//(expected = JSONException.class)
-    public void testRecursiveList() {
-          
+    
+    @Test
+    public void testRecursiveList() {          
         new JSONObject(new RecursiveList());
     }
     
-    @Test//(expected = JSONException.class)
+    @Ignore
+    @Test
     public void testRecursiveArray() {        
         new JSONObject(new RecursiveArray());
     }
     
-    @Test//(expected = JSONException.class)
+    @Ignore
+    @Test
     public void testRecursiveMap() {        
         new JSONObject(new RecursiveMap());
     }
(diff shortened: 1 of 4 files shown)


hendrixjoseph:
The [Java 8 test build](https://github.com/stleary/JSON-java/actions/runs/3291548305/jobs/5425848970) did fail due to those two intentionally failing tests; let me know if you want me to add `@Ignore` to them.

stleary:
@hendrixjoseph Sorry for not responding sooner. Failing unit tests are a concern. 
What is the performance cost of this change (how much does it slow things down)?

src/test/java/org/json/junit/JSONObjectRecursiveExceptionTest.java
```diff
@@ -0,0 +1,75 @@
...
+        exceptionRule.expectMessage(StringContains.containsString("JavaBean object contains recursively defined member variable of key"));
+    }
+   
+    
+    @Test
+    public void testRecursiveList() {          
+        new JSONObject(new RecursiveList());
+    }
+    
+    @Ignore
```
javadev, line 59: Why?

hendrixjoseph: Why the `new JSONObject(new RecursiveList());` or why the `@Ignore`?

The `new JSONObject(new RecursiveList());` is testing if it throws the right exception if the list contains the object which has the list.

The `@Ignore` (and there's two of them) are for two additional locations where I identified stack overflows when having the object-within-something - one case is within a Map, one case is within an array. Since I didn't fix them as part of this PR (I felt it was out of scope), I added the `@Ignore` annotation so I could show where the stack overflows were happening, but not have the test fail.

hendrixjoseph:
> @hendrixjoseph Sorry for not responding sooner. Failing unit tests are a concern. What is the performance cost of this change (how much does it slow things down)?

I added an `@Ignore` to the two failing tests - see the [discussion above](https://github.com/stleary/JSON-java/pull/699#pullrequestreview-1159194775) with @javadev.

As far as performance costs - I didn't measure anything, but it does do a linear search through each list to see if there's an object that will be processes recursively. So it slows it down *O(n)* for each list (where *n* is the size of the list). If there are no lists, then there are no performance costs.

src/test/java/org/json/junit/JSONObjectRecursiveExceptionTest.java
javadev: Why do we need code that will be ignored? It may be an exception check or just remove it.

PR #703, opened 2022-10-28
Update Releases.md for JSONObject(Map): Throws NPE if key is null

### LINKED ISSUE #678: > JSONObject(Map): Throws NPE if key is null
> JSONObject(Map): Throws NPE if key is null

A bit late here, but I would consider throwing an exception for an input that used to be accepted a breaking API change. In my company's codebase we had some tests that verified this behavior that were broken as a result, and a currently unknown amount of code that could actually be depending on this behavior.

Obviously this ship has already sailed (and rejecting null keys makes much more sense), but can we at least point out the breaking change in the release history documentation? Would be very useful as a heads-up for anyone trying to upgrade from an old version.

_Originally posted by @data-enabler in https://github.com/stleary/JSON-java/issues/405#issuecomment-1113714068_

TomerPacific:
Fixes #678 

Reflecting breaking change that happened in version [20180813](https://github.com/stleary/JSON-java/releases/tag/20180813).

### DIFF
--- a/docs/RELEASES.md
+++ b/docs/RELEASES.md
@@ -20,6 +20,7 @@ and artifactId "json". For example:
 20190722    Recent commits
 
 20180813    POM change to include Automatic-Module-Name (#431)
+            JSONObject(Map) now throws an exception if any of a map keys are null (#405)
 
 20180130    Recent commits
 


stleary:
Documentation added

**Risks**
Low

**Changes to the API?**
N/A

No

No

N/A

N/A

**Review status**
**APPROVED**

Starting 3 day comment window

stleary (approved):

PR #712, opened 2022-12-01
add Serializable and Cloneable support via JSON string

nrhope:
In many legacy applications JSONObjects endup being part of larger objects that need to be deep-cloned via Serialization or because they are part of Value Objects included in UI Form bean etc. Hence is is very helpful for this class to support such functionality using the very handy and already present .toString() representation as a JSON string. If it isn't supported by the object itself, then an in-built wrapper utility object would still be better than users having to craft their own wrappers.
If this looks like something you'd be willing to accept into the library then I'd be happy to extend to to any other objects like JSONArray and/or add test cases as per your recommendations. 
Let me know what you think, thanks for a great library!

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -6,6 +6,8 @@ Public Domain.
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.io.ObjectStreamException;
+import java.io.Serializable;
 import java.io.StringWriter;
 import java.io.Writer;
 import java.lang.annotation.Annotation;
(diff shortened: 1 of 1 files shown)


stleary:
@nrhope Thanks, but I think this is outside the scope of the application. However, I will leave this PR up for a few days in case anyone wants to comment on it.

nrhope:
No problems.

src/main/java/org/json/JSONObject.java
```diff
@@ -81,7 +83,14 @@
  * @author JSON.org
  * @version 2016-08-15
  */
-public class JSONObject {
+public class JSONObject implements Serializable, Cloneable {
```
javadev, line 86: It may be implemented with custom class.

stleary:
Closing due to lack of activity, and prior decision.

PR #715, opened 2023-01-03
Add Naming Strategies at class level

saulgiordani:
I'm using the library for an application and i'm missing the possibility to change all the class' properties to a particular format, eg snake case. This development introduces it through the annotation @JSONNamingStrategy.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,8 +7,8 @@ apply plugin: 'eclipse'
 apply plugin: 'maven-publish'
 
 //plugins {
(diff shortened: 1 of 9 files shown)


stleary:
Does not compile with Java 1.6. Also, it is

stleary:
Thanks for the pull request, but not accepting it because I don't think it aligns with the project goal of being a reference application. If you need this functionality, please consider cloning the repo and making the change locally.

PR #723, opened 2023-02-10
JSONML should be protected from stack overflow exceptions caused by recursion, resolving #722

### LINKED ISSUE #722: JSONObject/Array should be protected from stack overflow exceptions caused by recursion
See #720 for an example of this might be done.

TamasPergerDWP:
The fix follows the method used in fixing the same issue for XML in the previous PR: 
Limit the XML nesting depth for CVE-2022-45688 #720

Additionally, fixing a minor issue with XMLParserConfiguration.clone() and amending some Javadoc.

Resolves #722

### DIFF
--- a/src/main/java/org/json/JSONMLParserConfiguration.java
+++ b/src/main/java/org/json/JSONMLParserConfiguration.java
@@ -7,7 +7,7 @@ Public Domain.
  * Configuration object for the XML to JSONML parser. The configuration is immutable.
  */
 @SuppressWarnings({""})
-public class XMLtoJSONMLParserConfiguration {
+public class JSONMLParserConfiguration {
     /**
      * Used to indicate there's no defined limit to the maximum nesting depth when parsing a XML
      * document to JSONML.
@@ -20,11 +20,11 @@ public class XMLtoJSONMLParserConfiguration {
     public static final int DEFAULT_MAXIMUM_NESTING_DEPTH = 512;
 
     /** Original Configuration of the XML to JSONML Parser. */
-    public static final XMLtoJSONMLParserConfiguration ORIGINAL
-        = new XMLtoJSONMLParserConfiguration();
+    public static final JSONMLParserConfiguration ORIGINAL
+        = new JSONMLParserConfiguration();
     /** Original configuration of the XML to JSONML Parser except that values are kept as strings. */
-    public static final XMLtoJSONMLParserConfiguration KEEP_STRINGS
-        = new XMLtoJSONMLParserConfiguration().withKeepStrings(true);
+    public static final JSONMLParserConfiguration KEEP_STRINGS
+        = new JSONMLParserConfiguration().withKeepStrings(true);
 
     /**
      * When parsing the XML into JSONML, specifies if values should be kept as strings (<code>true</code>), or if
@@ -40,7 +40,7 @@ public class XMLtoJSONMLParserConfiguration {
     /**
      * Default parser configuration. Does not keep strings (tries to implicitly convert values).
      */
-    public XMLtoJSONMLParserConfiguration() {
+    public JSONMLParserConfiguration() {
         this.keepStrings = false;
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/json/XMLtoJSONMLParserConfiguration.java
```diff
@@ -0,0 +1,128 @@
+package org.json;
+/*
+Public Domain.
+*/
+
+/**
+ * Configuration object for the XML to JSONML parser. The configuration is immutable.
+ */
+@SuppressWarnings({""})
+public class XMLtoJSONMLParserConfiguration {
```
stleary, line 10: Is there a reason for not re-using XMLParserConfiguration?
Either way, please don't expose any new APIs for JSONML besides the max nesting depth.

TamasPergerDWP: The main reason: not all options of XMLParserConfiguration is supported with JSONML parsing, so it seemed cleaner to create a new config class.

TamasPergerDWP: The new APIs for JSONML:
- public static JSONObject toJSONObject(String string, XMLtoJSONMLParserConfiguration config)
- public static JSONObject toJSONObject(XMLTokener x, XMLtoJSONMLParserConfiguration config)

These are both necessary to support max nesting depth and the `keepStrings` option in one function.
The alternative would be to add create 4 new `toJSONObject` functions that include the `int maxNestingDepth` parameter, on top of the existing 4 such functions - the potential parameter combinations doubling with a new optional parameter.

Is this acceptable?

TamasPergerDWP: Just noticed that I did not add the max depth limit override option for the `toJSONArray` calls - await a fix soon.

TamasPergerDWP:
@stleary I have a fix for the `JSONArray` and `JSONObject` as well - by adding the depth limit to the `JSONTokener` class. (See my internal PR: https://github.com/TamasPergerDWP/JSON-java/pull/1)
Three questions:
1. Shall I add the fix to the current PR, or open a new one?
2. `JSONArray` and `JSONObject` are used internally in the `CDL`, `Cookie` and `CookieList` classes. I introduced the `maxDepthLimit` optional parameter into those classes, too. Is that OK? Do those classes need additional tests for the maxDepthLimit behaviour?
3. The `JSONTokener` class is the parent of `XMLTokener` - after introducing the max nesting depth feature in `JSONTokener`, we may simplify the solution that is in place for the XML parsing (unifying both approaches). Is this something we should look into?

stleary:
Let's work through the JSONML changes before starting JSONObject and JSONArray.
Please rename `XMLtoJSONMLParserConfiguration` to something simpler, like `JSONMLParserConfiguration`.
The new API methods for JSONML look OK. Code also looks OK, but still needs an in-depth review, working on it.

TamasPergerDWP:
> Let's work through the JSONML changes before starting JSONObject and JSONArray. 

Fair enough, I will leave the other changes for another day.

> Please rename `XMLtoJSONMLParserConfiguration` to something simpler, like `JSONMLParserConfiguration`. 

Done, ready for review.

src/main/java/org/json/JSONMLParserConfiguration.java
```diff
@@ -0,0 +1,128 @@
...
+        this.keepStrings = false;
+    }
+
+    /**
+     * Configure the parser string processing and use the default CDATA Tag Name as "content".
+     * @param keepStrings <code>true</code> to parse all values as string.
+     *      <code>false</code> to try and convert XML string values into a JSON value.
+     * @param maxNestingDepth <code>int</code> to limit the nesting depth
+     */
+    public JSONMLParserConfiguration(final boolean keepStrings, final int maxNestingDepth) {
```
stleary, line 53: This method should be private

stleary:
The code looks good, just one comment

src/main/java/org/json/JSONMLParserConfiguration.java
TamasPergerDWP: @stleary Amended as requested.

TamasPergerDWP:
> @TamasPergerDWP The code looks good, just one comment

@stleary Code amended as requested

stleary:
JSONML should have the same StackOverflow protection as XML.

**Risks**
Low

**Changes to the API?**
Some new API methods were added for those who don't want to use the default nesting depth.

Yes

No

No

Future refactoring will unite the parserconfig classes in a class hierarchy of some sort.

**Review status**
**APPROVED**

stleary:

stleary (approved):

PR #726, opened 2023-02-12
avoid using recursion in for JSONML and JSONTokener

### LINKED ISSUE #722: JSONObject/Array should be protected from stack overflow exceptions caused by recursion

cleydyr:
Fixes #722

### DIFF
--- a/src/test/java/org/json/junit/JSONArrayTest.java
+++ b/src/test/java/org/json/junit/JSONArrayTest.java
@@ -1323,20 +1323,6 @@ public class JSONArrayTest {
         Util.checkJSONArrayMaps(jsonArray);
 	}
 
-    /**
-    * Tests for stack overflow. See https://github.com/stleary/JSON-java/issues/654
-    */
-    @Test(expected = JSONException.class)
-    public void issue654StackOverflowInputWellFormed() {
-        //String input = new String(java.util.Base64.getDecoder().decode(base64Bytes));
-        final InputStream resourceAsStream = JSONObjectTest.class.getClassLoader().getResourceAsStream("Issue654WellFormedArray.json");
-        JSONTokener tokener = new JSONTokener(resourceAsStream);
-        JSONArray json_input = new JSONArray(tokener);
-        assertNotNull(json_input);
-        fail("Excepected Exception.");
-        Util.checkJSONArrayMaps(json_input);
-    }
-
     @Test
     public void testIssue682SimilarityOfJSONString() {
         JSONArray ja1 = new JSONArray()
--- a/src/test/java/org/json/junit/JSONObjectTest.java
+++ b/src/test/java/org/json/junit/JSONObjectTest.java
@@ -3476,19 +3476,6 @@ public class JSONObjectTest {
         fail("Excepected Exception.");
     }
     
-    /**
-    * Tests for stack overflow. See https://github.com/stleary/JSON-java/issues/654
-    */
-    @Test(expected = JSONException.class)
-    public void issue654StackOverflowInputWellFormed() {
-        //String input = new String(java.util.Base64.getDecoder().decode(base64Bytes));
-        final InputStream resourceAsStream = JSONObjectTest.class.getClassLoader().getResourceAsStream("Issue654WellFormedObject.json");
-        JSONTokener tokener = new JSONTokener(resourceAsStream);
-        JSONObject json_input = new JSONObject(tokener);
-        assertNotNull(json_input);
-        fail("Excepected Exception.");
-    }
-
     @Test
     public void testIssue682SimilarityOfJSONString() {
         JSONObject jo1 = new JSONObject()
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -85,47 +85,8 @@ public class JSONArray implements Iterable<Object> {
      */
     public JSONArray(JSONTokener x) throws JSONException {
         this();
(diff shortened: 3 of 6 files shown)


stleary:
Nice work!
A description of what was changed and why it was changed would be helpful. 
The changes to JSONML should probably not be included in this pull request, since that file is being updated in #723 and will be accepted before this PR. 
Good to see the unit tests did not find any regressions. 
What is needed to ensure that there have been no functional changes - i.e. that the new code produces exactly the same results as the old? I don't think the current unit tests are sufficient to confirm this.

cleydyr:
No problem. I can drop the first commit, which is related to JSONML.

The commits prefixed by `refactoring` can be proven to make no functional changes.

The last commit is trickier, as I tried to simulate what the call stack does: push local variables (at least the relevant ones) and the instruction into the stack, but it would be hard to prove that the two algorithms yield the same results.

So, I see two alternatives: 1) generate a considerable amount of data in the JSON format that corresponds to the grammar that this parser obeys (the grammar should be defined formally, as in https://www.json.org/json-en.html) and add tests to _reduce_ the likelihood of missing a case (never eliminating it, of course) or 2) assume the tests we have currently are "good enough" and, should anybody find a regression, fix the parser accordingly.

What do you think?

stleary:
A lot of code was updated. The unit tests are good for catching regressions but for this PR I don't think that is enough. Probably needs a careful review and some additional unit testing. Otherwise, I think someone could fairly ask why not create a ParserConfiguration class similar to XML and JSONML and avoid all the refactoring.

src/test/java/org/json/junit/JSONArrayTest.java
```diff
@@ -1323,20 +1323,6 @@ public void jsonArrayClearMethodTest() {
...
-    * Tests for stack overflow. See https://github.com/stleary/JSON-java/issues/654
-    */
-    @Test(expected = JSONException.class)
-    public void issue654StackOverflowInputWellFormed() {
-        //String input = new String(java.util.Base64.getDecoder().decode(base64Bytes));
-        final InputStream resourceAsStream = JSONObjectTest.class.getClassLoader().getResourceAsStream("Issue654WellFormedArray.json");
-        JSONTokener tokener = new JSONTokener(resourceAsStream);
-        JSONArray json_input = new JSONArray(tokener);
-        assertNotNull(json_input);
-        fail("Excepected Exception.");
```
javadev, line 1336: Why test was removed?

src/test/java/org/json/junit/JSONObjectTest.java
```diff
@@ -3476,19 +3476,6 @@ public void issue654IncorrectNestingNoKey2() {
...
-    * Tests for stack overflow. See https://github.com/stleary/JSON-java/issues/654
-    */
-    @Test(expected = JSONException.class)
-    public void issue654StackOverflowInputWellFormed() {
-        //String input = new String(java.util.Base64.getDecoder().decode(base64Bytes));
-        final InputStream resourceAsStream = JSONObjectTest.class.getClassLoader().getResourceAsStream("Issue654WellFormedObject.json");
-        JSONTokener tokener = new JSONTokener(resourceAsStream);
-        JSONObject json_input = new JSONObject(tokener);
-        assertNotNull(json_input);
-        fail("Excepected Exception.");
```
javadev, line 3489: Test should not be removed

javadev:
I see so many changes and few unit tests

javadev:
This branch has conflicts that must be resolved

src/test/java/org/json/junit/JSONArrayTest.java
cleydyr: Because with my changes a stack overflow is not expected to happen as the algorithm to parse strings to JSON is not recursive anymore. So the test would fail 100% of the time, instead of intermittently as it does currently.

src/test/java/org/json/junit/JSONObjectTest.java
cleydyr: Why?

src/test/java/org/json/junit/JSONArrayTest.java
javadev: It may be changes in test

src/test/java/org/json/junit/JSONObjectTest.java
javadev: It is not good remove tests after code changes

src/test/java/org/json/junit/JSONArrayTest.java
JosephWitthuhnTR: For both of these, these tests looks like they were **testing to make sure that a stack overflow occurred**. So these tests "pass" if the bug isn't fixed, but will fail if the bug is fixed. So... not the best test to leave in.

That said, a better test might be to ensure that the use case that previously threw an exception no longer throws an exception? So instead of removing the tests entirely, update them to test that use case, but ensure that no exception is thrown instead of ensuring that an exception is thrown?

(I'm not a maintainer on this project, just my two cents as someone who was watching this CVE.)

melloware (approved):

javadev (changes requested):
There are many code changes and few unit tests

javadev:
This branch has conflicts that must be resolved

javadev:
This branch has conflicts that must be resolved

cleydyr:
I'll send this again when I rebase and have some test coverage.

PR #730, opened 2023-03-01
Added object methods for optBooleanObj (returns Boolean vice boolean)…

davejbur:
…, optFloatObj (returns Float vice float), optIntegerObj (returns Integer vice integer), optLongObj (returns Long vice long); changes to pom/nb-configuration  can be ignored.

### DIFF
--- a/nb-configuration.xml
+++ b/nb-configuration.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project-shared-configuration>
+    <!--
+This file contains additional configuration written by modules in the NetBeans IDE.
+The configuration is intended to be shared among all the users of project and
+therefore it is assumed to be part of version control checkout.
+Without this configuration present, some functionality in the IDE may be limited or fail altogether.
+-->
+    <properties xmlns="http://www.netbeans.org/ns/maven-properties-data/1">
+        <!--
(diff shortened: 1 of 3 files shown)


stleary:
PR should include a description of what problem this code solves.
pom.xml and nb-configuration.xml should be removed.
JSONObject should only contain the changes you want to add.

davejbur:
Thanks - this is me trying to work out how to make github do what I want it to, clearly I've got something wrong somewhere. I forked the original repo, clones it to my local repo, made the changes (including having to disable the gpg plugin to get it to compile), committed it locally, pushed it back to my github fork, then tried to do a pull request.

However, there seemed to be no way of selecting just the one file - all I could do was PR all the changes as one wrapped up mod. I imagine there must be an option somewhere like "only make PR for this file" but I couldn't find it...

Nor can I work out why it looks like the whole file is getting reformatted - nothing I've consciously done!

Anyway, the reason for the changes was to add in four methods to return objects instead of primitives. The reason for doing so is that, sometimes, I want (e.g.) optInt to return null if a value isn't found (i.e. I want to do ```optInt(theKey,null)```). However, that gives a compiler error ```<null> cannot be converted to int```, since optInt returns an ```int```. Hence adding an almost identical function that returns an ```Integer``` instead.

Anyway, if you can let me know how only to submit the changes and not the pom.xml etc, I'll add that explanation into the PR. Thanks!

Issue #718, opened 2023-01-28
JSONTokener should implement java.io.Closeable


jtotht:
While `JSONTokener(java.io.Reader)` and `JSONTokener(java.io.InputStream)` constructors have explicit notes about JSONTokener not closing the reader/stream, I’d like to have it close the reader/stream. Implementing `java.io.Closeable` isn’t backward-incompatible (if the library user doesn’t explicitly close it, nothing happens), and for those who use try-with-resources construct, having to keep a separate reference for the reader/stream is a pain. Although try-with-resources itself is Java 7+, the [`java.io.Closeable` interface](https://docs.oracle.com/javase/6/docs/api/java/io/Closeable.html) is `@since 1.5`, so implementing the interface can be done without losing Java 6 compatibility.

tsufiyan:
Shall I work on this? If yes, please assign

haribabu-dev:
Its completed & reviewed at commit https://github.com/stleary/JSON-java/commit/e1eabc9c27f954ce8fe8032f12f92f51c0e7c9eb

jtotht:
And reverted before getting merged. 🙁

stleary:
Closed due to lack of activity. Please post here if you think it should be reopened.

jtotht:
Should I spam you with “what’s going on?” messages every few weeks? Doing so would just waste time of both of us. Lack of activity is not the same as lack of interest or usefulness. Please don’t close the issue until it gets implemented and released; especially not as “completed”, which is simply not true.

stleary:
@jtotht Thanks for the feedback. Historically, issues have been left open on this repo for an extended period, which has led to its own set of problems. I am making an effort to better manage the project by closing issues that are unlikely to be acted upon in the near term. The label "completed" that GitHub applies when an issue is closed with a comment is just GitHub behavior and does not necessarily mean that the requested feature or fix has been implemented. Given that the project is now primarily in maintenance mode, I am cautious about making major changes to the API, such as making `JSONTokener` implement `Closeable`. A change of this magnitude requires a compelling use case, given its potential impact on backward compatibility and API complexity. Please feel free to make that case if you wish.

jtotht:
> I am making an effort to better manage the project by closing issues that are unlikely to be acted upon in the near term.

[...]

PR #733, opened 2023-03-04
JSONTokener implemented java.io.Closeable

haribabu-dev:


stleary:
Why should JSONTokener implement Closeable? It is not a closeable resource. Can you just add the close() method without the override or interface?

haribabu-dev:
Ok @stleary , I will do that.

stleary:
Users may want to close the reader associated with a JSONTokener

**Risks**
Low

**Changes to the API?**
A new public API function was added, with unit test

No

No

No

No

**Review status**
**APPROVED**


Starting 3-day comment window

stleary (approved):

jtotht:
> @haribabu-dev Why should JSONTokener implement Closeable? It is not a closeable resource. Can you just add the close() method without the override or interface?

Because it (potentially) uses closable resources. See #718, which would have been closed by the original version of this PR, but not the one that got merged.

Issue #737, opened 2023-05-19
(JSONArray).toList() corrupts entries 


JulWas797:
Unfortunately invoking method `.toList()` on `JSONArray` Object causes entries of the `JSONArray` to become malformated.

**Issue examples:**

Output from `(JSONArray).get()`
`{"data":{"pressure_avg":1012.3399999999999,"humidity_avg":61.5,"temperature_avg":12.85,"pm10_avg":0.6900000000000001,"pm25_avg":0.36},"school":{"city":"DŁUGOŁĘKA","street":null,"post_code":"19-111","latitude":"53.2753471","name":"SZKOŁA PODSTAWOWA W DŁUGOŁĘCE","longitude":"22.8683732"},"timestamp":"2023-05-19 23:56:45"}`

Output from `(JSONArray).toList().get()`
`{data={pressure_avg=1012.3399999999999, humidity_avg=61.5, temperature_avg=12.85, pm10_avg=0.6900000000000001, pm25_avg=0.36}, school={city=DŁUGOŁĘKA, street=null, post_code=19-111, latitude=53.2753471, name=SZKOŁA PODSTAWOWA W DŁUGOŁĘCE, longitude=22.8683732}, timestamp=2023-05-19 23:56:45}`

JulWas797:
I decided to change the title, as it's late for me and i made mistake in it. 

According to your javadocs, this is not a desired behavior.

stleary:
@JulWas797 JSONArray.toString() will output a JSON doc, if that is what you want. We cannot change toList() due to backwards compatibility. Can you describe what you are trying to do?

JulWas797:
> @JulWas797 JSONArray.toString() will output a JSON doc, if that is what you want. We cannot change toList() due to backwards compatibility. Can you describe what you are trying to do?

Thanks for responding!

What I'm trying to basically acheive is to access elements of `JSONArray` using Streams API, and using `.toList()` is the only option to do so. 

As `JSONArray` already has Iterator, i would propose to integrate Streams directly in it (I will also check what I can do, and maybe try to make PR with this).

stleary:
Streams support proposed implementation: #768

Aryant-Tripathi:
Sir @stleary , What will be the expected result we want? I just started putting my steps in the open-source community. I will be glad if I can help :)

ZachsCoffee:
@stleary 
[...]

PR #741, opened 2023-05-29
Configure Java 8 as the minimum required version

javadev:


javadev:
image

michael-o:
This is really unapropriate for a low-level library. There is zero benefit from a user's perspective.

javadev:
> This is really inapropriate for a low-level library. There is zero benefit from a user's perspective. Java 8 is more than enough.

I understand your perspective, but I respectfully disagree. While Java 8 may be sufficient for some use cases, it is important to consider security implications when choosing a minimum secure version for a low-level library.

Java 11 introduced several security enhancements and bug fixes compared to Java 8. It includes updates to the Java Security Architecture, improvements to the TLS protocol, and the introduction of the HTTP/2 protocol, among other security-related features.

By requiring a minimum secure version like Java 11, developers ensure that their library takes advantage of the latest security improvements and patches. This can help prevent potential vulnerabilities and protect users' systems from security threats.

Furthermore, as time progresses, older Java versions may become more susceptible to security risks as new vulnerabilities are discovered. Requiring a more recent version ensures that the library remains compatible with modern security standards.

While there might be an initial cost associated with upgrading to a newer version, the long-term benefits of improved security outweigh the inconvenience. It's essential to prioritize security in software development, especially when developing low-level libraries that could be utilized in various contexts.

Ultimately, the decision of the minimum required version depends on the specific needs and considerations of the library and its users. However, from a security standpoint, Java 11 or a newer version would be a more appropriate choice than Java 8 for a low-level library.

michael-o:
> > This is really inapropriate for a low-level library. There is zero benefit from a user's perspective. Java 8 is more than enough.
> 
> I understand your perspective, but I respectfully disagree. While Java 8 may be sufficient for some use cases, it is important to consider security implications when choosing a minimum secure version for a low-level library.
> 
> Java 11 introduced several security enhancements and bug fixes compared to Java 8. It includes updates to the Java Security Architecture, improvements to the TLS protocol, and the introduction of the HTTP/2 protocol, among other security-related features.
> 
> By requiring a minimum secure version like Java 11, developers ensure that their library takes advantage of the latest security improvements and patches. This can help prevent potential vulnerabilities and protect users' systems from security threats.
> 
> Furthermore, as time progresses, older Java versions may become more susceptible to security risks as new vulnerabilities are discovered. Requiring a more recent version ensures that the library remains compatible with modern security standards.
> 
> While there might be an initial cost associated with upgrading to a newer version, the long-term benefits of improved security outweigh the inconvenience. It's essential to prioritize security in software development, especially when developing low-level libraries that could be utilized in various contexts.
> 
> Ultimately, the decision of the minimum required version depends on the specific needs and considerations of the library and its users. However, from a security standpoint, Java 11 or a newer version would be a more appropriate choice than Java 8 for a low-level library.

What? How does all this matter for a library which parses JSON? TLS? HTTP/2? Totally unrelated. OpenJDK 8 receives all necessary security fixes until 2030. At the end, as a library developer I should not impose something which isn't required from a technical PoV.

javadev:
Twistlock, a security analysis platform, has flagged Java 8 as insecure despite its long-term support and security fixes until 2030. While Java 8 may still receive critical security updates, it's worth considering the recommendations provided by Twistlock to ensure optimal security for your low-level library. Ultimately, as a library developer, it's important to weigh the technical requirements, user needs, and security concerns when determining the minimum required Java version.

javadev:
We can choose to wait for years and continue to maintain Java 1.6 as the minimum required JDK version.

michael-o:
> We can choose to wait for years and continue to maintain Java 1.6 as the minimum required JDK version.

It'd be advisable to release a last 1.6 version with last fixes and move to Java 8 to have a maintained stack with a new (at least) minor version. Here, I doubt that people will object Java 8.

javadev:
This pull request was specifically created for Java 11, indicating that it targets the features and functionality available in that version. Whether the pull request will be closed, merged, or remain open depends on the project maintainers and their evaluation of its compatibility and relevance. While it's possible that some individuals may prefer Java 8, the decision ultimately lies with the project maintainers and their consideration of the pull request's alignment with the project's goals and requirements.

stleary:
I think it's reasonable to require Java 8 if there is a valid PR that uses it. 
For example, #737 would be a good candidate to add an API method that works with Streams, which is only available from Java 8.
Once this is ready I can do a final v6 release, followed by a v8 build. 
Will let the v8 build sit for a month or so in case there are concerns.
I don't think v11 is advisable since some projects, including my current and previous employers, are still stuck on v8.
I think Android is no longer an issue due to its good-enough support for v8.

johnjaylward:
I really don't see a need to drop Java8 at this time. The library is not today using any features above Java 1.6, and form a Bytecode perspective, there isn't a good reason to limit to Java11+. As stated elsewhere in this thread and other issues, Java8 will be around for a long time.

Just because there are valid Application use-cases for using Java11+, this application really does not have a use case for forcing that JVM restriction on downstream developers at this time.

I would not recommend merging this unless the v8 build is put back in with the source and target outputs being v8.

Also, the pipeline.yml can get cleaned up quite a bit. There is no reason to keep the section that was used for the old 1.6 code. Simply using the section dedicated to JVM 8+ would be advisable. I would also not build/test against every JVM version out there. Sticking with the LTS versions (8,11,17) only is a valid approach for libraries.

.github/workflows/pipeline.yml
```diff
@@ -11,19 +11,21 @@ on:
 
 jobs:
   # old-school build and jar method. No tests run or compiled.
-  build-1_6: 
+  build-11:
```
johnjaylward, line 14: This whole pipeline file needs a lot of cleanup. The entire old "build-1_6" (original lines 14 through 39) section should just be deleted.

Keep just the "Build" section

.github/workflows/pipeline.yml
```diff
@@ -42,14 +44,16 @@ jobs:
     strategy:
       matrix:
         # build against supported Java LTS versions:
-        java: [ 8, 11 ]
+        java: [ 11, 17 ]
```
johnjaylward, line 47: Add 8 back in here.

pom.xml
```diff
@@ -93,16 +93,16 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>2.3.2</version>
+                <version>3.11.0</version>
                 <configuration>
-                    <source>1.6</source>
-                    <target>1.6</target>
+                    <source>11</source>
```
johnjaylward, line 98: Source and target should both be set back to 1.8.

The changes from #761 will also need to be incorporated once merged to master

src/test/java/org/json/junit/JSONObjectTest.java
```diff
@@ -620,10 +620,6 @@ public void jsonObjectByBean1() {
         assertTrue("expected h\be\tllo w\u1234orld!", "h\be\tllo w\u1234orld!".equals(jsonObject.query("/escapeStringKey")));
         assertTrue("expected 42", Integer.valueOf("42").equals(jsonObject.query("/intKey")));
         assertTrue("expected -23.45e7", Double.valueOf("-23.45e7").equals(jsonObject.query("/doubleKey")));
-        // sorry, mockito artifact
```
johnjaylward, line 623: If we aren't going to validate the Mockito artifacts here, then I see no reason to use Mockito for this test case at all. It adds no benefit if we remove the assertions against it.

johnjaylward (changes requested):

src/test/java/org/json/junit/JSONObjectTest.java
johnjaylward: I understand these checks don't work against the new version of Mockito which is required for Java 17+, If you can please try to find a solution that will do similar validations.

.github/workflows/pipeline.yml
```diff
@@ -11,19 +11,21 @@ on:
 
 jobs:
   # old-school build and jar method. No tests run or compiled.
-  build-1_6: 
+  build-8:
```
johnjaylward, line 14: This whole `build-8` section still needs to be deleted. All it does is waste time in the pipeline.

johnjaylward (changes requested):

.github/workflows/pipeline.yml
johnjaylward: lines 13-41 should be deleted.

johnjaylward (approved):
I'm OK with this set of changes as long as @stleary agrees with the direction.

johnjaylward:
Why are you pushing more commits after approval? You are making me feel like I'm wasting my time.

javadev:
I noticed that the master branch isn't currently up to date. Therefore, I took the initiative to rebase the branch with the most current version.

javadev:
> 

I apologize if my actions have caused any inconvenience. The reason for pushing additional commits after approval is to ensure the code's integrity and address any potential issues that may have been missed earlier. Please know that your time and contributions are valued, and I'm committed to delivering the best quality work. If you prefer a different approach or have specific concerns, please feel free to share them, and I'll do my best to accommodate your preferences.

johnjaylward:
I got an email that looked like this:

> [@javadev](https://github.com/javadev) requested your review on: [#741](https://github.com/stleary/JSON-java/pull/741) Configure Java 8 as the minimum required version.

So I re-reviewed and approved, then you pushed again... then I got another email...

If it's automated by GitHub, then there isn't much you could have done about it I suppose, but if there was a button you hit to re-request review, I'd ask that you refrain from hitting the button before finalizing your changes.

The other option would be to place the PR in draft mode, so that reviewers know you are still working out changes.

johnjaylward (approved):

stleary:
The maven build works fine with Java 8, but this error with gradlew. Did something get missed?
image

javadev:
> The maven build works fine with Java 8, but this error with gradlew. Did something get missed? image

I have prepared a solution.

stleary:
Update builds to support Java 8 and above.
Java 6 and 7 users will need to stay with v20230618.

**Risks**
Moderate

**Changes to the API?**
No

Yes

No

No

No

**Review status**
**APPROVED**
This change will be included in the next release, which will also need to include a useful feature that requires Java 8.

javadev:
[johnjaylward](https://github.com/johnjaylward) reviewed [3 days ago](https://github.com/stleary/JSON-java/pull/741#pullrequestreview-1599351650)

stleary:
Will merge when we have a feature that requires Java 8, e.g. #737

johnjaylward (approved):

johnjaylward:
> Will merge when we have a feature that requires Java 8, e.g. #737

I think that's putting the horse (v8 support) after the cart (v8 features). Merging this will allow version 8 features to be implemented.

I think this should be merged before features. Also, if this waits too long it will become stale and a pain to maintain if there are conflicting changes.

stleary:
Good point @johnjaylward

stleary (approved):

PR #744, opened 2023-06-08
Added object methods for methods returning primitives

davejbur:
Added object methods for:
- optBooleanObj (returns Boolean vice boolean)
- optFloatObj (returns Float vice float)
- optIntegerObj (returns Integer vice integer)
- optLongObj (returns Long vice long)

The reason for doing so is that, sometimes, I want (e.g.) ``optInt`` to return ``null`` if a value isn't found (i.e. I want to do ``optInt(theKey,null))``. However, that gives a compiler error ``<null> cannot be converted to int``, since ``optInt`` needs to return an ``int``. Hence adding an almost identical function that returns an ``Integer`` instead.

(I've finally worked out why, in a previous attempt at doing this, Github was trying to allege I'd changed the whole file...! GH was changing all the line endings for me instead of just leaving well alone.)

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -1132,6 +1132,45 @@ public class JSONObject {
     }
 
     /**
+     * Get an optional boolean object associated with a key. It returns false if there
+     * is no such key, or if the value is not Boolean.TRUE or the String "true".
+     *
+     * @param key
+     *            A key string.
+     * @return The truth.
+     */
+    public Boolean optBooleanObj(String key) {
+        return this.optBooleanObj(key, false);
+    }
+
+    /**
+     * Get an optional boolean object associated with a key. It returns the
+     * defaultValue if there is no such key, or if it is not a Boolean or the
+     * String "true" or "false" (case insensitive).
+     *
+     * @param key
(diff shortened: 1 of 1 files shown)


javadev:
I suggest these method names

```
getOptionalBooleanObject()
getOptionalFloatObject()
getOptionalIntegerObject()
getOptionalLongObject()
```

spaffrath:
@davejbur Can't this be done using the current API with `optInt("x", (Integer) null)` and so on?

stleary:
I believe `optInt("x", (Integer) null)` will throw a NullPointerException because the 2nd param is a primitive int.
But this code could work:
```
Integer i = null;
if (j.has("a")) {
    i = j.getInt("a");
}
```

spaffrath:
@stleary Yes, of course 👍

davejbur:
> I suggest these method names
> 
> ```
> getOptionalBooleanObject()
> getOptionalFloatObject()
> getOptionalIntegerObject()
> getOptionalLongObject()
> ```

Thanks - I was trying to keep in line with the optBoolean etc naming convention already in use.

davejbur:
Closing this as now dealt with by #753.

Issue #745, opened 2023-06-12
org.json.junit.XMLTest.testIssue537CaseSensitiveHexEscapeFullFile fails


davejbur:
I'm having problems compiling, as testIssue537CaseSensitiveHexEscapeFullFile fails. I'm using:
- Windows 11 64bit
- Netbeans 16
- JDK1.8

The error I get is:
```
org.json.junit.XMLTest.testIssue537CaseSensitiveHexEscapeFullFile -- Time elapsed: 0.056 s <<< FAILURE!
org.junit.ComparisonFailure: values should be equal expected:<... sensitive CRP, IL-1[?])> but was:<... sensitive CRP, IL-1[╬▓])>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.json.junit.Util.compareActualVsExpectedObjects(Util.java:93)
	at org.json.junit.Util.compareActualVsExpectedJsonObjects(Util.java:52)
	at org.json.junit.Util.compareActualVsExpectedObjects(Util.java:68)
	at org.json.junit.Util.compareActualVsExpectedJsonArrays(Util.java:33)
	at org.json.junit.Util.compareActualVsExpectedObjects(Util.java:74)
	at org.json.junit.Util.compareActualVsExpectedJsonObjects(Util.java:52)
	at org.json.junit.Util.compareActualVsExpectedObjects(Util.java:68)
	at org.json.junit.Util.compareActualVsExpectedJsonObjects(Util.java:52)
	at org.json.junit.XMLTest.testIssue537CaseSensitiveHexEscapeFullFile(XMLTest.java:923)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
[...]

PR #747, opened 2023-06-15
Designed and tested to use on Java versions 1.6 - 17

javadev:


stleary:
I think we need v17 in the GitHub actions pipeline in order to claim designed and tested.
But that will probably require #745 to be fixed first. I believe that #745 is an issue with the unit tests.

javadev:
Let us exercise patience while we await the resolution of the bug fixing process.

Issue #714, opened 2022-12-16
Add object-based versions of optInt & optFloat


davejbur:
When reading a JSON file, I would like to return a null value if a key is not found. At present, optInt and optFloat only return primitives - so you can't set null as the default.

It's a fairly simple change, to add in the following code to JSONObject:
```
    /**
     * Get the optional double (obj) value associated with an index. NaN is returned
     * if there is no value for the index, or if the value is not a number and
     * cannot be converted to a number.
     *
     * @param key
     *            A key string.
     * @return The value.
     */
    public Float optFloatObj(String key) {
        return this.optFloatObj(key, Float.NaN);
    }

    /**
     * Get the optional double (obj) value associated with an index. The defaultValue
     * is returned if there is no value for the index, or if the value is not a
     * number and cannot be converted to a number.
     *
     * @param key
     *            A key string.
     * @param defaultValue
     *            The default value.
     * @return The value.
     */
    public Float optFloatObj(String key, Float defaultValue) {
        Number val = this.optNumber(key);
        if (val == null) {
            return defaultValue;
        }
        final Float floatValue = val.floatValue();
        // if (Float.isNaN(floatValue) || Float.isInfinite(floatValue)) {
        // return defaultValue;
        // }
        return floatValue;
    }

    /**
     * Get an optional integer value associated with a key, or zero if there is no
     * such key or if the value is not a number. If the value is a string, an
     * attempt will be made to evaluate it as a number.
     *
     * @param key
     *            A key string.
     * @return An object which is the value.
     */
    public Integer optIntegerObj(String key) {
        return this.optIntegerObj(key, 0);
    }

    /**
     * Get an optional integer value associated with a key, or the default if there
     * is no such key or if the value is not a number. If the value is a string,
     * an attempt will be made to evaluate it as a number.
     *
     * @param key
     *            A key string.
     * @param defaultValue
     *            The default.
     * @return An object which is the value.
     */
    public Integer optIntegerObj(String key, Integer defaultValue) {
        final Number val = this.optNumber(key, null);
        if (val == null) {
            return defaultValue;
        }
        return val.intValue();
    }
```

Unfortunately I've forgotten all the little I learnt about forking/creating branches/submitting pull requests etc, and Github doesn't make it particularly easy unless you're already familiar with how to do it - I've made the changes over on https://github.com/davejbur/JSON-java/tree/add-object-based-versions-int-float but the diff seems to think I've deleted and redone huge chunks instead of just adding four methods!

[...]

PR #753, opened 2023-06-29
Updated new object methods

davejbur:
Further to #744 & #714, I've:
- Renamed the object methods from ...Obj to ...Object.
- Added an object method for optDoubleObject (returns Double vice double).
- Added similar methods in JSONArray.
- Added test methods.

_To recap, #744 was the following:_
_- optBooleanObj (returns Boolean vice boolean)_
_- optFloatObj (returns Float vice float)_
_- optIntegerObj (returns Integer vice integer)_
_- optLongObj (returns Long vice long)_
_The reason for doing so is that, sometimes, I want (e.g.) optInt to return null if a value isn't found (i.e. I want to do optInt(theKey,null)). However, that gives a compiler error <null> cannot be converted to int, since optInt needs to return an int. Hence adding an almost identical function that returns an Integer instead._

I don't know why I couldn't update the existing PR, it wouldn't let me, hence the new one containing all the above changes in one hit.

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -600,6 +600,38 @@ public class JSONArray implements Iterable<Object> {
     }
 
     /**
+     * Get the optional Boolean object associated with an index. It returns false
+     * if there is no value at that index, or if the value is not Boolean.TRUE
+     * or the String "true".
+     *
+     * @param index
+     *            The index must be between 0 and length() - 1.
+     * @return The truth.
+     */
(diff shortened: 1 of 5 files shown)


stleary:
API should allow optional wrapper class access (e.g. Double object vs double primitive)

**Risks**
Low

**Changes to the API?**
New opt* methods have been added

No

No

No, unit tests were added

No

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

PR #754, opened 2023-07-10
Added the control to use self closing tag in XML module and added two unit cases to test the same.

madhavgupta2011:
Added the control to use self closing tag in XML module and added two unit cases to test the same.

### DIFF
--- a/src/main/java/org/json/XML.java
+++ b/src/main/java/org/json/XML.java
@@ -782,6 +782,23 @@ public class XML {
      */
     private static String toString(final Object object, final String tagName, final XMLParserConfiguration config, int indentFactor, int indent)
             throws JSONException {
+        return toString(object, tagName, config, indentFactor, indent, true);
+    }
+
+    /**
+     * Convert a JSONObject into a well-formed, element-normal XML string,
+     * either pretty print or single-lined depending on indent factor as well as having the choice to keep self closing tags.
+     *
+     * @param object              A JSONObject.
+     * @param tagName             The optional name of the enclosing tag.
+     * @param config              Configuration that can control output to XML.
+     * @param indentFactor        The number of spaces to add to each level of indentation.
+     * @param indent              The current ident level in spaces.
+     * @param keepSelfClosingTags The boolean value to either keep or remove self-closing tags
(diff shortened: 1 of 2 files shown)


stleary:
* Tests are failing, please fix. 
* The default behavior should not change. Instead, try adding a property to XMLParserConfiguration to allow the developer to change the empty tag behavior. This object is passed to the private XML toString() method, so you don't need to add another parameter.

madhavgupta2011:
Sure. Will do it. Thanks @stleary

src/main/java/org/json/XML.java
```diff
@@ -853,8 +870,15 @@ private static String toString(final Object object, final String tagName, final
                     sb.append(indent(indent));
                     sb.append('<');
                     sb.append(key);
-                    sb.append("/>");
-                    if(indentFactor > 0){
+                    if (!keepSelfClosingTags) {
```
javadev, line 873: `if (!condition) code1 else code2`

can be replaces with

`if (condition) code2 else code1`

javadev:
```
[INFO] Finished at: 2023-07-11T12:12:16Z
[INFO] ------------------------------------------------------------------------
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.3.2:testCompile (default-testCompile) on project json: Compilation failure
Error:  /home/runner/work/JSON-java/JSON-java/src/test/java/org/json/junit/XMLTest.java:[1325,33] error: no suitable method found for toString(JSONObject,boolean)
```

src/main/java/org/json/XML.java
madhavgupta2011: Yeah, sure. I'll change it. I am thinking to change my approach though.

javadev:
Since July 24, there have been no changes. Therefore, I believe it would be appropriate to consider canceling the pull request.

stleary:
@madhavgupta2011 Are you able to make the recommended corrections for this PR?
Do you think there is still a need for this change?

stleary:
Thanks, but not accepting this PR. The issue has been reopened.

PR #759, opened 2023-08-01
JSON parsing should detect embedded `\0` values

### LINKED ISSUE #758: Confusion between `\0` and EOF can lead to OutOfMemoryError
`JSONTokener.next()` [uses](https://github.com/stleary/JSON-java/blob/402db6ad84353c84e4747510a8ac5322b32a8df4/src/main/java/org/json/JSONTokener.java#L182) a 0 return to indicate EOF. But 0 is also returned when an actual `\0` character is read. In some circumstances that can be used to circumvent parser checks. Parsing untrusted input could then potentially lead to OutOfMemoryError even for quite small input strings.

eamonnmcmanus:
A better solution might be to use -1 instead 0 to represent EOF everywhere, which of course means changing `char` variables to `int`. The solution here is enough to solve the immediate problem, though.

Fixes #758.

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -253,12 +253,11 @@ public class JSONObject {
             switch (x.nextClean()) {
             case ';':
             case ',':
-                c = x.nextClean();
-                if (c == 0) {
-                    throw x.syntaxError("A JSONObject text must end with '}'");
-                }
-                if (c == '}') {
+                if (x.nextClean() == '}') {
                     return;
+                }
+                if (x.end()) {
+                    throw x.syntaxError("A JSONObject text must end with '}'");
                 }
                 x.back();
                 break;
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -253,7 +253,11 @@ public class JSONObject {
             switch (x.nextClean()) {
             case ';':
             case ',':
-                if (x.nextClean() == '}') {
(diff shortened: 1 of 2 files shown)


src/main/java/org/json/JSONObject.java
```diff
@@ -253,7 +253,11 @@ public JSONObject(JSONTokener x) throws JSONException {
             switch (x.nextClean()) {
             case ';':
             case ',':
-                if (x.nextClean() == '}') {
+                c = x.nextClean();
+                if (c == 0) {
```
johnjaylward, line 257: maybe change to `if (x.end()) {`

`end` would match the expected use of the Tokener API better.

eamonnmcmanus: Yes, that's cleaner!

johnjaylward (approved):

stleary:
JSONObject Tokenizer does not handle chars with value '\0' correctly, since this value is internally used to signal the end of text. JSONArray does not have this problem because this returned value results in a syntax exception.

**Risks**
Low

**Changes to the API?**
No

No

No

No. A new unit test was added.

No

**Review status**
**APPROVED**

eamonnmcmanus:
> **Will this require a new release?**
> No

There probably is going to be a CVE for the DoS issue, so you might want to plan for a release that includes the fix.

stleary (approved):

stleary:
No worries, a new version can be released if/when the CVE is added

PR #769, opened 2023-09-08
Addressed Java 17 compile warnings

javadev:


stleary:
@javadev Thanks for taking care of this, will review later today.

johnjaylward (changes requested):
It looks like there is a warning about the bootstrap not being set properly.

Source Main:

image

Source Test:

image


Can you try a few option in the Maven configuration to see about adding the new "release" flag?

I found this StackOverflow relating to the warning: https://stackoverflow.com/questions/59097317/warning-options-bootstrap-class-path-not-set-in-conjunction-with-source-8

johnjaylward:
> It looks like there is a warning about the bootstrap not being set properly.
> 

Do note that this may be a non-issue as the warning is relating to accidentally compiling with a newer version of the JDK, thus accidentally using new APIs that may not be available in the old JDK being targeted. Since our build pipeline is cross compiling with explicit versions of the JDK that we support, this should NOT be an issue as the JDK 8 part of the pipeline would fail if someone accidentally used an API that wasn't available in JDK 8.

Point being, If this isn't easily fixable, I would not worry about it.

javadev:
@stleary 

The pull request is prepared for merging.

johnjaylward (approved):
Thanks for looking into it. At this time I don't think multiple POM are needed based on the documentation for the warning.

stleary (approved):

stleary:
Fixes unit test compiler warnings when building with Java 17 (the pipeline does this)

**Risks**
Low - this only affects the unit tests

**Changes to the API?**
N/A

No

No

No

N/A

**Review status**
**APPROVED**

Starting 3-day comment window

javadev:
[stleary](https://github.com/stleary) commented [3 days ago](https://github.com/stleary/JSON-java/pull/769#issuecomment-1718231226)

PR #770, opened 2023-09-19
Small test fixes.

eamonnmcmanus:
One test method was missing `@Test` so it was never run.

One test method used another test class as the base for finding a test resource. While this works in practice with Maven, it is not strictly right.


johnjaylward (approved):

stleary:
Fixed 2 unit tests

**Risks**
Low

**Changes to the API?**
N/A

No

No

No

N/A

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

PR #772, opened 2023-09-20
Disallow nested objects and arrays as keys in objects.

### LINKED ISSUE #771: Logic to exclude object keys that are themselves objects is imperfect
`JSONObject` contains [this logic](https://github.com/stleary/JSON-java/blob/01727fd0ed00c1f1f5582bdc97244356ac029f3b/src/main/java/org/json/JSONObject.java#L218-L222) to prevent a key in a JSON object from being another JSON object or an array:
```java
            case '{':
            case '[':
                if(prev=='{') {
                    throw x.syntaxError("A JSON Object can not directly nest another JSON Object or JSON Array.");
                }
```
However this doesn't cover all cases. For example, `JSONObject` currently accepts this string:
```
{"a":1, [{"b": 2}]: 3}
```
which it interprets as an object with two keys, one that is the string `a` and one that is the string `[{"b":2}]`.

Strict JSON, of course, only accepts string literals as keys in objects. JSON-java is more liberal, but it does balk at nested objects since the fix for #654. The example above suggests that that fix was incomplete. I think it would make sense to disallow nested objects or arrays as keys always. They're not valid JSON and it's unlikely that anyone is relying on them for legitimate purposes. Meanwhile they can straightforwardly be used for DoS attacks, similar to #654 and #758.

I have a proposed fix which I will send shortly.

eamonnmcmanus:
Fixes #771.

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -208,22 +208,14 @@ public class JSONObject {
             throw x.syntaxError("A JSONObject text must begin with '{'");
         }
         for (;;) {
-            char prev = x.getPrevious();
             c = x.nextClean();
             switch (c) {
             case 0:
                 throw x.syntaxError("A JSONObject text must end with '}'");
             case '}':
                 return;
-            case '{':
-            case '[':
-                if(prev=='{') {
-                    throw x.syntaxError("A JSON Object can not directly nest another JSON Object or JSON Array.");
-                }
-                // fall through
             default:
-                x.back();
-                key = x.nextValue().toString();
+                key = x.nextSimpleValue(c).toString();
             }
 
             // The key is followed by ':'.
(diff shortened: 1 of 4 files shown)


stleary:
@eamonnmcmanus I think this is a good fix and probably overdue. However, since JSONTokenizer is central to the lib, more testing would be helpful. For example, the non-quoted value might be an object key, object value, or an array element. The value itself might be a simple string or might contain one or more JSON control chars: backslash, curly brace, bracket, comma, colon, or double quote. It might be number-like (contains one or more periods, hyphens, plus signs, or 'e' chars). It would be good to know if the tokenizer processes these examples in a reasonable way. There might also be other potential tests I haven't thought of. This could get tedious to code and might be more than you signed up for. Let me know if you want some help with the unit tests. Either way, thanks for getting this started.

eamonnmcmanus:
I've added some extra tests to cover unquoted text. They are not as extensive as listed in @stleary's [comment](https://github.com/stleary/JSON-java/pull/772#issuecomment-1738283598). Further test coverage for unquoted text would surely be good, but I think is a bit out of scope for this PR. The goal here is to fix a DoS problem, while not breaking any existing reasonable uses. I think the tests cover that quite well.

stleary (approved):

stleary:
Unquoted object and array values should be restricted to simple types

**Risks**
Moderate. This is a change to existing behavior and may impact users.

**Changes to the API?**
No

No

No

No. New unit tests were added.

A test method name typo was fixed.

**Review status**
**APPROVED**

PR #773, opened 2023-09-25
Add optJSONArray method to JSONObject with a default value

eedijs:
The `JSONObject` class had a missing `optJSONArray` method with a default return value, which is very useful in some cases, so you don't have to write code like this:

```java
JSONArray jsonArray = jsonObject.has("key") ? jsonObject.getJSONArray("key") : new JSONArray;
```

And instead just write:

```java
JSONArray jsonArray = jsonObject.optJSONArray("key", new JSONArray());
```

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -924,30 +924,57 @@ public class JSONArray implements Iterable<Object> {
     }
 
     /**
-     * Get the optional JSONArray associated with an index.
+     * Get the optional JSONArray associated with an index. Null is returned if
+     * there is no value at that index or if the value is not a JSONArray.
      *
      * @param index
-     *            subscript
-     * @return A JSONArray value, or null if the index has no value, or if the
-     *         value is not a JSONArray.
+     *            The index must be between 0 and length() - 1.
+     * @return A JSONArray value.
      */
     public JSONArray optJSONArray(int index) {
-        Object o = this.opt(index);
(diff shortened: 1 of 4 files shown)


stleary:
@eedijs Thanks, agree it should be possible to return a default value. Is there a reason why this method was not also added to JSONArray?

eedijs:
> @eedijs Thanks, agree it should be possible to return a default value. Is there a reason why this method was not also added to JSONArray?

Sorry, didn't even check the `JSONArray` class. Seems like it is also missing the `optJSONObject` method with a default value.

I will go over both `JSONObject` and `JSONArray` classes once more and see if anything else need an `opt` method with default values.

stleary:
I think there's a good case to be made for optJSONArray() with default. Probably not necessary to do an exhaustive review of all of the opt methods.

eedijs:
> I think there's a good case to be made for optJSONArray() with default. Probably not necessary to do an exhaustive review of all of the opt methods.

Ok, I just commited `optJSONArray` and `optJSONObject` methods for the `JSONArray` class.

stleary:
Opt methods for JSONArray and JSONObject should allow a default value to be returned.

**Risks**
Low

**Changes to the API?**
Yes, new API methods are added where a default parameter is included. The default param is the same type as the return value.

No

No

No. New unit tests were added for the new API methods.

No


**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

PR #775, opened 2023-09-30
Upgrading to Gradle 8.3

mccartney:
Upgrading to Gradle 8.3 for no specific reason other than staying up to date.

This PR is the result of:
- editing the properties file to name `8.3` instead of `6.3`
- and running `gradlew wrapper`


stleary:
@mccartney  A detailed explanation for these changes is needed.

johnjaylward (approved):
@stleary This looks like a fairly standard Gradle Wrapper update. The newer versions of Gradle support newer versions of Java but can still be run on Java 8 if needed.

In fact to support building for Java 17 (as we do with the maven build) we would need to upgrade to at least Gradle 7.3

https://docs.gradle.org/current/userguide/compatibility.html

stleary:
Thanks for the PR, but not accepting a Gradle upgrade at this time.

Issue #748, opened 2023-06-16
Add a closing tag flag for XML output


derily:
when JSONObject value is empty string,i need XML.toString output `<name></name> not <name/>`

stleary:
@derily Can you provide a sample input and the output that you want to get?

derily:
> @derily Can you provide a sample input and the output that you want to get?

input(JSONObject):
`{
  "name":"derily",
  "job":""
}`
output:
`<name>derily</name>
<job/>
`
expect(just not use self closing tag syntax):
`<name>derily</name>
<job></job>
`

stleary:
@derily It might be possible to accommodate this with an XMLConfiguration option. If no one else picks this up, I will try to get something working this weekend.

madhavgupta2011:
Hi, Can I take this up? I'm thinking of adding one more boolean parameter in toString() method of XML class handle the string as per the boolean parameter. Please let me know if this solution can be improved. I'm willing to take up on this issue.

stleary:
@madhavgupta2011 Thanks, yes you can work on this ticket - FYI you don't need to ask first. Please check whether you can use XMLParserConfiguration to trigger the new behavior.

madhavgupta2011:
@stleary I have opened a pull request with my changes and added two unit test cases as well. 
Can you please review it?
[Link to PR](https://github.com/stleary/JSON-java/pull/754)

stleary:
Fix in progress - #754.

stleary:
Re-opening since the PR was not accepted

stleary:
Please use XMLParserConfiguration for the implementation.

dubeyaditya29:
Please ignore the above commits, I am still on it

dubeyaditya29:
https://github.com/dubeyaditya29/JSON-java/commit/fa49d83612c5494a659423a34343c966b57ec303

@stleary  can you verify this.

stleary:
See comments in your repo.

rudrajyotib:
@stleary  - is this issue still open to be picked?

rudrajyotib:
@stleary - if no one else is working on it, can you assign it to my name?

stleary:
@rudrajyotib You can work on this issue. The new behavior should be opt-in, by adding a new property to XMLParserConfiguration.

rudrajyotib:
@stleary sure, I'll start working on it and share the design.

rudrajyotib:
[...]

Issue #713, opened 2022-12-13
JSONObject constructors do not check for non-finite numbers


Madjosz:
**Problem**
The constructors `org.json.JSONObject.JSONObject(Map<?, ?>)` and `org.json.JSONObject.JSONObject(Object)` do not check occuring numbers in values for finiteness and thus allow creating of invalid JSON object which then run into strange behaviour when stringifying.

The `Map` constructor even mentiones `@throws JSONException - If a value in the map is non-finite number.` but this exception can only occur here when a nested call throws it, e.g. a value with a `List` containing non-finite numbers.

**Code example**
```java
import static org.junit.jupiter.api.Assertions.assertThrows;
import java.util.Map;
import org.json.JSONException;
import org.json.JSONObject;
import org.junit.jupiter.api.Test;

class JSONObjectTest {
    @Test void jsonObject_map() {
        assertThrows(JSONException.class, () -> new JSONObject(Map.of("a", Double.POSITIVE_INFINITY)));
    }

    @Test void jsonObject_bean() {
        assertThrows(JSONException.class, () -> new JSONObject(new MyBean()));
    }
    public static class MyBean {
        public double getA() { return Double.POSITIVE_INFINITY; }
    }
}
```
Both tests fail and instead the constructors return an instance of a `JSONObject` with a key `"a"` mapped to a non-finite double.

**Cascading problems**
Calling `toString()` on the returned `JSONObject` will internally throw a `JSONException` in `toString(0)` while writing the non-finite value and this will be catched resulting in a `null` returned from the method. `obj.getDouble("a")` on the other hand will return the non-finite double.

**Idea**
While `JSONObject.put(String, Object)` checks for finiteness before putting the value to the map https://github.com/stleary/JSON-java/blob/5920eca2d7694c8d9a714bb280f31007ff64ade4/src/main/java/org/json/JSONObject.java#L1870-L1871 the other two calls https://github.com/stleary/JSON-java/blob/5920eca2d7694c8d9a714bb280f31007ff64ade4/src/main/java/org/json/JSONObject.java#L291 https://github.com/stleary/JSON-java/blob/5920eca2d7694c8d9a714bb280f31007ff64ade4/src/main/java/org/json/JSONObject.java#L1552 just `wrap()` the value so it might be a good idea to do the finiteness-check there and document the JSONException.

tsufiyan:
Shall I work on this? if yes, please assign

stleary:
[...]

PR #777, opened 2023-10-03
fix: Added Non-Finite check

dubeyaditya29:
This PR fixes the missing check for non-finite values being added to the map object in JSONObject 

Added a new method `isFiniteNumber(Object value)`  returns a boolean value false in case the value is non-finite.
The wrap throws a runtime exception in case the value is non-finite.


dubeyaditya29:
#713

stleary:
@dubeyaditya29 Good start, but more work is needed.

Please revert the change in pom.xml:
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project json: Fatal error compiling: invalid target release: 9 -> [Help 1]

Are there any other numeric types that should be checked?
Does JSONArray need this check too?
Each new line of code should receive coverage from one or more of the new unit tests.

src/main/java/org/json/JSONObject.java
```diff
@@ -2629,8 +2632,34 @@ public static Object wrap(Object object) {
         return wrap(object, null);
     }
 
+    // Custom isFiniteNumber method
+    public static boolean isFiniteNumber(Object value) {
```
johnjaylward, line 2636: This method is sub-optimal, and redundant of `testValidity` and `numberIsFinite` methods that are already available. Please use the existing function(s)

pom.xml
```diff
@@ -95,8 +95,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.11.0</version>
                 <configuration>
-                    <source>1.8</source>
```
johnjaylward, line 98: As stated already, this needs to be reverted. Java 1.8 is our current target.

src/main/java/org/json/JSONObject.java
```diff
@@ -2629,8 +2632,34 @@ public static Object wrap(Object object) {
...
+            return Double.isFinite(doubleValue);
+        } else {
+            // Value is not a Number
+            return false;
+        }
+    }
     private static Object wrap(Object object, Set<Object> objectsRecord) {
         try {
+            // Calling the isFiniteNumber(Object object) -> to check if the object is Non-finite.
+            if(!isFiniteNumber(object)){
```
johnjaylward, line 2660: this check should happen after the `NULL` check.

johnjaylward (changes requested):

src/main/java/org/json/JSONObject.java
```diff
@@ -2629,8 +2632,34 @@ public static Object wrap(Object object) {
         return wrap(object, null);
     }
 
+    // Custom isFiniteNumber method
+    public static boolean isFiniteNumber(Object value) {
+        if (value instanceof Number) {
+            double doubleValue;
+            if (value instanceof Integer) {
+                doubleValue = ((Integer) value).doubleValue();
```
Madjosz, line 2640: Integers are always finite.

johnjaylward: agreed. Using the existing `testValidity` method would be the correct course and already covers what this PR is attempting.

dubeyaditya29:
hey @stleary Thankyou for reviewing this.
1. Have added Integer, Float, Double, I will add Long as well, let me know if missing any specific type.
2. Let me add test cases for that as well will get back on this.
3. For the last point I have created a different test class file for checking the JSONObject, let me know if I have missed something will add that too.

pom.xml
dubeyaditya29: I'll made the necessary changes here thanks.

src/main/java/org/json/JSONObject.java
dubeyaditya29: agree @johnjaylward  the testValidity can alone verify this, am somehow new to this code base, can you help me understand how did you found this method.

dubeyaditya29:
thanks to @johnjaylward suggestions I think there is no need for defining a new method here we already have `testValidaity` which can fulfil the requirements.

src/test/java/org/json/junit/data/JsonObjectTest.java
```diff
@@ -0,0 +1,26 @@
+package org.json.junit.data;
```
Madjosz, line 1: Tests should be added to the existing class `org.json.junit.JSONObjectTest`. I would suggest to name the methods `issue713...()` analogously to already existing issue-related test methods in that class.

src/main/java/org/json/JSONObject.java
```diff
@@ -2629,8 +2632,34 @@ public static Object wrap(Object object) {
...
+        } else {
+            // Value is not a Number
+            return false;
+        }
+    }
     private static Object wrap(Object object, Set<Object> objectsRecord) {
         try {
+            // Calling the isFiniteNumber(Object object) -> to check if the object is Non-finite.
+            if(!isFiniteNumber(object)){
+                throw new RuntimeException("Non-Finite Number encountered");
```
Madjosz, line 2661: Throw `JSONException`, not a generic `RuntimeException`.

johnjaylward:
#779 is a better implementation. This should be rejected

stleary:
Thanks, but not accepting this PR.

dubeyaditya29:
No problem @stleary thanks for reviewing.

stleary:
@dubeyaditya29 if you want to work on a different issue, I recommend #748, which I just reopened. Let me know if you have any questions.

PR #779, opened 2023-10-04
add validity check for JSONObject constructors

### LINKED ISSUE #713: JSONObject constructors do not check for non-finite numbers

Madjosz:
* fixes #713
* document JSONException in JavaDoc
* remove unused Comparable<T> boundary to reuse GenericBean in test

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -283,6 +283,7 @@ public class JSONObject {
         	    }
                 final Object value = e.getValue();
                 if (value != null) {
+                    testValidity(value);
                     this.map.put(String.valueOf(e.getKey()), wrap(value));
                 }
             }
@@ -346,6 +347,8 @@ public class JSONObject {
      * @param bean
      *            An object that has getter methods that should be used to make
      *            a JSONObject.
+     * @throws JSONException
+     *            If a getter returned a non-finite number.
      */
     public JSONObject(Object bean) {
         this();
@@ -1691,6 +1694,8 @@ public class JSONObject {
      *
      * @param bean
      *            the bean
+     * @throws JSONException
+     *            If a getter returned a non-finite number.
      */
     private void populateMap(Object bean) {
         populateMap(bean, Collections.newSetFromMap(new IdentityHashMap<Object, Boolean>()));
@@ -1726,6 +1731,7 @@ public class JSONObject {
 
                             objectsRecord.add(result);
 
+                            testValidity(result);
                             this.map.put(key, wrap(result, objectsRecord));
 
                             objectsRecord.remove(result);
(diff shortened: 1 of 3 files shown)


stleary (approved):

stleary:
@Madjosz Your branch may be out of sync due to recent changes in JSONObjectTest.java.

johnjaylward (approved):

stleary:
Numeric inputs in JSONObject constructors should check for validity.

**Risks**
Low

**Changes to the API?**
No

No

No

No, new unit tests were added

A few minor refactorings unrelated to this commit are included.

**Review status**
**APPROVED**

Starting 3-day comment window

PR #782, opened 2023-10-05
Fix XMLTest.testIndentComplicatedJsonObjectWithArrayAndWithConfig() for Windows - in the test

### LINKED ISSUE #781: XMLTest.testIndentComplicatedJsonObjectWithArrayAndWithConfig fails on Windows
When running the test suite on Windows, XMLTest.testIndentComplicatedJsonObjectWithArrayAndWithConfig fails due to differences in the newline character being used in the expected and actual XMLs being compared - differences that are no functionally interesting.

The test properly canonized the newline character in the actual string produced by calling `XML.toString`, but fails to do so for the expected string that was read from a file. Since this file was not produced on a Windows machine, it won't contain the linebreaks expected when the test is run on Windows.

mureinik:
`XMLTest.testIndentComplicatedJsonObjectWithArrayAndWithConfig` fails when run on Windows due to mismatching linebreaks (that aren't important for the test's functionality) between the actual and expected strings.

For the actual strings, linebreaks are canonized to the platform's native linebreak using `replaceAll("\\n|\\r\\n",
System.getProperty("line.separator")`. However, the expected result is read from a file, and is left with the linebreaks that were originally used to create it.

The solution is to perform the same canonization on both strings.

Closes #781

### DIFF
--- a/src/test/java/org/json/junit/XMLTest.java
+++ b/src/test/java/org/json/junit/XMLTest.java
@@ -1239,7 +1239,8 @@ public class XMLTest {
                     for (int numRead; (numRead = in.read(buffer, 0, buffer.length)) > 0; ) {
                         expected.append(buffer, 0, numRead);
                     }
-                    assertEquals(expected.toString(), actualString.replaceAll("\\n|\\r\\n", System.getProperty("line.separator")));
+                    assertEquals(expected.toString().replaceAll("\\n|\\r\\n", System.getProperty("line.separator")),
+                            actualString.replaceAll("\\n|\\r\\n", System.getProperty("line.separator")));
                 } finally {
                     if (xmlStream != null) {
                         xmlStream.close();
--- a/src/test/java/org/json/junit/XMLTest.java
+++ b/src/test/java/org/json/junit/XMLTest.java
@@ -1239,8 +1239,8 @@ public class XMLTest {
                     for (int numRead; (numRead = in.read(buffer, 0, buffer.length)) > 0; ) {
                         expected.append(buffer, 0, numRead);
                     }
-                    assertEquals(expected.toString().replaceAll("\\n|\\r\\n", System.getProperty("line.separator")),
-                            actualString.replaceAll("\\n|\\r\\n", System.getProperty("line.separator")));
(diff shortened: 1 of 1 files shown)


src/test/java/org/json/junit/XMLTest.java
```diff
@@ -1239,7 +1239,8 @@ public void testIndentComplicatedJsonObjectWithArrayAndWithConfig(){
                     for (int numRead; (numRead = in.read(buffer, 0, buffer.length)) > 0; ) {
                         expected.append(buffer, 0, numRead);
                     }
-                    assertEquals(expected.toString(), actualString.replaceAll("\\n|\\r\\n", System.getProperty("line.separator")));
+                    assertEquals(expected.toString().replaceAll("\\n|\\r\\n", System.getProperty("line.separator")),
+                            actualString.replaceAll("\\n|\\r\\n", System.getProperty("line.separator")));
```
johnjaylward, line 1243: Let's update these to use the more consise `System.lineSeparator()` now that it's supported instead of the longer `System.getProperty("line.separator")`.

johnjaylward (changes requested):

src/test/java/org/json/junit/XMLTest.java
mureinik: @johnjaylward good point, done.

(side note: not sure about this project's etiquette for code reviews, so I'm leaving this comment unresolved so you can review the change and decide yourself whether to resolve it or request further changes)

johnjaylward: I don't see an option to resolve the conversation. I believe only you (as the opener) or @stleary is able to resolve, so please resolve the discussion if you feel the changes have met the feedback request. If more changes are needed, a new review conversation can be started

image

johnjaylward (approved):

stleary:
Fixes a unit test that somehow became broken in Windows

**Risks**
Low

**Changes to the API?**
N/A

No

No

N/A

N/A

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

PR #783, opened 2023-10-06
optLong vs getLong inconsistencies

rudrajyotib:
For exponential decimal conversion, number is not touched. Leading zeros removed from numeric number strings before converting to number.

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -2392,8 +2392,14 @@ public class JSONObject {
      */
     protected static Number stringToNumber(final String input) throws NumberFormatException {
         String val = input;
+        if (val.startsWith(".")){
+            val = "0"+val;
+        }
+        if (val.startsWith("-.")){
+            val = "-0."+val.substring(2);
+        }
         char initial = val.charAt(0);
-        if ((initial >= '0' && initial <= '9') || initial == '-') {
+        if ((initial >= '0' && initial <= '9') || initial == '-' ) {
             // decimal representation
             if (isDecimalNotation(val)) {
                 // Use a BigDecimal all the time so we keep the original
@@ -2424,13 +2430,13 @@ public class JSONObject {
             if(initial == '0' && val.length() > 1) {
                 char at1 = val.charAt(1);
                 if(at1 >= '0' && at1 <= '9') {
-                    throw new NumberFormatException("val ["+val+"] is not a valid number.");
+                    throw new NumberFormatException("val ["+input+"] is not a valid number.");
                 }
             } else if (initial == '-' && val.length() > 2) {
                 char at1 = val.charAt(1);
                 char at2 = val.charAt(2);
                 if(at1 == '0' && at2 >= '0' && at2 <= '9') {
-                    throw new NumberFormatException("val ["+val+"] is not a valid number.");
+                    throw new NumberFormatException("val ["+input+"] is not a valid number.");
                 }
             }
             // integer representation.
@@ -2450,7 +2456,7 @@ public class JSONObject {
             }
             return bi;
         }
-        throw new NumberFormatException("val ["+val+"] is not a valid number.");
+        throw new NumberFormatException("val ["+input+"] is not a valid number.");
     }
 
     /**
@@ -2486,14 +2492,35 @@ public class JSONObject {
          * produced, then the value will just be a string.
          */
 
-        char initial = string.charAt(0);
-        if ((initial >= '0' && initial <= '9') || initial == '-') {
+        if (potentialNumber(string)) {
             try {
                 return stringToNumber(string);
             } catch (Exception ignore) {
             }
         }
         return string;
+    }
+
+
+    private static boolean potentialNumber(String value){
+        if (value == null || value.isEmpty()){
+            return false;
+        }
+        return potentialPositiveNumberStartingAtIndex(value, (value.charAt(0)=='-'?1:0));
+    }
+
+    private static boolean potentialPositiveNumberStartingAtIndex(String value,int index){
+        if (index >= value.length()){
+            return false;
+        }
+        return digitAtIndex(value, (value.charAt(index)=='.'?index+1:index));
+    }
+
+    private static boolean digitAtIndex(String value, int index){
+        if (index >= value.length()){
+            return false;
+        }
+        return value.charAt(index) >= '0' && value.charAt(index) <= '9';
     }
 
     /**
@@ -2902,26 +2929,15 @@ public class JSONObject {
      * @return number without leading zeros
      */
     private static String removeLeadingZerosOfNumber(String value){
-        char[] chars = value.toCharArray();
-        int leftMostUnsignedIndex = 0;
-        if (chars[0] == '-'){
-            leftMostUnsignedIndex = 1;
-        }
-        int firstNonZeroCharIndex = -1;
-        for (int i=leftMostUnsignedIndex;i<value.length();i++){
-            if (chars[i] != '0'){
-                firstNonZeroCharIndex = i;
-                break;
+        if (value.equals("-")){return value;}
+        boolean negativeFirstChar = (value.charAt(0) == '-');
+        int counter = negativeFirstChar ? 1:0;
+        while (counter < value.length()){
+            if (value.charAt(counter) != '0'){
+                return String.format("%s%s", negativeFirstChar?'-':"",value.substring(counter));
             }
+            ++counter;
         }
-        if (firstNonZeroCharIndex == -1){
-            return value;
-        }
-        StringBuilder result = new StringBuilder();
-        if (leftMostUnsignedIndex == 1){
-            result.append('-');
-        }
-        result.append(value.substring(firstNonZeroCharIndex));
-        return result.toString();
+        return String.format("%s%s", negativeFirstChar?'-':"",'0');
     }
 }
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -2379,12 +2379,13 @@ public class JSONObject {
      * returns for this function are BigDecimal, Double, BigInteger, Long, and Integer.
      * When a Double is returned, it should always be a valid Double and not NaN or +-infinity.
      *
-     * @param val value to convert
+     * @param input value to convert
      * @return Number representation of the value.
      * @throws NumberFormatException thrown if the value is not a valid number. A public
      *      caller should catch this and wrap it in a {@link JSONException} if applicable.
      */
-    protected static Number stringToNumber(final String val) throws NumberFormatException {
+    protected static Number stringToNumber(final String input) throws NumberFormatException {
+        String val = input;
         char initial = val.charAt(0);
         if ((initial >= '0' && initial <= '9') || initial == '-') {
             // decimal representation
@@ -2411,6 +2412,8 @@ public class JSONObject {
                     }
                 }
             }
+            val = removeLeadingZerosOfNumber(input);
+            initial = val.charAt(0);
             // block items like 00 01 etc. Java number parsers treat these as Octal.
             if(initial == '0' && val.length() > 1) {
                 char at1 = val.charAt(1);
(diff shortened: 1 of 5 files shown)


stleary:
optLong() and getLong() may produce different results depending on whether leading zeros are present in the value being retrieved.

**Risks**
Low

**Changes to the API?**
Yes. stringToNumber() behavior was modified. This may change how the calling methods behave when the value contains leading zeros. It is not believed that this will be a common occurrence.

No

No

No. New unit tests were added.

Minimal refactoring of existing code was done (parameter name change, etc)

**Review status**
**APPROVED**

src/main/java/org/json/JSONObject.java
```diff
@@ -2886,4 +2889,33 @@ private static JSONException recursivelyDefinedObjectException(String key) {
...
     }
+
+    /**
+     * For a prospective number, remove the leading zeros
+     * @param value prospective number
+     * @return number without leading zeros
+     */
+    private static String removeLeadingZerosOfNumber(String value){
+        char[] chars = value.toCharArray();
+        int leftMostUnsignedIndex = 0;
```
johnjaylward, line 2900: `leftMostUnsignedIndex` could just be a boolean value. I'm not sure using an `int` here are repurposing it in the `for` initializer adds enough performance benefit over the readability of just using a boolean.

src/main/java/org/json/JSONObject.java
```diff
@@ -2886,4 +2889,33 @@ private static JSONException recursivelyDefinedObjectException(String key) {
...
+        for (int i=leftMostUnsignedIndex;i<value.length();i++){
+            if (chars[i] != '0'){
+                firstNonZeroCharIndex = i;
+                break;
+            }
+        }
+        if (firstNonZeroCharIndex == -1){
+            return value;
+        }
+        StringBuilder result = new StringBuilder();
```
johnjaylward, line 2914: The StringBuilder should only be used if `leftMostUnsignedIndex` is `1` (or `true`). Otherwise, just the result of the `substring` can be used.

Really, I'm not sure `StringBuilder` should be used directly at all here. The compiler will usually generate that for us by doing a simple concat. The following 2 samples should output almost identical byte code, however, the first is much easier to read:

```java
if (leftMostUnsignedIndex) {
  return '-' + value.substring(firstNonZeroCharIndex);
} else {
  return value.substring(firstNonZeroCharIndex);
}
```

```java
if (leftMostUnsignedIndex) {
  String val = value.substring(firstNonZeroCharIndex);
  StringBuilder tmp = new StringBuilder(val.length+1)
                             .append('-').append(val);
  return tmp.toString();
} else {
  return value.substring(firstNonZeroCharIndex);
}
```

johnjaylward (changes requested):

stleary:
Merge pending resolution of comments.

src/main/java/org/json/JSONObject.java
johnjaylward: I also am not a fan of the `toCharArray`. I don't think we need to make a copy of the string to do the evaluation. I think a solution more like this is in line with performance and readability.

```java
/**
 * For a prospective number, remove the leading zeros
 * @param value prospective number.
 * @return number without leading zeros
 */
private static String removeLeadingZerosOfNumber(String value){
    if(value == null || value.length() == 0){
        return value;
    }

    boolean leftMostUnsignedIndex = value.charAt(0) == '-';
    int firstNonZeroCharIndex=-1;
    char firstNonZeroChar = 0;

    for(int i=leftMostUnsignedIndex ? 1 : 0; i < value.length(); i++){
    	firstNonZeroChar = value.charAt(i);
        if(firstNonZeroChar != '0') {
        	firstNonZeroCharIndex = i;
            break;
        }
    }

	if(firstNonZeroCharIndex < 0) {
    	return value;
    }
        
    if(firstNonZeroChar == '.' && value.charAt(firstNonZeroCharIndex-1)=='0') {
        firstNonZeroCharIndex--;
    }

    if (leftMostUnsignedIndex) {
        return '-' + value.substring(firstNonZeroCharIndex);
    } else {
        return value.substring(firstNonZeroCharIndex);
    }
}
```

src/test/java/org/json/junit/JSONObjectNumberTest.java
```diff
@@ -23,7 +23,10 @@ public class JSONObjectNumberTest {
     @Parameters(name = "{index}: {0}")
     public static Collection<Object[]> data() {
         return Arrays.asList(new Object[][]{
-            {"{value:50}", 1},
+            {"{value:0050}", 1},
+            {"{value:0050.0000}", 1},
+            {"{value:-0050}", -1},
+            {"{value:-0050.0000}", -1},
```
johnjaylward, line 29: Add a test cases for:
* 0.5 -> expects "0.5"
* -0.5 -> expects "-0.5"
* 00.05 -> expects "0.05"
* -00.05 -> expects "-0.05"

johnjaylward (changes requested):

src/test/java/org/json/junit/JSONObjectNumberTest.java
johnjaylward: also add:

* `.5` -> expects ".5"
* `-.5` -> expects "-.5"

johnjaylward: Also add test case for:
* `00000` -> Expects "0"

rudrajyotib: For decimal cases, the numbers which does not have anything before decimal point, e.g., .5, -.5 etc, what should be the behavior when getLong is called.
As of now, they throw  NumberFormatException. It works for optLong though.
@johnjaylward  @stleary

johnjaylward: I believe in the Issue it was noted that the `opt` parsing was preferred, and I tend to agree.

src/main/java/org/json/JSONObject.java
rudrajyotib: Agreed. toCharArray was unnecessary. This method has been cleaned.
Also, there was another fatal error in initial string parsing to number logic, that prevented any decimal number starting with '.' to be considered as decimal number, that has been fixed as well.

src/main/java/org/json/JSONObject.java
```diff
@@ -2417,17 +2424,19 @@ protected static Number stringToNumber(final String val) throws NumberFormatExce
                     }
                 }
             }
+            val = removeLeadingZerosOfNumber(input);
+            initial = val.charAt(0);
             // block items like 00 01 etc. Java number parsers treat these as Octal.
```
johnjaylward, line 2429: let's remove this incorrect comment about Octal as well.

src/main/java/org/json/JSONObject.java
```diff
@@ -2385,14 +2385,21 @@ protected static boolean isDecimalNotation(final String val) {
...
+        }
+        if (val.startsWith("-.")){
+            val = "-0."+val.substring(2);
+        }
         char initial = val.charAt(0);
-        if ((initial >= '0' && initial <= '9') || initial == '-') {
+        if ((initial >= '0' && initial <= '9') || initial == '-' ) {
             // decimal representation
             if (isDecimalNotation(val)) {
                 // Use a BigDecimal all the time so we keep the original
```
johnjaylward, line 2405: The exception messages in this block need to be updated for the parameter rename to `input` as well.

src/main/java/org/json/JSONObject.java
```diff
@@ -2892,4 +2922,22 @@ private static JSONException recursivelyDefinedObjectException(String key) {
...
+     * @param value prospective number
+     * @return number without leading zeros
+     */
+    private static String removeLeadingZerosOfNumber(String value){
+        if (value.equals("-")){return value;}
+        boolean negativeFirstChar = (value.charAt(0) == '-');
+        int counter = negativeFirstChar ? 1:0;
+        while (counter < value.length()){
+            if (value.charAt(counter) != '0'){
+                return String.format("%s%s", negativeFirstChar?'-':"",value.substring(counter));
```
johnjaylward, line 2937: I'm still not a fan of `String.format` here.

```java
if (value.charAt(counter) != '0') {
   if (negativeFirstChar) return "-"+value.substring(counter);
   return value.substring(counter);
}
++counter;
```

Reads better and doesn't require the extra overhead of `String.format`

johnjaylward (changes requested):

src/main/java/org/json/JSONObject.java
johnjaylward: If you REALLY hate using the `+` operator on strings, my last recommendation would be the following:

```java
while (counter < value.length()){
    if (value.charAt(counter) != '0') {
        if (negativeFirstChar) return "-".concat(value.substring(counter));
        return value.substring(counter);
    }
    ++counter;
}
if (negativeFirstChar) return "-".concat(value.substring(counter));
return value.substring(counter);
```

src/main/java/org/json/JSONObject.java
rudrajyotib: Done. Code is more readable now.

src/main/java/org/json/JSONObject.java
rudrajyotib: Fair enough. I really feel String concatenation using overloaded "+" operator, makes the code much more unreadable.
I do agree, that there is an overhead of String formatting, but it gives the coder a clear picture of the end result.
Concat wins the middle ground, and I have implemented it.

src/main/java/org/json/JSONObject.java
rudrajyotib: Done.

src/main/java/org/json/JSONObject.java
rudrajyotib: Right, now that we are taking care of zeros, there isn't a chance of interpreting as Octal.

rudrajyotib:
@stleary @johnjaylward - can you review the latest commits please.

johnjaylward (approved):
These changes look good now I think. @stleary if we want to maintain similar processing for XML, these changes will need to be copied over to the XML class

stleary:
@rudrajyotib Nice work addressing the comments.

stleary (approved):

parker-mar:
Hello there! I am curious about the new behavior introduced here, where the JSONObject constructor trims 0-led integers and converts them to int type instead of String type. 

While I can understand the rationale behind trimming, the change in behavior could be problematic for things like IDs with leading 0s, which may have previously relied on the JSONObject constructor preserving the value as String type, only to have it converted to int, and thus failing a subsequent call to JSONObject [`getString("id")`](https://www.javadoc.io/doc/org.json/json/latest/org/json/JSONObject.html#getString-java.lang.String-) because of type mismatch.

Comparing with other Java JSON libraries, org.json seems to behave in a way that's potentially unsafe as it transforms the input and therefore "hides" the error, versus the other libraries which take a safer approach by throwing an error on input (javax.json, Jackson) or treating the input as "pseudo"-json and letting the getter decide the type (Gson).

```Java
import com.fasterxml.jackson.core.JsonProcessingException;
import com.fasterxml.jackson.databind.JsonNode;
import com.fasterxml.jackson.databind.ObjectMapper;
import com.google.gson.JsonParser;
import org.json.JSONObject;
import org.junit.jupiter.api.Test;

import javax.json.Json;
import java.io.StringReader;

class MyTest {
    @Test
    void myTest() throws JsonProcessingException {
        String personId = "0123";

        // org.json transforms on construction 
        JSONObject j1 = new JSONObject("{personId: " + personId + "}");
        System.out.println(j1.getString("personId")); // Throws exception

        // javax.json throws error on construction
        javax.json.JsonObject j2 = Json.createReader(new StringReader("{\"personId\": " + personId + "}")).readObject(); 
        System.out.println(j2.getString("personId"));

        // gson treats it as pseudo-json and leaves it to the getter to decide: getAsInt will trim, getAsString will not
        com.google.gson.JsonObject j3 = JsonParser.parseString("{\"personId\": " + personId + "}").getAsJsonObject();
        System.out.println(j3.get("personId").getAsString());

        // jackson throws error on construction
        JsonNode jsonNode = new ObjectMapper().readTree("{\"personId\": " + personId + "}");
        System.out.println(jsonNode.get("personId").asText());
    }
}
```

PR #784, opened 2023-10-07
Fixed get() and opt() inconsistency

### LINKED ISSUE #653: optLong vs getLong inconsitencies

nayak-yash:
## What does this PR do ? 
This PR fixes the optLong() - getLong() and optDouble() and getDouble() inconsistencies
Fixes #653 

## Type of change
Previously `stringToNumber()` checked for strings starting with either digit or - . Now it only checks for whether the string is a decimal number or integer.

## Other information
```
JSONObject object1 = new JSONObject("{\"number\":\"-.1234\"}");
JSONObject object2 = new JSONObject("{\"number\":\"-0000000.32879\"}");
JSONObject object3 = new JSONObject("{\"number\":\"-07.89\"}");
JSONObject object4 = new JSONObject("{\"number\":\"-0.00000000\"}");
System.out.println("1. getDouble() = " + object1.getDouble("number"));
System.out.println("1. optDouble() = " + object1.optDouble("number"));
System.out.println("2. getDouble() = " + object2.getDouble("number"));
System.out.println("2. optDouble() = " + object2.optDouble("number"));
System.out.println("3. getDouble() = " + object3.getDouble("number"));
System.out.println("3. optDouble() = " + object3.optDouble("number"));
System.out.println("4. getDouble() = " + object4.getDouble("number"));
System.out.println("4. optDouble() = " + object4.optDouble("number"));
 ```
Output : 
```
1. getDouble() = -0.1234
1. optDouble() = -0.1234
2. getDouble() = -0.32879
2. optDouble() = -0.32879
3. getDouble() = -7.89
3. optDouble() = -7.89
4. getDouble() = -0.0
4. optDouble() = -0.0
```

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -2385,63 +2385,47 @@ public class JSONObject {
      *      caller should catch this and wrap it in a {@link JSONException} if applicable.
      */
     protected static Number stringToNumber(final String val) throws NumberFormatException {
-        char initial = val.charAt(0);
-        if ((initial >= '0' && initial <= '9') || initial == '-') {
-            // decimal representation
-            if (isDecimalNotation(val)) {
-                // Use a BigDecimal all the time so we keep the original
-                // representation. BigDecimal doesn't support -0.0, ensure we
-                // keep that by forcing a decimal.
+        // decimal representation
+        if (isDecimalNotation(val)) {
+            // Use a BigDecimal all the time so we keep the original
+            // representation. BigDecimal doesn't support -0.0, ensure we
+            // keep that by forcing a decimal.
+            char initial = val.charAt(0);
+            try {
+                BigDecimal bd = new BigDecimal(val);
+                if(initial == '-' && BigDecimal.ZERO.compareTo(bd)==0) {
+                    return Double.valueOf(-0.0);
+                }
+                return bd;
+            } catch (NumberFormatException retryAsDouble) {
+                // this is to support "Hex Floats" like this: 0x1.0P-1074
                 try {
-                    BigDecimal bd = new BigDecimal(val);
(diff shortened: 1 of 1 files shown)


nayak-yash:
Can you please add hacktober-accepted label or tell me if this doesn't solve your issue. I would look into it.

stleary:
@nayak-yash Thanks for submitting this PR, but #783 has already been accepted to fix this bug. Your fix also successfully addressed the bug, however, the other solution was preferred because the code changes were directly related to the bug fix, and numerous unit tests were added. Please feel free to pick another issue and work on that.

PR #788, opened 2023-10-11
Fix string compare unit tests

hofi1:
This PR fixes two flaky tests:
- org.json.junit.JSONObjectTest#valueToString
- org.json.junit.JSONMLTest#testToJSONObject_reversibility

The flaky test have been found by using the [NonDex](https://mvnrepository.com/artifact/edu.illinois/nondex-maven-plugin) tool.

# org.json.junit.JSONObjectTest#valueToString

## Problem:

The string in the assertion can change, because the Map which is used to store the data in the JSONObject returns the data in non-deterministic order (the order of child elements in JSON strings do not matter, and JSON strings are equal regardless of the ordering of the elements on the same hierarchy level)

The flakiness was discovered on the following lines
https://github.com/stleary/JSON-java/blob/c4cd526c53e99f20851546fca92368738f783884/src/test/java/org/json/junit/JSONObjectTest.java#L2028-L2029

as well as 
https://github.com/stleary/JSON-java/blob/c4cd526c53e99f20851546fca92368738f783884/src/test/java/org/json/junit/JSONObjectTest.java#L2039-L2040

## Solution:
Changed the assert statement from a JUnit Assertion to a JSON Assertion, so the strings are not compared charwise but are compared the way as specified for JSON strings (not taking care of the order of the elements on the same level).
The JSONAssertion library was chosen because it provides general assertions to compare JSON strings without the need of writing complex JUnit Assertions, what would need a lot of work to convert the strings in a different data structure, a lot of boilerplate code as well as the chance of adding more errors than fixing. 

## Result:

The test is deterministic and not flaky. This improves the quality of the test and reduces the time to search for the bug during future development.

## Reproduce:
```shell
mvn edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=org.json.junit.JSONObjectTest#valueToString
``` 

# org.json.junit.JSONMLTest#testToJSONObject_reversibility

## Problem:

The string in the assertion can change, because the Map which is used to store the data in the JSONObject returns the data in non-deterministic order.

## Solution:
Change the assertion is to a JSONAssertion (e.g. the order of child elements in an element do not matter)

## Result:

The test is deterministic and not flaky. This improves the quality of the test and reduces the time to search for the bug during future development.

## Reproduce:
```shell
mvn edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=org.json.junit.JSONMLTest#testToJSONObject_reversibility
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,6 @@ dependencies {
     testImplementation 'junit:junit:4.13.1'
     testImplementation 'com.jayway.jsonpath:json-path:2.1.0'
     testImplementation 'org.mockito:mockito-core:4.2.0'
-    testImplementation 'org.skyscreamer:jsonassert:1.5.1'
 }
 
 subprojects {
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,7 @@ dependencies {
     testImplementation 'junit:junit:4.13.1'
     testImplementation 'com.jayway.jsonpath:json-path:2.1.0'
     testImplementation 'org.mockito:mockito-core:4.2.0'
+    testImplementation 'org.skyscreamer:jsonassert:1.5.1'
 }
 
 subprojects {
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,7 @@ repositories {
 dependencies {
     testImplementation 'junit:junit:4.13.1'
     testImplementation 'com.jayway.jsonpath:json-path:2.1.0'
-    testImplementation 'org.mockito:mockito-core:1.9.5'
+    testImplementation 'org.mockito:mockito-core:4.2.0'
 }
 
 subprojects {
@@ -30,9 +30,9 @@ subprojects {
 }
 
 group = 'org.json'
-version = 'v20211205-SNAPSHOT'
+version = 'v20230618-SNAPSHOT'
 description = 'JSON in Java'
-sourceCompatibility = '1.7'
+sourceCompatibility = '1.8'
 
 configurations.all {
 }
--- a/pom.xml
+++ b/pom.xml
@@ -72,6 +72,13 @@
             <version>4.2.0</version>
             <scope>test</scope>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/org.skyscreamer/jsonassert -->
+        <dependency>
+            <groupId>org.skyscreamer</groupId>
+            <artifactId>jsonassert</artifactId>
+            <version>1.5.1</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/pom.xml
+++ b/pom.xml
@@ -72,13 +72,6 @@
             <version>4.2.0</version>
             <scope>test</scope>
         </dependency>
-        <!-- https://mvnrepository.com/artifact/org.skyscreamer/jsonassert -->
-        <dependency>
-            <groupId>org.skyscreamer</groupId>
-            <artifactId>jsonassert</artifactId>
-            <version>1.5.1</version>
-            <scope>test</scope>
-        </dependency>
     </dependencies>
 
     <build>
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
-            <version>1.9.5</version>
+            <version>4.2.0</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
@@ -79,7 +79,7 @@
             <plugin>
                 <groupId>org.apache.felix</groupId>
                 <artifactId>maven-bundle-plugin</artifactId>
-                <version>3.0.1</version>
+                <version>5.1.9</version>
                 <extensions>true</extensions>
                 <configuration>
                     <instructions>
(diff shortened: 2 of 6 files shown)


stleary:
@hofi1 Thanks for the PR. The thorough documentation is appreciated. 
Agreed there is an issue, but are you sure that JSONAssert is the best way to address it?

hofi1:
@stleary I appreciate the fast feedback!
I totally understand that it is not seen with pleasure if someone wants to introduce a new dependency to a project. 

For the org.json.junit.JSONObjectTest#valueToString test case:
Since the reference is stored as a String, I don't see any other way of fixing it. If it would be stored in a map or a similar data structure, it would be possible to use a normal (but really not beautiful) JUnit Assertion. This would need to cover 3! different cases (equals 6 cases, namely all permutations for all different key-value pairs), to check for all different orderings. If one would add more elements to the initial JSON string, this would result in a lot of work to add all possible different orderings of the string. 
If these different options would not be added, this would result in a test which is flaky again. This means it is not really a solution but rather a future vulnerability of the code (and definitely a code smell).  

Regarding the org.json.junit.JSONMLTest#testToJSONObject_reversibility test case:
The goal of this test is, to compare two strings and check for the reversibility. Taking a look at the complex structure of this JSON example (see attached screenshot), I don't think it that there is really a different way than using JSONAssertions - a hierarchy of 6 levels plus arrays. Building own matchers would create a whole amount of boilerplate code, and I think it would require adding the hamcrest library to build these – and this is not really necessary if there are already JSON matchers available. 
You could also think about adding a special case for each permutation, but this would grow dramatically due to the size of this JSON string (compared to the only 3 values resulting in 6 permutations in the case above) and would still leave the project with the same drawbacks as explained above. 

Screenshot

stleary:
@hofi1 JSONAssert is just parsing the strings to JSON using the less-capable Android port and then doing a key-by-key comparison. If that is the goal, then wouldn't it be easier, more reliable, and a better test to parse with JSON-Java, and then use the `JSONObject.similar()` and `JSONArray.similar()` API methods?
@johnjaylward I think you wrote or updated these tests, let us know if you disagree.

johnjaylward:
I haven't had time to review this yet, but can't we use our existing 'similar' function?

hofi1:
@stleary I see your point. Thanks for pointing that out. 
Regarding the testToJSONObject_reversibility test, this would be possible. I thought the goal of this test might be to check for equality of the JSON strings - my bad. 

Regarding the valueToString test, I don't see any other way. Because the goal of this test is to compare the strings, and to complete this, a string comparison is needed. It would be possible to do this by adding each different case, but this is definitely not best practice.

stleary:
@hofi1 I think you are correct. Ideally, the actual string results need to be compared, but so far as I can tell, there is no good way to do this.

hofi1:
@stleary I changed the assertion in JSONMLTest#testToJSONObject_reversibility to a comparison of the JSONObjects

Mayanksundesha (approved):

WizOzHeir (approved):

build.gradle
```diff
@@ -23,6 +23,7 @@ dependencies {
     testImplementation 'junit:junit:4.13.1'
     testImplementation 'com.jayway.jsonpath:json-path:2.1.0'
     testImplementation 'org.mockito:mockito-core:4.2.0'
+    testImplementation 'org.skyscreamer:jsonassert:1.5.1'
```
stleary, line 26: Please remove

pom.xml
```diff
@@ -72,6 +72,13 @@
             <version>4.2.0</version>
             <scope>test</scope>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/org.skyscreamer/jsonassert -->
```
stleary, line 75: Please remove this section

src/test/java/org/json/junit/JSONObjectTest.java
```diff
@@ -57,6 +57,7 @@
 
 import com.jayway.jsonpath.Configuration;
 import com.jayway.jsonpath.JsonPath;
+import org.skyscreamer.jsonassert.JSONAssert;
```
stleary, line 60: Please remove

src/test/java/org/json/junit/JSONObjectTest.java
```diff
@@ -2025,8 +2026,8 @@ public void valueToString() {
                 "\"key3\":\"val3\""+
              "}";
         JSONObject jsonObject = new JSONObject(jsonObjectStr);
-        assertTrue("jsonObject valueToString() incorrect",
-                JSONObject.valueToString(jsonObject).equals(jsonObject.toString()));
+        JSONAssert.assertEquals("jsonObject valueToString() incorrect",
```
stleary, line 2029: Please replace this line with code that converts the strings to JSONObjects and compare them using JSONObject.similar()

src/test/java/org/json/junit/JSONMLTest.java
```diff
@@ -762,8 +762,7 @@ public void testToJSONObject_reversibility() {
         final String xml = JSONML.toString(originalObject);
         final JSONObject revertedObject = JSONML.toJSONObject(xml, false);
         final String newJson = revertedObject.toString();
-        assertTrue("JSON Objects are not similar",originalObject.similar(revertedObject));
-        assertEquals("original JSON does not equal the new JSON",originalJson, newJson);
```
stleary, line 766: Instead of deleting line 766, replace it with code that converts the strings to JSONObjects, and compares them using JSONObject.similar().

src/test/java/org/json/junit/JSONObjectTest.java
```diff
@@ -2036,8 +2037,8 @@ public void valueToString() {
         map.put("key1", "val1");
         map.put("key2", "val2");
         map.put("key3", "val3");
-        assertTrue("map valueToString() incorrect",
-                jsonObject.toString().equals(JSONObject.valueToString(map))); 
+        JSONAssert.assertEquals("map valueToString() incorrect",
```
stleary, line 2040: Please replace this line with code that converts the strings to JSONObjects and compare them using JSONObject.similar()

stleary:
[Nondex](https://mvnrepository.com/artifact/edu.illinois/nondex-maven-plugin) has found tests that might fail due to assumptions about JSONObject ordering. Since a direct string comparison is not reliable, the strings are converted to JSONObject and compared using similar(). This is an imperfect solution, but the best available. 

**Risks**
Low

**Changes to the API?**
N/A

No

No

No, the change is limited to the unit tests

No

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

PR #804, opened 2023-10-16
Create (JSONArray).toList() breaks format of entries

Priyanshu-ai902:


stleary:
No, thank you. Needs to be coded in Java. Also, we don't want to change the existing API. Instead, add a new API method with a different name, with the desired functionality.

PR #801, opened 2023-10-16
Updates the pipeline to validate that packaging a jar works properly

johnjaylward:
Fixes #796 

See also #793, #761 and #760

**What problem does this code solve?**

Re-adds proper support for Modules on Java 9+. PR #761 was not tested with builds using JDK 8 and the changes only worked when using JDK 9+.

* Validate that the `mvn package` step completes during the pipeline
* fix issue with package step failing when run on jvm8
* Update the README.md to have correct supported versions
* Adds package artifacts to Action results. see the [Actions -> Artifacs](https://github.com/stleary/JSON-java/actions/runs/6538228155) section for compiled JARs

A Follow-on PR that enables [distribution through GitHub Actions](https://docs.github.com/en/actions/publishing-packages/publishing-java-packages-with-maven) would be worthwhile.

**Risks**

Low. Changes add new steps to the GitHub actions to ensure that packaging works properly going forward and will only affect the JAR creation

**Changes to the API?**

None

**Will this require a new release?**

Yes

**Should the documentation be updated?**

No, but slight changes were made to the README to:
1. Include workflow status badges
2. Correct the supported Java versions

**Does it break the unit tests?**

No. New GitHub actions were added to ensure that similar changes will not break release management going forward.

**Was any code refactored in this commit?**

No.

**Review status**

**APPROVED**

Starting 3-day comment window

PR #806, opened 2023-10-16
Add new deployment pipeline

johnjaylward:

Adds a new pipeline that should allow automatic deployment to Maven Central when a new GitHub Release is published. This reduces the manual work involved in creating and publishing releases. If the project needs to be handed off again, this should also reduce the hand-off steps.

The pipeline was based on [GitHub documentation](https://github.com/actions/setup-java/blob/v3.13.0/docs/advanced-usage.md#Publishing-using-Apache-Maven) and [StackOverflow how to use GPG in Workflows](https://stackoverflow.com/questions/61096521/how-to-use-gpg-key-in-github-actions) and modified slightly for our use-case as defined in the [Wiki Release document](https://github.com/stleary/JSON-java/wiki/Tech:-How-to-make-a-new-JSON-java-release) and [Sonatype release documentation](https://central.sonatype.org/publish/publish-guide/).  In order for the pipeline to run successfully, The following [Secrets](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions) will need to be added to the workflow configuration in the GitHub Project. Only the project maintainer can do this configuration.

Secrets names:
* `OSSRH_USERNAME` set to your Sonatype Jira username
* `OSSRH_TOKEN` set to your Sonatype Jira password/token
* `MAVEN_GPG_PRIVATE_KEY` set to your GPG private key: `gpg --list-secret-keys --keyid-format LONG` to get the Key ID, then `gpg --export-secret-keys --armor {your_keyId}` to get the actual key that goes in the secret
* `MAVEN_GPG_PASSPHRASE` set to the passphrase needed to use your private key

**Risks**

Low. If the auto-deployment fails, manual deployment can still be done. Corrections could be made on-the-fly, or could be queued up for the next release.

**Changes to the API?**

No

No

It would be a good idea.

No

No

**Review status**

**APPROVED**

Starting 3-day comment window

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -1,4 +1,8 @@
-# For more information see: https://docs.github.com/en/actions/learn-github-actions or https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions
+# For more information see:
+#  * https://docs.github.com/en/actions/learn-github-actions
+#  * https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions
+#  * https://github.com/actions/setup-java/blob/v3.13.0/docs/advanced-usage.md#Publishing-using-Apache-Maven
+#
 
 name: Deployment workflow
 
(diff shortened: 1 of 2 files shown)


johnjaylward:
Here's an example of a failed deployment because I did not set up the secrets:
https://github.com/johnjaylward/JSON-java/actions/runs/6539967284

johnjaylward:
Related to #801

stleary (approved):

PR #815, opened 2023-10-23
Ensure java 6 compatable

johnjaylward:

Adds back in the build/jar for java6 to help ensure binary source compatibility. This leaves the actual deployment as a Java 8 compatibility.

Sample build run: https://github.com/johnjaylward/JSON-java/actions/runs/6617867256

As part of the java6 build, I made sure to output the current version of the java compiler:

image

Yes, that's what this PR is made to ensure.

**Risks**

To compile Java 1.6 on GitHub, it requires using the deprecated action "setup-java@v1". `v2` and greater of the action do not support java6. It's unknown when GitHub will end support for this action completely. Once then do, builds will break and we will need to "cross compile" from a newer version of the JDK down to 1.6, which is not guaranteed to ensure full 1.6 compatibility.

**Changes to the API?**

No

No

No

No. However, no testing is done against JDK 6. Only compilation and packing is performed. The JDK jar that is uploaded to the build artifacts is also not compatible with JDK 9+ as no module information is provided

No

**Review status**

NOT YET REVIEWED

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -10,6 +10,34 @@ on:
     branches: [ master ]
 
 jobs:
+  # old-school build and jar method. No tests run or compiled.
+  build-1_6: 
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # build for java 1.6, however don't run any tests
+        java: [ 1.6 ]
+    name: Java ${{ matrix.java }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup java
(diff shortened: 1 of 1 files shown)


stleary:
@johnjaylward Looks good, thanks!

stleary (approved):

johnjaylward:
you removed the approved label but didn't merge. Was that intentional?

stleary:
@johnjaylward  My bad, it was still in the 3-day window. Going ahead and merging now.

PR #820, opened 2023-10-25
Close XML tag explicitly for empty tags with configuration.

rudrajyotib:
Added configuration option - closeEmptyTag
When this flag is set to true, all the empty tags will be closed using explicit start and end tag.

### DIFF
--- a/src/main/java/org/json/XMLParserConfiguration.java
+++ b/src/main/java/org/json/XMLParserConfiguration.java
@@ -149,6 +149,7 @@ public class XMLParserConfiguration extends ParserConfiguration {
      *                   xsi:type="integer" as integer,  xsi:type="string" as string
      * @param forceList  <code>new HashSet<String>()</code> to parse the provided tags' values as arrays
      * @param maxNestingDepth <code>int</code> to limit the nesting depth
+     * @param closeEmptyTag <code>boolean</code> to turn on explicit end tag for tag with empty value
      */
     private XMLParserConfiguration (final boolean keepStrings, final String cDataTagName,
             final boolean convertNilAttributeToNull, final Map<String, XMLXsiTypeConverter<?>> xsiTypeMap, final Set<String> forceList,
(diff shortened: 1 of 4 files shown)


rudrajyotib:
@stleary - mvn:test is failing for java 11 execution, which seems to be a one-off and temporary problem. Can the CI steps be executed once more please. I have run the tests locally with Java versions 8, 11, 17 and they do not fail. I think another run should mark it clear, if not, I will look into it.

src/main/java/org/json/XMLParserConfiguration.java
```diff
@@ -303,4 +313,18 @@ public XMLParserConfiguration withForceList(final Set<String> forceList) {
...
         return super.withMaxNestingDepth(maxNestingDepth);
     }
+
+    /**
+     * To enable explicit end tag with empty value.
+     * @param closeEmptyTag
+     * @return same instance of configuration with empty tag config updated
+     */
+    public XMLParserConfiguration withCloseEmptyTag(boolean closeEmptyTag){
+        this.closeEmptyTag = closeEmptyTag;
```
stleary, line 323: Please follow the model of the other "with*" methods:
* Clone 'this' object
* Update the clone with the new property value
* Return the clone

stleary:
@rudrajyotib Looks good, just one comment to address.

src/main/java/org/json/XMLParserConfiguration.java
rudrajyotib: @stleary -updated code and test cases accordingly.

rudrajyotib:
@stleary  - updated configuration builder and added test cases.

stleary:
Fixes a request made to support optional alternate XML close tags

Yes

**Risks**
Low

**Changes to the API?**
Adds a new XMLParserConfiguration property

No

No

No, new unit tests were added

No

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

PR #824, opened 2023-11-06
Ignore tests that fail due to resource differences

### LINKED ISSUE #821: Flaky test fails intermittently in GitHub Action
Test case passes on some runs and fails on others: `org.json.junit.JSONObjectTest.issue654StackOverflowInputWellFormed`

- Test Run: https://github.com/stleary/JSON-java/actions/runs/6644866305/job/18054932906#step:5:79
- Code Link: https://github.com/stleary/JSON-java/blob/b5b9f636ff150fbad9dce3cd3e77a2747ac73190/src/test/java/org/json/junit/JSONObjectTest.java#L3668-L3676

Appears to be due to differences in resources available on each run. The exception expected is a "JSONException", but that's because it is wrapping a "StackOverflowError". StackOverflowError is going to be dependent on the environment and may not hit consistently.

johnjaylward:

Fixes #821

Add ignore annotation to tests that may fail due to differences in machine resources and can't be controlled via the test framework

Yes

**Risks**

Small. If a change is made to the parser, it's possible that the test case that this is trying to solve for may break.

**Changes to the API?**

No

No

No, it disables 2 tests that were expecting Stack Overflow errors, but would sometimes not trigger due to systems having differing resources available. The 2 tests were left intact, just marked with the `@Ignore` annotation.

No.

**Review status**

**APPROVED**

### DIFF
--- a/src/test/java/org/json/junit/JSONArrayTest.java
+++ b/src/test/java/org/json/junit/JSONArrayTest.java
@@ -32,6 +32,7 @@ import org.json.JSONPointerException;
 import org.json.JSONString;
 import org.json.JSONTokener;
 import org.json.junit.data.MyJsonString;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import com.jayway.jsonpath.Configuration;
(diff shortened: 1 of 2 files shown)


stleary:
Starting 3-day comment window

stleary (approved):

PR #831, opened 2023-11-19
Refactor NumberConversionUtil and toString() of CookieList & XML Classes. 

adityap27:
Hello @stleary,

I have made 3 refactorings.

**1. Renamed Variable boolean b in toString() method of CookieList Class.**
boolean `isEndOfPair` is more meaningful as compared to just boolean `b`.

**2. Introduced indentationSuffix variable in toString() method of XML Class.**
There are multiple long expressions which are using `(indentFactor > 0) ? "\n" : ""` sub-expression. `indentationSuffix` is introduced  to replace this sub-expression to improve readability and reduce duplication.

**3. Added method isNumericChar(...) in NumberConversionUtil Class.**
There are some digit-related conditions which may look complex to some developers due to multiple conditional operators. 
Eg: `if ((initial >= '0' && initial <= '9') || initial == '-' )`
and `if(at1 == '0' && at2 >= '0' && at2 <= '9')`

The sub-condition `c >= '0' && c <= '9'` is moved to a method `isNumericChar(...)` to reduce the complexity of these conditions and better readability.

**Notes:**
All the unit tests are passing successfully.
image

This code is compatible with Java 6.

### DIFF
--- a/src/main/java/org/json/NumberConversionUtil.java
+++ b/src/main/java/org/json/NumberConversionUtil.java
@@ -24,7 +24,7 @@ class NumberConversionUtil {
             val = "-0."+val.substring(2);
         }
         char initial = val.charAt(0);
-        if ((initial >= '0' && initial <= '9') || initial == '-' ) {
+        if ( isNumericChar(initial) || initial == '-' ) {
             // decimal representation
             if (isDecimalNotation(val)) {
                 // Use a BigDecimal all the time so we keep the original
@@ -53,13 +53,13 @@ class NumberConversionUtil {
             initial = val.charAt(0);
             if(initial == '0' && val.length() > 1) {
                 char at1 = val.charAt(1);
-                if(at1 >= '0' && at1 <= '9') {
+                if(isNumericChar(at1)) {
                     throw new NumberFormatException("val ["+input+"] is not a valid number.");
                 }
             } else if (initial == '-' && val.length() > 2) {
                 char at1 = val.charAt(1);
                 char at2 = val.charAt(2);
-                if(at1 == '0' && at2 >= '0' && at2 <= '9') {
+                if(at1 == '0' && isNumericChar(at2)) {
                     throw new NumberFormatException("val ["+input+"] is not a valid number.");
                 }
             }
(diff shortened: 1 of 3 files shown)


adityap27:
Hello @stleary and @johnjaylward,

Can this be reviewed and merged please?

adityap27:
Thanks for starting the review process @stleary.

I would like to ask if there is any possibility of merging this before the **26 Nov day end**?
Actually, this is a part of my university course assignment, where I would be getting a few bonus points if this gets merged on time. Sorry, if this gentle request sounds casual. I really understand and respect your time.

I am happy to modify the changes as per your feedback.

stleary:
@adityap27 Out of sync, please merge master to your branch.

adityap27:
@stleary - done, now in sync.

stleary:
Refactoring to improve code readability

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No

Yes, see first prompt

**Review status**
**APPROVED**

Starting 3-day comment window

src/main/java/org/json/NumberConversionUtil.java
```diff
@@ -83,6 +83,16 @@ static Number stringToNumber(final String input) throws NumberFormatException {
...
     }
 
+    /**
+     * Checks if the character is a numeric digit ('0' to '9').
+     *
+     * @param c The character to be checked.
+     * @return true if the character is a numeric digit, false otherwise.
+     */
+    private static boolean isNumericChar(char c) {
+        return (c >= '0' && c <= '9');
```
johnjaylward, line 93: probably a stupid optimization, but switching this comparison to be:

```java
return (c <= '9' && c >= '0');
```

should provide slightly better performance. My assumption here is that in both ASCII and Unicode, there are a lot more code points that are `>'9'` than there are `<'0'`. That means that for most inputs we can fail-fast with a single comparison instead of having to do both compares.

johnjaylward (changes requested):

src/main/java/org/json/NumberConversionUtil.java
stleary: @adityap27 Please make this change, it won't affect the comment window duration.

adityap27: @stleary and @johnjaylward - done, updated.

johnjaylward (approved):

stleary (approved):

Issue #695, opened 2022-10-11
XML.toJSONObject - xml content trimmed


ilmagowalter:
when create a JSONObject from XML String like 

`<?xml version="1.0" encoding="utf-8"?><tagA>  :  </tagA>`

spaces at then begin and at the end of string are trimmed

```
	public static void main(String[] args) {
		String xml = "<?xml version=\"1.0\" encoding=\"utf-8\"?><tagA>  :</tagA>";
		JSONObject jo = XML.toJSONObject(xml, true);
		System.out.println("Test 1");
		System.out.println(jo.toString());

		xml = "<?xml version=\"1.0\" encoding=\"utf-8\"?><tagA>  :  </tagA>";
		jo = XML.toJSONObject(xml, true);
		System.out.println("Test 2");
		System.out.println(jo.toString());

		xml = "<?xml version=\"1.0\" encoding=\"utf-8\"?><tagA>:  </tagA>";
		jo = XML.toJSONObject(xml, true);
		System.out.println("Test 3");
		System.out.println(jo.toString());
	}
```

output
```
Test 1
{"tagA":":"}
Test 2
{"tagA":":"}
Test 3
{"tagA":":"}

```

is possible add a parameter to avoid trimming ?

i think that method involved is nextContent() of XMLTokener.java

iashok22:
@ilmagowalter , is there any reason why do we need to retain trailing/beginning spaces?

ilmagowalter:
i'm facing this use case

i receive a xml file with field like 
`<tagA>  :  </tagA>
`
this tag is defined in xsd schema like

```
                              <xs:element name="tagA" minOccurs="0"> 
                                 <xs:annotation> 
                                    <xs:appinfo> 
									<RicSDO:exampleValues> 
                                          <RicSDO:example value="18:29"/> 
                                       </RicSDO:exampleValues> 
                                    </xs:appinfo> 
                                 </xs:annotation> 
                                 <xs:simpleType> 
                                    <xs:restriction base="xs:string"> 
                                       <xs:length value="5"/> 
                                    </xs:restriction> 
                                 </xs:simpleType> 
                              </xs:element> 

```
so the parsed is succesful; i work with this file, trasform to json and store on nosql database. 
One of possibilities is export data to xml again; so...if i store in json 
`{"tagA": ":" }
`i will export 
`<tagA>:</tagA>
`
and with this element the parse with xsd fail; i have to store 
`{"tagA": "  :  " }
`

[...]

PR #832, opened 2023-11-20
Add a config flag to disable whitespace trimming

keatontaylor10:
Related issue: https://github.com/stleary/JSON-java/issues/695#issuecomment-1815041209

Notes about implementation:
- An XMLParserConfiguration object with a boolean `shouldTrimWhitespace` has been added and can be passed into the toJSONObject method to disable whitespace trimming. This is as discussed in the issue.
- Inputs with tags that are the same as the cDataTagName variable ("content" by default) are unsupported with whitespace trimming disabled. This has been documented in the javadoc comments for the new methods.
- Tests have been added to verify the new behaviour as well as the old. A test also documents the effects of using the unsupported usage.


src/main/java/org/json/XMLTokener.java
```diff
@@ -83,7 +94,7 @@ public Object nextContent() throws JSONException {
         StringBuilder sb;
         do {
             c = next();
-        } while (Character.isWhitespace(c));
+        } while (Character.isWhitespace(c) && configuration.shouldTrimWhiteSpace());
```
stleary, line 97: note to self: does this change existing behavior?

src/main/java/org/json/XMLParserConfiguration.java
```diff
@@ -61,16 +61,19 @@ public class XMLParserConfiguration extends ParserConfiguration {
...
-     * values), and the CDATA Tag Name is "content".
+     * values), and the CDATA Tag Name is "content". Trims whitespace.
      */
     public XMLParserConfiguration () {
         super();
         this.cDataTagName = "content";
         this.convertNilAttributeToNull = false;
         this.xsiTypeMap = Collections.emptyMap();
         this.forceList = Collections.emptySet();
+        this.shouldTrimWhiteSpace = true;
```
stleary, line 76: note to self: a config param that initializes to true makes it hard to think about default vs override behavior.

stleary:
@keatontaylor10 Thanks for the PR and the careful work. Code review is in progress, but have patience due to the holiday.

src/main/java/org/json/XMLParserConfiguration.java
```diff
@@ -153,13 +156,14 @@ public XMLParserConfiguration (final boolean keepStrings, final String cDataTagN
      */
     private XMLParserConfiguration (final boolean keepStrings, final String cDataTagName,
             final boolean convertNilAttributeToNull, final Map<String, XMLXsiTypeConverter<?>> xsiTypeMap, final Set<String> forceList,
-            final int maxNestingDepth, final boolean closeEmptyTag) {
+            final int maxNestingDepth, final boolean closeEmptyTag, final boolean shouldTrimWhiteSpace) {
```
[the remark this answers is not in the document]
keatontaylor10: Thanks for the feedback, this was done to ensure the clone() methods functions as required and the public constructor was left untouched. I have pushed a new commit to adjust this so that the constructor is untouched and the parameter is set in the clone method.

stleary:
@keatontaylor10 Out of sync, please merge master to your branch.

image

keatontaylor10:
@stleary I have merged and pushed to my branch

src/main/java/org/json/XMLTokener.java
stleary: It does not, since shouldTrimWhiteSpace() is true by default.

src/main/java/org/json/XMLTokener.java
```diff
@@ -20,6 +20,8 @@ public class XMLTokener extends JSONTokener {
     */
    public static final java.util.HashMap<String, Character> entity;
 
+   private static XMLParserConfiguration configuration = XMLParserConfiguration.ORIGINAL;;
```
stleary, line 23: If configuration must be static, then add a comment explaining why. Otherwise, it probably should not be static.

src/main/java/org/json/XML.java
```diff
@@ -445,6 +448,48 @@ private static boolean parse(XMLTokener x, JSONObject context, String name, XMLP
             }
         }
     }
+    /**
```
stleary, line 451: What is the reasoning for trimming whitespace from CDATA? Is it a requirement for your project, or is there a more general reason?

src/main/java/org/json/XML.java
```diff
@@ -445,6 +448,48 @@ private static boolean parse(XMLTokener x, JSONObject context, String name, XMLP
...
+     * This method removes any JSON entry which has the key set by XMLParserConfiguration.cDataTagName
+     * and contains whitespace as this is caused by whitespace between tags. See test XMLTest.testNestedWithWhitespaceTrimmingDisabled.
+     * @param jsonObject JSONObject which may require deletion
+     * @param config The XMLParserConfiguration which includes the cDataTagName
+     */
+    private static void removeEmpty(final JSONObject jsonObject, final XMLParserConfiguration config) {
+        if (jsonObject.has(config.getcDataTagName()))  {
+            final Object s = jsonObject.get(config.getcDataTagName());
+            if (s instanceof String) {
```
stleary, line 460: Isn't CDATA content always a string?

src/main/java/org/json/XML.java
```diff
@@ -445,6 +448,48 @@ private static boolean parse(XMLTokener x, JSONObject context, String name, XMLP
...
+                    final Object eachString = sArray.get(k);
+                    if (eachString instanceof String) {
+                        String s1 = (String) eachString;
+                        if (isStringAllWhiteSpace(s1)) {
+                            sArray.remove(k);
+                        }
+                    }
+                }
+                if (sArray.isEmpty()) {
+                    jsonObject.remove(config.getcDataTagName());
```
stleary, line 477: Is there a reason for removing the key?

src/main/java/org/json/XMLParserConfiguration.java
```diff
@@ -61,16 +61,19 @@ public class XMLParserConfiguration extends ParserConfiguration {
      */
     private Set<String> forceList;
 
+    private boolean shouldTrimWhiteSpace;
```
stleary, line 64: Please include a comment block that describes the purpose of this flag, and all of the ways it changes the behavior of the library.

src/main/java/org/json/XMLTokener.java
keatontaylor10: Will update this

src/main/java/org/json/XML.java
keatontaylor10: It is a more general reason. When whitespace is not trimmed (new behaviour), an issue occurs with inputs similar to the following: 
``` 
             "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"+
                        "<addresses>\n"+
                        "   <address>\n"+
                        "      <name> Sherlock Holmes </name>\n"+
                        "   </address>\n"+
                        "</addresses>";
```
Where whitespace between the closing and opening tags is incorrectly no longer trimmed and ends up in JSON with the key "content" (or whatever CDataTag is set to). The solution is to remove this value and key at this stage in the parsing if it is empty. 
At this stage the an empty value with tag CData is always going to be unintended unless the input itself has tags with the CData name, for this reason we have stated that inputs with the tags named the same as CDataTag are unsupported with the shouldTrimWhitespace set to false.

src/main/java/org/json/XML.java
keatontaylor10: We are checking here if the value in the JSONObject at key CData is a string or JSONArray to remove it correctly if required.

src/main/java/org/json/XML.java
keatontaylor10: The value at this key will always be empty as it was inadvertently added due to the whitespace in between tags as described above. We remove it as we don't want a key with an empty value

src/main/java/org/json/XMLParserConfiguration.java
keatontaylor10: Will update

src/main/java/org/json/XMLTokener.java
```diff
@@ -45,6 +47,15 @@ public XMLTokener(String s) {
         super(s);
     }
 
+    public XMLTokener(Reader r, XMLParserConfiguration configuration) {
```
stleary, line 50: Please add a method comment explaining what it is used for.

src/main/java/org/json/XML.java
```diff
@@ -431,6 +431,9 @@ private static boolean parse(XMLTokener x, JSONObject context, String name, XMLP
                                             && jsonObject.opt(config.getcDataTagName()) != null) {
                                         context.accumulate(tagName, jsonObject.opt(config.getcDataTagName()));
                                     } else {
+                                        if (!config.shouldTrimWhiteSpace()) {
```
stleary, line 434: Why was removeEmpty() placed here, and not anywhere else in the parse block? e.g. line 425

src/main/java/org/json/XMLTokener.java
```diff
@@ -20,6 +20,8 @@ public class XMLTokener extends JSONTokener {
     */
    public static final java.util.HashMap<String, Character> entity;
 
+   private XMLParserConfiguration configuration = XMLParserConfiguration.ORIGINAL;;
+
```
javadev, line 24: I have found ;; in code

src/main/java/org/json/XML.java
keatontaylor10: This was where the extra whitespace between tags was inadvertently being added and so is removed immediately if the conditions are met

src/main/java/org/json/XMLTokener.java
keatontaylor10: Thanks, will update

src/main/java/org/json/XML.java
stleary: @keatontaylor10 Setting a global flag changes the processing of the entire doc. Maybe it would be better to restrict the whitespace flag to specified tags, similar to how forceList works. Could you make that work for your project?

keatontaylor10:
Hi, is anything else required to get this merged?

stleary:
@keatontaylor10 Please respond to this question:
"Setting a global flag changes the processing of the entire doc. Maybe it would be better to restrict the whitespace flag to specified tags, similar to how forceList works. Could you make that work for your project?"

src/main/java/org/json/XML.java
stleary: This check seems incorrect, I believe CDATA is always a string. Do you have a counterexample?

src/main/java/org/json/XML.java
stleary: Could an empty array have been valid XML text?

keatontaylor10:
> @keatontaylor10 Please respond to this question: "Setting a global flag changes the processing of the entire doc. Maybe it would be better to restrict the whitespace flag to specified tags, similar to how forceList works. Could you make that work for your project?"

For our use case this would not work unfortunately as we don't know what all the tags will be at this point and we need to disable whitespace trimming on all the tags. Would this feature be required to be set per tag?

src/main/java/org/json/XML.java
keatontaylor10: I'm not sure if it is valid or not, but the reason it would be empty at this point is because we are removing all the entries as they are confirmed to have been added inadvertently due to the input having whitespace in between tags. 
The reason for this extra check and removal of tags / entries is because it was noticed during testing that our solution to disable whitespace trimming also disables trimming in the space between closing and opening tags. Some inputs are formated with whitespace to make them readable. The parser's behaviour in this situation is to put the whitespace in an array with the tag "content" (or whatver getcDataTagName() returns). As it would be a rather complicated change to the parser to change this, we opted to rather remove anything that was inadvertently added at this point in the processing. We have also added that using inputs that share the same name as cDataTagName is unsupported when whitespace trimming is disabled to avoid issues. There is an example of the incorrect output in the linked issue as well as a test case which asserts the behaviour of using a tag with the cDataTag name to show why it is unsupported

src/main/java/org/json/XML.java
keatontaylor10: CDATA itself is a string but the value at the tag labelled CDATA is possibly an array as we found in our testing due to some weird behaviour when whitespace in between closing and opening tags gets added to the result instead of trimmed. This is explained a bit more in my comment below

keatontaylor10:
I have synced my branch. I also left a couple more comments to hopefully answer your questions in a bit more detail.
Thanks for taking the time to have a look!

stleary:
@keatontaylor10 My testing did not raise any red flags. However, there is an edge case that's worth noting. Are you sure you want this behavior? If you are OK with it, I can go ahead and approve the PR.

image

Here is a slightly different variation that illustrates CDATA handling can give similar unexpected results even for the legacy code. I guess I would consider this a newly discovered bug in the existing code.

image

keatontaylor10:
@stleary Thanks for testing. This behaviour is fine for my use case as our inputs won't have CDATA tags so we won't run into this problem.

stleary:
Opt-in feature to retain whitespace in XML content when converting to JSON

Yes

**Risks**
Low

**Changes to the API?**
A new flag is available in XMLParserConfiguration

No

No

No, new unit tests were added

No

**Review status**
**APPROVED** 

Starting 3-day comment window

javadev (approved):

stleary (approved):

keatontaylor10:
Thank you for reviewing and getting this merged. 
@stleary what would it take to get these commits built into a new release as that is required for me to pull this into my project, thanks.

stleary:
@keatontaylor10 it's a little early for the next release, but I will try for this weekend.

PR #833, opened 2023-11-27
Refactoring for design smells

adityap27:
Move field.
Change bidirectional association to unidirectional association.
Extract Method.

### DIFF
--- a/src/main/java/org/json/JSONMLArray.java
+++ b/src/main/java/org/json/JSONMLArray.java
@@ -1,67 +1,65 @@
 package org.json;
 
-/*
-Public Domain.
-*/
-
 /**
- * This provides static methods to convert an XML text into a JSONArray or
- * JSONObject, and to covert a JSONArray or JSONObject into an XML text using
+ * This provides static methods to convert an XML text into a
+ * JSONArray, and to covert a JSONArray into an XML text using
  * the JsonML transform.
  *
  * @author JSON.org
  * @version 2016-01-30
  */
-public class JSONML {
+public class JSONMLArray {
     /**
      * Parse XML values and store them in a JSONArray.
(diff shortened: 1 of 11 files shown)


src/main/java/org/json/XMLParserConfiguration.java
```diff
@@ -22,9 +22,6 @@ public class XMLParserConfiguration extends ParserConfiguration {
      */
 //    public static final int DEFAULT_MAXIMUM_NESTING_DEPTH = 512; // We could override
 
-    /** Original Configuration of the XML Parser. */
```
stleary, line 25: Please don't remove public members - users may be using them.

src/main/java/org/json/XML.java
```diff
@@ -24,40 +22,17 @@
 @SuppressWarnings("boxing")
 public class XML {
 
-    /** The Character '&amp;'. */
```
stleary, line 27: Please don't remove public members - users may be using them.

src/main/java/org/json/JSONMLArray.java
```diff
@@ -1,67 +1,65 @@
...
- * This provides static methods to convert an XML text into a JSONArray or
- * JSONObject, and to covert a JSONArray or JSONObject into an XML text using
+ * This provides static methods to convert an XML text into a
+ * JSONArray, and to covert a JSONArray into an XML text using
  * the JsonML transform.
  *
  * @author JSON.org
  * @version 2016-01-30
  */
-public class JSONML {
```
stleary, line 15: Public class names should not be changed

stleary:
Closing due to PR cannot be accepted in its current form. 
Consider making one small set of related changes at a time, instead.

PR #839, opened 2023-12-14
stleary#838 - Fix enum serialization in JSONObject

ThestralWarrior:
Resolves an issue where JSONObject.toString() used Enum.name() instead of Enum.toString() when serializing enum values. Updated the relevant code in JSONObject to use the overridden toString() method for enum values.

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -2653,7 +2653,7 @@ public class JSONObject {
         } else if (value instanceof Boolean) {
             writer.write(value.toString());
         } else if (value instanceof Enum<?>) {
-            writer.write(quote(((Enum<?>)value).name()));
+            writer.write(quote(value.toString()));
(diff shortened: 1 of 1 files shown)


johnjaylward:
Enum.name is correct. This PR is introducing a bug

stleary:
Closing due to failed unit tests.
@ThestralWarrior Thanks for the PR, but it looks like it cannot be accepted, for the reason cited by @johnjaylward.
Developers who wish to submit pull requests are strongly encouraged to run the unit tests first, using Gradle or Maven:
`gradlew clean build test`
Or:
`mvn clean compile test`

robross0606:
> Enum.name is correct. This PR is introducing a bug

Why would you say this is correct?  If you extend and enum and want something else to be used instead of `name`, you should be able to override `toString()` on that enum class and have JSONObject abide by it just like it does for any other class.  I consider current behavior to be a bug.

Issue #850, opened 2024-01-10
Externalize JSON Test Files


MaibuMaddi:
This pull request aims to enhance the test class by externalizing JSON data, improving readability, and promoting maintainability. Instead of embedding JSON directly in the test class, this change stores JSON data in external files, making it easier to manage, update, and collaborate on test data.

Changes Made:

1. **New Directory**: Introduced a new directory named test-data to store externalized JSON files.
2. **JSON Files**: Moved JSON data from the test class to individual files within the test-data directory. Each JSON file corresponds to a specific test scenario or dataset.
3. **Updated Test Class**: Modified the test class to read JSON data from external files in the test-data directory.
4. **Improved Readability**: This change improves the readability of the test class by reducing clutter and separating concerns.

Reasoning:

1. **Maintainability**: Externalizing JSON data into separate files improves maintainability. Changes to test data can now be made without altering the test class, making it easier to manage and update.
2. **Collaboration**: Storing test data in external files facilitates collaboration. Team members can modify or add test data without the need to touch the actual test class, minimizing the risk of conflicts.
3. **Readability**: The separation of concerns enhances the readability of the test class. Test scenarios are now clearly defined in individual JSON files, making it easier to understand the purpose of each test.

Testing:

1. **Existing Tests**: All existing tests have been verified and updated to use the externalized JSON data.
2. **New Test Data**: Added a few new test data files to demonstrate the flexibility of the externalized approach.

stleary:
Closing because the proposed change is refactoring-only

PR #851, opened 2024-01-10
Externalize JSON Test Files #850

MaibuMaddi:
This pull request aims to enhance the test class by externalizing JSON data, improving readability, and promoting maintainability. Instead of embedding JSON directly in the test class, this change stores JSON data in external files, making it easier to manage, update, and collaborate on test data.


stleary:
@MaibuMaddi Thanks for the PR, I appreciate the amount of work that went into it. As you may know, maintaining backward compatibility is important to this project, and the unit tests are a big part of that effort. When very large changes are submitted in a single PR, it's easy to miss something. For example, it looks like you were not working from the latest branch, resulting in a regression in `WeirdList.java`:

image

And 4 lost unit tests in `JsonObjectTest.java`:

image

There may be other issues, I stopped checking after finding the first two. I don't object to refactoring the unit tests, but it would be better to attack this problem a little at a time. Recommend that you start with a small, easily checked change, take care to work from the latest code, and build with both Maven and Gradle before submitting. 

Closing due to not working from the latest branch, and changeset too large to effectively review.

PR #856, opened 2024-02-09
add ability for custom delimiters

michael-ameri:


stleary:
Users should be able to specify the delimiter when converting to/from CDL

Yes

**Risks**
Low

**Changes to the API?**
New API methods were added that include a delimiter param

No

No

No, new unit tests were added

Some of the unit tests were touched, but no functional changes were made to existing tests.
Tabs were used instead of spaces in CDL.java, this will be fixed in a follow-on commit.

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

Issue #861, opened 2024-02-22
Enhancement: Replace assetTrue with assetEquals in tests


Simulant87:
There are a lot of tests with an Aaron pattern like

`assertTrue("message", expected.equals(actual))`

If such an fails, it only returns the message as indicator of the failure, but the difference in expected to actual value is not visible. 

The assertions should be changed to:

`assertEquals("message", actual, expected)`

This way a failing assertion would also output the difference, which would make it easier to find the reason for it. 

I would like to work on this issue and provide a PR to improve the assertions.

stleary:
@Simulant87 Thanks for raising this issue. It would be better to get a 'go-ahead' before proceeding with the PR.   `assertEquals` may be marginally better than `assertTrue`, and it would be reasonable to use it going forward, but I don't see a compelling reason for changing existing tests in a non-functional way.

rikkarth:
I believe it will become more relevant if and when it is decided to move forward with the JUnit 5 upgrade.

Maybe then, if possible we could correct these declarations to more conventional ones.

stleary:
Closing this issue due to recent decision to discourage changes just for the sake of refactoring. Also, there are no plans to upgrade to JUnit 5 at this time.

PR #862, opened 2024-02-22
Ticket #861 replace assertTrue with assertEquals

Simulant87:
Regarding https://github.com/stleary/JSON-java/issues/861

replace assertTrue with assertEquals statements to see the difference between actual and expected output on a failing test.

I can extend on this for all the test classes.


stleary:
No, thank you. The existing `assertTrue` expressions work, and I don't see a compelling reason to change them, especially across all tests.

Issue #863, opened 2024-02-22
Enhancement: Improve toString performance


Simulant87:
The toString methods of JSONObject and JSONArray are not very fast, mostly because of the usage of `StringWriter` and the generic code to support an indent, although in the default toString case it is always 0.

You can compare the performance of the library e.g. with this comparison project, which also list the results in a graph:
https://github.com/fabienrenaud/java-json-benchmark?tab=readme-ov-file#users-model

My suggestion is to improve the performance for the default toString method without indent by duplicating by using a `StringBuilder `instead of a `StringWriter`.

nfaupel:
Closing due to implementation complete in https://github.com/stleary/JSON-java/pull/867

stleary:
Closing due to implementation completed in #867

PR #864, opened 2024-02-22
Ticket #863 rewrite toString with StringBuilder for better performance

Simulant87:
Implementation for https://github.com/stleary/JSON-java/issues/863

Using a StringBuilder to improve the performance for the default `toString` methods without indent.

This duplicates a lot of code, as I assume you want to keep the existing public method interfaces accepting and returning a `Writer` for backwards compatibility. But is results in ~25% faster `toString` method, mostly used for serialisation.

I validated the performance with this project: https://github.com/fabienrenaud/java-json-benchmark
`./run ser --libs ORGJSON`

on my local machine result for the benchmark (thoughput measured in operations per second, higher is better):

currently latest library version 20240205:
```
Benchmark               Mode  Cnt       Score       Error  Units
Serialization.orgjson  thrpt   20  719375,045 ▒ 63677,634  ops/s
```

Improved implementation from this PR:
```
Benchmark               Mode  Cnt       Score       Error  Units
Serialization.orgjson  thrpt   20  905794,579 ▒ 23428,330  ops/s
```

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -52,6 +52,8 @@ jobs:
           java-version: ${{ matrix.java }}
           cache: 'maven'
       - name: Compile Java ${{ matrix.java }}
+        env:
+          MAVEN_OPTS: -Xss4m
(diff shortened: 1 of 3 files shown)


Simulant87:
Closed in favour of https://github.com/stleary/JSON-java/pull/867 as it has comparable or even better performance with way less code changes and no code duplication.

PR #867, opened 2024-02-23
Improve toString Performance: Use StringBuilderWriter for toString methods

Simulant87:
Implementation for https://github.com/stleary/JSON-java/issues/863

Using a `StringBuilder` wrapped in a new `StringBuilderWriter` class to improve the performance for the default `toString` methods.

using the not synchronised `StringBuilder` instead of the `StringBuffer` results in a ~37% faster `toString` method, which is mostly used for serialisation, so this improvement is valuable for any service communication receiving JSON from this library.

I validated the performance with this project: https://github.com/fabienrenaud/java-json-benchmark
`./run ser --libs ORGJSON`

Here is the result for the benchmark on my local machine (thoughput measured in operations per second, higher is better):

currently latest library version 20240205:
```
Benchmark               Mode  Cnt       Score      Error  Units
Serialization.orgjson  thrpt   20  741310,103 ▒ 6250,938  ops/s
```

Improved implementation from this PR:
```
Benchmark               Mode  Cnt        Score      Error  Units
Serialization.orgjson  thrpt   20  1019863,958 ▒ 2538,416  ops/s
```


stleary:
@Simulant87 The benchmarks look good, but I tried testing against a 30m JSON file. The improvement was less than 2% between the latest master branch code (943 ms) and this PR (927 ms). I don't think that is enough to justify accepting this PR. If you want to run the test case yourself, I used the file `train-v1.1.json` from https://www.kaggle.com/datasets/stanfordu/stanford-question-answering-dataset

Simulant87:
@stleary Thank you for having a look at my PR and sharing your source file for your performance test. I repeated the performance test with your input file and wrote this "poor mans performance test" (see below). It was just faster to implement than a correct JMH setup and it should still give an approximate feelingabout the difference. I executed it on master and my branch with the following result:

master branch:
84.533ms for 200 iterations of toString. 422ms per toString on average.
my branch:
54.706ms for 200 iterations of toString. 273ms per toString on average.

the improvement is comparable to the java-json-benchmark I used before with ~35% improvement.
I would be interested in the execution result in your environment.

```
package org.json;

import java.io.FileInputStream;
import java.io.FileNotFoundException;
import java.io.InputStream;

public class Main {

    public static void main(String[] args) throws FileNotFoundException {
        // TODO: update path to 30mb input file train-v1.1.json from https://www.kaggle.com/datasets/stanfordu/stanford-question-answering-dataset
        InputStream inputStream = new FileInputStream("C:\\Users\\admin\\gitprojects\\JSON-java\\train-v1.1.json");
        JSONTokener tokener = new JSONTokener(inputStream);
        JSONObject object = new JSONObject(tokener);

        String output = null;
        // JVM warm up: 100 iterations of toString
        for (int i = 0; i < 100; i++) {
            output = object.toString();
            System.out.println(output.charAt(0));
        }

        // measured run: 200 iterations of toString
        long start = System.currentTimeMillis();
        int iterations = 200;
        for (int i = 0; i < iterations; i++) {
            output = object.toString();
            // access the to string output, so it does not get optimised away
            System.out.println(output.charAt(0));
        }
        long end = System.currentTimeMillis();

        long totalTime = end - start;
        System.out.println(totalTime + "ms for " + iterations + " iterations of toString. " + totalTime / iterations + "ms per toString on average.");
    }
}
```

stleary:
@Simulant87 I think that measuring how fast `toString()` runs in isolation does not provide useful information for users' applications. The test I am starting with is: "What is the performance improvement when parsing a moderately large/complex JSON doc?". 
```
@Test
public void foo() {
    final InputStream resourceAsStream = JSONObjectTest.class.getClassLoader().getResourceAsStream(
            "train-v1.1.json");
    JSONTokener tokener = new JSONTokener(resourceAsStream);
    long startTime = System.currentTimeMillis();
    JSONObject jsonObject = new JSONObject(tokener);
    long endTime = System.currentTimeMillis();
    System.out.println("Time to execute in ms: " + Long.toString(endTime-startTime));
}
```

Simulant87:
@stleary You are right, in my performance improvement I only focused on the `toString()` method, with a noticeable improvement. My use case it that my application creates JSONObjects as internal model and serialises them as response to a REST request. The time to create JSONObject will be the same, but the time to write the response will be improved, as there the `toString()` method is used (at least in my context).

Simulant87:
@stleary I also took a look into the deserialisation parsing a JSONObject from String. I was able to pull of a ~9% improvement by replacing the `switch-case` statement with rather few branches with `if-else` statements in the most used method `JSONTokener#nextString`.

Performance checks I did with [java-json-benchmark](https://github.com/fabienrenaud/java-json-benchmark)
`./run deser --libs ORGJSON`

Here is the result for the benchmark on my local machine (thoughput measured in operations per second, higher is better):
currently latest library version 20240205:
```
Benchmark                 Mode  Cnt       Score       Error  Units
Deserialization.orgjson  thrpt   20  747134,406 ▒ 14445,770  ops/s
```

improvement on this branch:
```
Benchmark                 Mode  Cnt       Score      Error  Units
Deserialization.orgjson  thrpt   20  815610,662 ▒ 1275,573  ops/s
```


Also doing my simple performance test on your large JSON input file:

on master branch:
76.403ms for 200 iterations of parsing JSONObject. 382ms per parsing on average.

Improvement on my branch:
70.433ms for 200 iterations of parsing JSONObject. 352ms per parsing on average.

```
    public static void main(String[] args) throws FileNotFoundException {
        // TODO: update path to 30mb input file train-v1.1.json from https://www.kaggle.com/datasets/stanfordu/stanford-question-answering-dataset
        InputStream inputStream = new FileInputStream("/path/to/train-v1.1.json");
        JSONTokener tokener = new JSONTokener(inputStream);
        JSONObject object = new JSONObject(tokener);

        String output = null;
        // JVM warm up: 100 iterations of toString
        for (int i = 0; i < 100; i++) {
            inputStream = new FileInputStream("C:\\Users\\admin\\gitprojects\\JSON-java\\train-v1.1.json");
            tokener = new JSONTokener(inputStream);
            object = new JSONObject(tokener);
        }

        // measured run: 200 iterations of toString
        long start = System.currentTimeMillis();
        int iterations = 200;
        for (int i = 0; i < iterations; i++) {
            inputStream = new FileInputStream("C:\\Users\\admin\\gitprojects\\JSON-java\\train-v1.1.json");
            tokener = new JSONTokener(inputStream);
            object = new JSONObject(tokener);
        }
        long end = System.currentTimeMillis();

        long totalTime = end - start;
        System.out.println(totalTime + "ms for " + iterations + " iterations of parsing JSONObject. " + totalTime / iterations + "ms per parsing on average.");
    }
```

.github/workflows/pipeline.yml
```diff
@@ -52,6 +52,8 @@ jobs:
           java-version: ${{ matrix.java }}
           cache: 'maven'
       - name: Compile Java ${{ matrix.java }}
+        env:
```
stleary, line 55: Please remove.

src/main/java/org/json/JSONObject.java
```diff
@@ -215,14 +214,13 @@ public JSONObject(JSONTokener x) throws JSONException {
         if (x.nextClean() != '{') {
             throw x.syntaxError("A JSONObject text must begin with '{'");
         }
+        c = x.nextClean();
```
stleary, line 217: Please revert all changes in this method.

src/main/java/org/json/JSONTokener.java
```diff
@@ -295,12 +295,9 @@ public String nextString(char quote) throws JSONException {
         StringBuilder sb = new StringBuilder();
         for (;;) {
             c = this.next();
-            switch (c) {
```
stleary, line 298: Please revert all changes in this file.

.github/workflows/pipeline.yml
Simulant87: I got a stack overflow exception during compiling with JDK 11 on the pipeline and increasing the stack size like this fixes it.

stleary: The pipeline problem should have been fixed in #874. Please remove this change and merge the latest master branch into your code - the pipeline issue should go away.

Simulant87: Yes, that fixes it. removed.

src/main/java/org/json/JSONObject.java
Simulant87: done, all changes in this method are reverted.

src/main/java/org/json/JSONTokener.java
Simulant87: > Please revert all changes in this file. JSONTokener.java

All changes in the file are reverted, JSONTokener.java is no longer in the diff.

All requested changes are done.

src/main/java/org/json/StringBuilderWriter.java
```diff
@@ -0,0 +1,92 @@
+package org.json;
+
+import java.io.IOException;
+import java.io.Writer;
+
+/**
+ * Performance optimised alternative for {@link java.io.StringWriter}
+ * using internally a {@link StringBuilder} instead of a {@link StringBuffer}.
+ */
+class StringBuilderWriter extends Writer {
```
stleary, line 10: Please make this class public

src/main/java/org/json/StringBuilderWriter.java
```diff
@@ -0,0 +1,92 @@
...
+    private final StringBuilder builder;
+
+    /**
+     * Create a new string builder writer using the default initial string-builder buffer size.
+     */
+    StringBuilderWriter() {
```
stleary, line 16: Both constructors should be public

src/test/java/org/json/StringBuilderWriterTest.java
```diff
@@ -0,0 +1,59 @@
+package org.json;
```
stleary, line 1: please move to test/java/org/json/junit

src/main/java/org/json/JSONObject.java
```diff
@@ -2809,8 +2819,6 @@ static final Writer writeValue(Writer writer, Object value,
             }
         } else if (value instanceof Boolean) {
             writer.write(value.toString());
-        } else if (value instanceof Enum<?>) {
```
stleary, line 2812: Please revert this change. I believe the code works correctly with the Enum case in the original location.

src/main/java/org/json/StringBuilderWriter.java
Simulant87: done

src/main/java/org/json/StringBuilderWriter.java
Simulant87: done

src/main/java/org/json/JSONObject.java
Simulant87: yes, it does. Done.

src/test/java/org/json/StringBuilderWriterTest.java
Simulant87: done

stleary:
This is a performance optimization. StringBuilderWriter (uses a StringBuilder) replaces the built-in class StringWriter (uses a StringBuffer). Large JSON docs that contain many strings can be stringified in about half the time with this enhancement. 

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No

**Review status**
**APPROVED**

Starting 3-day comment window

src/main/java/org/json/JSONArray.java
```diff
@@ -1695,7 +1694,10 @@ public String toString() {
      */
     @SuppressWarnings("resource")
     public String toString(int indentFactor) throws JSONException {
-        StringWriter sw = new StringWriter();
+        // each value requires a comma, so multiply the count my 2
```
rikkarth, line 1697: typo in comment

Simulant87:
```suggestion
        // each value requires a comma, so multiply the count by 2
```

rikkarth (commented):
Reviewed and tested locally, ok.

stleary (approved):

PR #870, opened 2024-02-25
add a test case for an enum implementing JSONString

Simulant87:
This test case would fail when the instanceof check for `JSONString` in the `writeValue` method of `JSONObject` would be moved to the bottom below the check for `enum`.

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -2701,6 +2701,7 @@ public class JSONObject {
         if (value == null || value.equals(null)) {
             writer.write("null");
         } else if (value instanceof JSONString) {
+            // JSONString must be checked first, so it can overwrite behaviour of other types below
             Object o;
             try {
                 o = ((JSONString) value).toJSONString();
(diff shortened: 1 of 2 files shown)


stleary:
What is the purpose of this change?

Simulant87:
When the ordering of the instanceof checks in the writeValue method is changed, so that the instanceof enum check if done before the instanceof JsonString, it changed unexpectedly the behaviour of the method, without any failing test. This new text would fail on the described change, ensuring that this expected behaviour does not break.

stleary:
Is this change related to #863? If so, it should be added to the existing PR.

Simulant87:
I noticed this gap in the test cover, while working on #863. So I can add it to that branch. 

I just thought you could merge this test already independent of it, as we still an active discussion on the other branch, and the added test case is also valid without the other case. But if you prefer to have it on the other branch for context, I will add it there.

Simulant87:
I merged it to https://github.com/stleary/JSON-java/pull/867 , which is the branch for ticket https://github.com/stleary/JSON-java/issues/863

stleary:
Closing since the code was moved to #867

PR #874, opened 2024-03-02
Deployment and Pipeline action updates

stleary:
**Description**

* Deployment pipeline is ~~disabled~~ removed until secrets can be added and tested.
* pipeline matrix is split into separate actions for each version. Java 11 builds have been failing intermittently due to concurrent parallel execution.


stleary:
Deployment workflow is failing because the secrets have not been set yet. It is being removed until the secrets can be added. Java 11 build is failing intermittently when run in parallel with other builds. Each build runs separately now.

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No

No

**Review status**
**APPROVED** - by myself

PR #880, opened 2024-03-29
Refactor changes

tanujdoshi:
Hi @stleary 

I have refactored some changes in following files : - 

1) populateMap method in JSONObject  [Decompose Conditional](https://refactoring.guru/decompose-conditional) : - 
  - There was complex condition which i have moved to another method name : isValidMethod [COMMIT_LINK](https://github.com/tanujdoshi/JSON-java/commit/0e0d7ad43006ba184358217693b1aeb211a5cffc)

2) Rename variables in nextToken() method of HTTPTOkener file
- There were two variables name c, and q i have renamed them [COMMIT_LINK](https://github.com/tanujdoshi/JSON-java/commit/134ccb3b40b70780d3c70825f98d635ce9409ab1)

3) [Extract method](https://refactoring.guru/extract-method)  in toString() of HTTP
- toString() method was long, i have divided that method for more readibility. [COMMIT_LINK](https://github.com/tanujdoshi/JSON-java/commit/5abfaf72a70b7937729255a776aa116343ebbd87)

### DIFF
--- a/src/main/java/org/json/HTTP.java
+++ b/src/main/java/org/json/HTTP.java
@@ -113,24 +113,38 @@ public class HTTP {
     public static String toString(JSONObject jo) throws JSONException {
         StringBuilder       sb = new StringBuilder();
         if (jo.has("Status-Code") && jo.has("Reason-Phrase")) {
-            sb.append(jo.getString("HTTP-Version"));
-            sb.append(' ');
-            sb.append(jo.getString("Status-Code"));
-            sb.append(' ');
-            sb.append(jo.getString("Reason-Phrase"));
+            appendResponseHeaderLine(sb, jo);
         } else if (jo.has("Method") && jo.has("Request-URI")) {
-            sb.append(jo.getString("Method"));
-            sb.append(' ');
-            sb.append('"');
-            sb.append(jo.getString("Request-URI"));
-            sb.append('"');
-            sb.append(' ');
-            sb.append(jo.getString("HTTP-Version"));
+            appendRequestHeaderLine(sb, jo);
         } else {
             throw new JSONException("Not enough material for an HTTP header.");
         }
         sb.append(CRLF);
         // Don't use the new entrySet API to maintain Android support
+        appendMainHeaders(sb, jo);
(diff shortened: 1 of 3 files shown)


src/main/java/org/json/JSONObject.java
```diff
@@ -1806,6 +1801,15 @@ && isValidMethodName(method.getName())) {
         }
     }
 
+    private boolean isValidMethod(int modifiers,Method method) {
```
stleary, line 1804: Please add a JavaDoc comment to this new method.

src/main/java/org/json/HTTP.java
```diff
@@ -113,24 +113,38 @@ public static JSONObject toJSONObject(String string) throws JSONException {
     public static String toString(JSONObject jo) throws JSONException {
         StringBuilder       sb = new StringBuilder();
         if (jo.has("Status-Code") && jo.has("Reason-Phrase")) {
-            sb.append(jo.getString("HTTP-Version"));
```
stleary, line 116: I am not sure that the refactoring update in this method is a good idea. Although it shortens the method, it hides a more interesting potential refactoring - notice how many times `status-code`, `reason-phrase`, `method`, `request-uri`, and `http-version` are referenced in the conditionals. It makes the code look half-finished. Can you come up with a way to reduce the number of conditions being checked without affecting the output ordering?

src/main/java/org/json/JSONObject.java
```diff
@@ -1762,12 +1762,7 @@ private void populateMap(Object bean, Set<Object> objectsRecord) {
         Method[] methods = includeSuperClass ? klass.getMethods() : klass.getDeclaredMethods();
         for (final Method method : methods) {
             final int modifiers = method.getModifiers();
```
Simulant87, line 1764: The variable `modifiers` is only used with the method `isValidMethod`. So there is no need for this local variable here.
Instead, it should be a variable inside the method `isValidMethod` and the parameter should be removed from the method signature.

stleary:
Closed due to inactivity

Issue #887, opened 2024-04-13
Complete strictMode for JSONArray


stleary:
Garbage chars at the end of JSONArray docs are still allowed in strict mode. This needs to be fixed.

Here is a list of examples of JSON text strings that should be allowed in default mode, but not in strict mode. This is not an exhaustive list. Unit tests should confirm expected behavior when strictMode is true and false.
```
"[]asdf",
"[]]",
"[]}",
"[][",
"[]{",
"[],",
"[]:",
"[],[",
"[],{"
```

**Note:** It should not be necessary for the code to check for specific trailing chars; strict mode is violated if parsing is completed and there are still chars left in the JSONTokener instance.

rikkarth:
Hi stleary would you please be able to provide one or two examples?

stleary:
Description updated

rikkarth:
![image](https://github.com/stleary/JSON-java/assets/76500344/07be6c4d-e11c-415d-8c12-e1e89f4364ad)

stleary:
Closing due to feature accepted and merged.

PR #888, opened 2024-04-14
fix(#887): complete strictMode for JSONArray

rikkarth:
### Description

If in strict mode, and having reached the end of the **JSONArray** construction life-cycle with no errors, we then perform an input validation.
image

In this validation we explicitly check if (1.) the last bracket was found and (2.) if more characters exist after that.

If the next character is not the end of the array (0), then it means we have more characters after the final closing square bracket, so we throw an error message informing the array is not compliant and we give back the complete input to the user for easy debugging.

image

To ensure this logic doesn't affect 2D arrays I also created a new test for it.
image

From the examples provided in the issue created: https://github.com/stleary/JSON-java/issues/887

### Additional Note
Since the logic is fully encapsulated inside `validateInput()` which is only run if strictMode is true, it is guaranteed that previous implementations are not affected when strictMode is false and that this implementation is modular and can be maintained, improved or removed by simply removing `validateInput()`.

### Test cases added
image

image

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -96,87 +96,75 @@ public class JSONArray implements Iterable<Object> {
      */
     public JSONArray(JSONTokener x, JSONParserConfiguration jsonParserConfiguration) throws JSONException {
         this();
-        if (x.nextClean() != '[') {
+        char nextChar = x.nextClean();
+
+        // check first character, if not '[' throw JSONException
+        if (nextChar != '[') {
             throw x.syntaxError("A JSONArray text must start with '['");
         }
 
-        char nextChar = x.nextClean();
-        if (nextChar == 0) {
-            // array is unclosed. No ']' found, instead EOF
-            throw x.syntaxError("Expected a ',' or ']'");
-        }
-        if (nextChar != ']') {
-            x.back();
-            for (;;) {
-                if (x.nextClean() == ',') {
-                    x.back();
-                    this.myArrayList.add(JSONObject.NULL);
-                } else {
-                    x.back();
-                    this.myArrayList.add(x.nextValue(jsonParserConfiguration));
+        parseTokener(x, jsonParserConfiguration); // runs recursively
+
+    }
+
+    private void parseTokener(JSONTokener x, JSONParserConfiguration jsonParserConfiguration) {
+        boolean strictMode = jsonParserConfiguration.isStrictMode();
+
+        char cursor = x.nextClean();
+
+        switch (cursor) {
+            case 0:
+                throwErrorIfEoF(x);
+                break;
+            case ',':
+                cursor = x.nextClean();
+
+                throwErrorIfEoF(x);
+
+                if (cursor == ']') {
+                    break;
                 }
-                switch (x.nextClean()) {
-                    case 0:
-                        // array is unclosed. No ']' found, instead EOF
-                        throw x.syntaxError("Expected a ',' or ']'");
-                    case ',':
-                        nextChar = x.nextClean();
-                        if (nextChar == 0) {
-                            // array is unclosed. No ']' found, instead EOF
-                            throw x.syntaxError("Expected a ',' or ']'");
-                        }
-                        if (nextChar == ']') {
-                            return;
-                        }
-                        x.back();
-                        break;
-                    case ']':
-                        if (jsonParserConfiguration.isStrictMode()) {
-                            nextChar = x.nextClean();
 
-                            if (nextChar == ','){
-                                x.back();
-                                return;
-                            }
+                x.back();
 
-                            if (nextChar == ']'){
-                                x.back();
-                                return;
-                            }
+                parseTokener(x, jsonParserConfiguration);
+                break;
+            case ']':
+                if (strictMode) {
+                    cursor = x.nextClean();
+                    boolean isNotEoF = !x.end();
 
-                            if (nextChar != 0) {
-                                throw x.syntaxError("invalid character found after end of array: " + nextChar);
-                            }
-                        }
+                    if (isNotEoF && x.getArrayLevel() == 0) {
+                        throw x.syntaxError(String.format("invalid character '%s' found after end of array", cursor));
+                    }
 
-                        return;
-                    default:
-                        throw x.syntaxError("Expected a ',' or ']'");
+                    x.back();
                 }
-            }
-        }
+                break;
+            default:
+                x.back();
+                boolean currentCharIsQuote = x.getPrevious() == '"';
+                boolean quoteIsNotNextToValidChar = x.getPreviousChar() != ',' && x.getPreviousChar() != '[';
 
-        if (jsonParserConfiguration.isStrictMode()) {
-            validateInput(x);
+                if (strictMode && currentCharIsQuote && quoteIsNotNextToValidChar) {
+                    throw x.syntaxError(String.format("invalid character '%s' found after end of array", cursor));
+                }
+
+                this.myArrayList.add(x.nextValue(jsonParserConfiguration));
+                parseTokener(x, jsonParserConfiguration);
         }
     }
 
     /**
-     * Checks if Array adheres to strict mode guidelines, if not, throws JSONException providing back the input in the
-     * error message.
+     * Throws JSONException if JSONTokener has reached end of file, usually when array is unclosed. No ']' found,
+     * instead EoF.
      *
-     * @param x tokener used to examine input.
-     * @throws JSONException if input is not compliant with strict mode guidelines;
+     * @param x the JSONTokener being evaluated.
+     * @throws JSONException if JSONTokener has reached end of file.
      */
-    private void validateInput(JSONTokener x) {
-        char cursor = x.getPrevious();
-
-        boolean isEndOfArray = cursor == ']';
-        char nextChar = x.nextClean();
-        boolean nextCharacterIsNotEoF = nextChar != 0;
-
-        if (isEndOfArray && nextCharacterIsNotEoF) {
-            throw x.syntaxError(String.format("Provided Array is not compliant with strict mode guidelines: '%s'", nextChar));
+    private void throwErrorIfEoF(JSONTokener x) {
+        if (x.end()) {
+            throw x.syntaxError(String.format("Expected a ',' or ']' but instead found '%s'", x.getPrevious()));
         }
     }
 
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -107,7 +107,7 @@ public class JSONArray implements Iterable<Object> {
         }
         if (nextChar != ']') {
             x.back();
-            for (; ; ) {
+            for (;;) {
                 if (x.nextClean() == ',') {
                     x.back();
                     this.myArrayList.add(JSONObject.NULL);
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -107,7 +107,7 @@ public class JSONArray implements Iterable<Object> {
         }
         if (nextChar != ']') {
             x.back();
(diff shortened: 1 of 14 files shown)


src/main/java/org/json/JSONArray.java
```diff
@@ -144,6 +144,42 @@ public JSONArray(JSONTokener x, JSONParserConfiguration jsonParserConfiguration)
...
+     * @throws JSONException if input is not compliant with strict mode guidelines;
+     */
+    private void validateInput(JSONTokener x) {
+        char cursor = x.getPrevious();
+
+        boolean isEndOfArray = cursor == ']';
+        boolean nextCharacterIsNotEoF = x.nextClean() != 0;
+
+        if (isEndOfArray && nextCharacterIsNotEoF) {
+            String completeInput = collectCompleteInput(x);
```
stleary, line 167: It's probably not a good idea to collect complete input, in the case of a large JSONArray. Just the char value, index, and line number should be sufficient.

src/main/java/org/json/JSONArray.java
```diff
@@ -144,6 +144,42 @@ public JSONArray(JSONTokener x, JSONParserConfiguration jsonParserConfiguration)
                 }
             }
         }
+
+        if (jsonParserConfiguration.isStrictMode()) {
+            validateInput(x);
```
stleary, line 149: Please use similar logic to the ']' case starting on line 133.

src/test/java/org/json/junit/JSONParserConfigurationTest.java
```diff
@@ -208,6 +218,15 @@ public void verifyMaxDepthThenDuplicateKey() {
...
         return Arrays.asList(
+            "[]asdf",
+            "[]]",
+            "[]}",
+            "[][",
+            "[]{",
+            "[],",
+            "[]:",
+            "[],[",
+            "[],{",
```
stleary, line 229: Please add this entry to the noncompliant list:
`"[[a]]"`
It will break the unit tests since JSONTokener.nextValue() does not forward the jsonParserConfiguration object when creating a new JSONArray, so some additional work will be needed.

stleary:
@rikkarth Sorry for taking so long to complete the review. Comments added.

src/test/java/org/json/junit/JSONParserConfigurationTest.java
rikkarth: Will do.

src/main/java/org/json/JSONArray.java
rikkarth: Will do.

rikkarth:
1. Fixed issue with double array breaking JSONTokener.nextValue() which now forwards jsonParserConfiguration object when creating a new JSONArray.
2. Changed input validation in JSONArray, removing collection of complete input in favor of leveraging x.syntaxError().

image

image

stleary:
@rikkarth  Please merge from master to pick up the latest changes:
image

stleary:
@rikkarth Looks like there is a regression when parsing array elements with non-string simple values. Please add this unit test:
```
    @Test
    public void shouldHandleNumericArray() {
        String expected = "[10]";
        JSONArray jsonArray = new JSONArray(expected, new JSONParserConfiguration().withStrictMode(true));
        assertEquals(expected, jsonArray.toString());
    }
```

This test passes when strict mode is false, but fails with this output when strict mode is true:
image

Similar problems occur in strict mode when the array elements are true/false, and null.

rikkarth:
Also boolean values are producing the same behaviour, I will fix it.
image

rikkarth:
I changed how I validated numbers, boolean and string values using existing logic, I've also added what I believe a few, more robust, test cases to prevent more regressions.

stleary:
null without quotes is a valid JSON value, just like true and false (see https://json.org). Strict mode should allow this value when it appears without quotes. Please add this test case for coverage:
```
    @Test
    public void allowNullInStrictMode() {
        String expected = "[null]";
        JSONArray jsonArray = new JSONArray(expected, new JSONParserConfiguration().withStrictMode(true));
        assertEquals(expected, jsonArray.toString());
    }

```

rikkarth:
Done as requested. Now it was simple as adding it to the this method:

```java
    private Object getValidNumberOrBooleanFromObject(Object value) {
        if (value instanceof Number || value instanceof Boolean || value.equals(JSONObject.NULL)) {
            return value;
        }

        throw new JSONException(String.format("Value is not surrounded by quotes: %s", value));
    }
```

stleary:
@rikkarth, This string fails in strict mode because there are chars after the end of an empty array:
`String s = "[[]]";`
This string is allowed in strict mode because the chars at the end are not validated:
 `String s = "[1],";`
You might be able to address the first problem by making the parser code block starting at line 160 exactly the same as the code block starting at line 134, as I mentioned sometime back. But it will probably still need to be reworked to handle values after the closing bracket, like JSONObject. The second problem means that the parser is still not reliably detecting invalid trailing chars. If the current implementation is unable to do that, it might be necessary to revisit the design for this feature.

rikkarth:
> @rikkarth, This string fails in strict mode because there are chars after the end of an empty array: `String s = "[[]]";` This string is allowed in strict mode because the chars at the end are not validated: `String s = "[1],";` You might be able to address the first problem by making the parser code block starting at line 160 exactly the same as the code block starting at line 134, as I mentioned sometime back. But it will probably still need to be reworked to handle values after the closing bracket, like a closing JSONObject brace. The second problem means that the parser is still not reliably detecting invalid trailing chars. If the current implementation is unable to do that, it might be necessary to revisit the design for this feature.

Yes you are absolutely right. I will come back to you with a more solid, robust solution and design.

Thank you for your patience.

rikkarth:
I'm near a solution for this I will share my results and implementation soon.

rikkarth:
Mr @stleary, 

After a few days of serious research of the codebase and work I believe I came up to a nice solution.

Ideally I'd like to present you with the changes I performed over a call, but I'll try to do my best through here.

Will present it in the next comment.

rikkarth:
# * DRAFT *

stleary:
@rikkarth Please see #891. Let's pause this review until the unit tests have been updated to give a better view of strict mode behavior.

stleary:
Branch [strict-mode-unit-tests](https://github.com/stleary/JSON-java/tree/strict-mode-unit-tests) contains all of your code from this branch. In the JSONParserConfiguration default constructor, I forced strictMode to be true, and then modified the broken unit tests or replaced @Test with @Ignore. 
* Most of the broken tests were due to inadvertently forgetting to use double quotes in otherwise valid JSON docs. You can ignore those changes. 
* A few tests were set to @Ignore because they intentionally used unquoted text. They will be restored later, and are not a concern. 
* There were a few grey areas, e.g. when an invalid numeric value is supposed to be converted to string, but instead throws an exception. It could be argued either way - should strictMode allow the conversion or not? No decision has been made yet, no action needed at this time. 
* There is a regression in single quote handling within double-quote delimited strings that must be fixed in strictMode.
* There are also several errors in JSONTokenerTest that suggest a regression.

You can fix these issues on your current branch, or you can take a new fork off of the strict-mode-unit-tests branch.

rikkarth:
I will review all this points and will come back to you with. We feedback.

Thank you for this analysis.

rikkarth:
@stleary 

**Update**:
I fixed all the test cases in JSONTokener apart from the last which I require more feedback:

image

## The Bug

The issue was, I didn't perform the most simple of validations which checks for end of file after finding ']' at 2nd, which I included now. After doing it, all tests pass accordingly.
image

As for this test case

```java
@Test
public void testValid() {
    // ...
    checkValid("1 2", String.class);
}

throws `JSONException` which is indeed valid, since you manually set `strictMode` true in `JSONParserConfiguration`.

The test case value is not inside an array;
The test case value is not inside quotes;

What should I do with this case?
```
org.json.JSONException: Value '1 2' is not surrounded by quotes at 3 [character 4 line 1]
```

**Comment by rikkarth** [comment]
This test is failing because the expected testCase is actually malformed, strictMode working accordingly
image

image

**Comment by rikkarth** [comment]
> Branch [strict-mode-unit-tests](https://github.com/stleary/JSON-java/tree/strict-mode-unit-tests) contains all of your code from this branch. In the JSONParserConfiguration default constructor, I forced strictMode to be true, and then modified the broken unit tests or replaced @test with @ignore.

All of the points you've highlighted were fixed. There are a few cases (e.g `checkValid("1 2", String.class);`) where I don't exactly know what I should do there, and will require some more feedback on this.

I removed the TODO's, comments, etc that you provided in [strict-mode-unit-tests](https://github.com/stleary/JSON-java/tree/strict-mode-unit-tests) as I fixed each issue.

I left some of the "important" TODO's the one's you wrote "TBD later" or "will revert later" so you could easily navigate back to them later.

> There is a regression in single quote handling within double-quote delimited strings that must be fixed in strictMode.

Fixed.

> There are also several errors in JSONTokenerTest that suggest a regression.

Fixed.

> You can fix these issues on your current branch, or you can take a new fork off of the strict-mode-unit-tests branch.

I merged them into this branch to keep a linear history of what was affected in this PR.

**Comment by rikkarth** [comment]
I added this TODO's because I strongly believe the test cases here are super bloated and could be hugely simplified.

I will make my case in a future PR, leaving this here for reference.
image

**Comment by stleary** [comment]
@rikkarth Nice work. 
The `1 2` test result signals a regression - in this case, a change in the behavior of a public API, which is always a concern. Might be good to see if the existing behavior can be retained.
Will review the code, and add some comments.

**Comment by rikkarth** [comment]
> @rikkarth Nice work. The `1 2` test result signals a regression - in this case, a change in the behavior of a public API, which is always a concern. Might be good to see if the existing behavior can be retained. Will review the code, and add some comments.

The test for this use-case passes in `strictMode=false` just not in `strictMode=true` which I think makes sense under JSON schema.
image

The tokener is trying to understand if the value is enclosed within an array or object.
image

This is the return when `strictMode=false` when:
```java
new JSONArray("1 2");
```
image

rikkarth:
Could the label for this PR be updated? It's no longer pending redesign but pending review.

Thank you.

stleary:
> Could the label for this PR be updated? It's no longer pending redesign but pending review.

I think the label still applies. The goal of issue #887 was just to catch trailing chars at the end of the JSON doc. There might be a simpler solution that hasn't been found yet. For example, adding properties to the tokener does not seem like the right direction, or the way that JSONArray() was rewritten.

stleary:
> Did you have time to read my previous comment?
> It is the last point I have to complete the points you've defined previously.
> I can make this last test pass, I'm just wondering why should this test pass in strict mode?
> Do you have an opinion on it?

Do you mean the "1 2" validity test? Do you think the result difference is due to code changes to enforce quote-delimited strings, or to detecting trailing chars? In the former case, no additional changes are needed. In the latter case, it should be handled by the redesign. In either case, you don't need to make it work now.

rikkarth:
> > Did you have time to read my previous comment?
> > It is the last point I have to complete the points you've defined previously.
> > I can make this last test pass, I'm just wondering why should this test pass in strict mode?
> > Do you have an opinion on it?
> 
> Do you mean the "1 2" validity test? Do you think the result difference is due to code changes to enforce quote-delimited strings, or to detecting trailing chars? In the former case, no additional changes are needed. In the latter case, it should be handled by the redesign. In either case, you don't need to make it work now.

Yes, the error is exactly that. Quote-delimited strings, which ... makes sense.. If you put quotes around the string it passes the test.

Please check screenshots bellow.

image

image

stleary:
Complete the implementation of strict mode for JSONArray

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No, new unit tests were added

No

**Review status**
**APPROVED** 

Starting 3-day comment window

rikkarth:
Cleaned up strictMode true flag as instructed.

stleary (approved):

PR #889, opened 2024-04-26
Fixed non-idempotent unit tests in `JSONObjectTest`

kaiyaok2:
## Motivation:

A couple of tests in `JSONObjectTest` are not idempotent and fail in the second runs in the same JVM, because they pollute states reused by themselves. Specifically, these tests first assume that `SingletonEnum.getInstance()` or `Singleton.getInstance()` are in a clean state, and then modify them. However, these instances are not cleaned up after the tests, so the repeated runs fail in the initial sanity check.  It shall be good to clean the state pollution so that potential newly introduced tests do not fail in the future due to the shared state polluted by these tests.

Two non-idempotent tests are:
- `testSingletonEnumBean()`
- `testSingletonBean()`

## Error messages for both failed tests in the second run:
```
java.lang.AssertionError: [someString, someInt] expected:<1> but was:<2>
```

## Proposed Fix
Add a `tearDown()` method to restore the instances of `SingletonEnum` and `Singleton` to default states.

### DIFF
--- a/src/test/java/org/json/junit/JSONObjectTest.java
+++ b/src/test/java/org/json/junit/JSONObjectTest.java
@@ -56,6 +56,7 @@ import org.json.junit.data.RecursiveBeanEquals;
 import org.json.junit.data.Singleton;
 import org.json.junit.data.SingletonEnum;
 import org.json.junit.data.WeirdList;
+import org.junit.After;
 import org.junit.Ignore;
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


stleary:
@kaiyaok2 Looks reasonable. How did you come across this issue?

kaiyaok2:
> @kaiyaok2 Looks reasonable. How did you come across this issue?

@stleary The issue is reported by our custom JUnit runner that supports rerunning test suite in the same JVM (different from Surefire where a new JVM is booted for each `mvn test` cycle)

stleary:
Unit tests failing when executed twice in succession in the same JVM

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No

No

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

Issue #901, opened 2024-07-15
JSONArray has a bug when calling addAll with a collection (infinite recursion)


jacob-wieland-gematik:
In Line 1913, the jsonParserConfiguration needs to be passed as last argument

stleary:
@jacob-wieland-gematik Can you provide an example of this bug? 
What version are you using? The referenced line number, 1913, does not match the latest JSONArray code.

hexetia:
I found this problem yesterday while upgrading from version 20231013 to 20240205.
In PR https://github.com/stleary/JSON-java/pull/823 a regression was introduced with the addition of more parameters to the function [addAll](https://github.com/stleary/JSON-java/pull/823/files#diff-fec26d6d9d923315c17c0bef371ffe219fbcd3c025232be18aaaefec838a4d77R1834), the call to it in the function in the addition in the method [addAll from collection](https://github.com/stleary/JSON-java/pull/823/files#diff-fec26d6d9d923315c17c0bef371ffe219fbcd3c025232be18aaaefec838a4d77R1939) did not contemplate the inclusion of all the new parameters.
![image](https://github.com/user-attachments/assets/d7e05527-76b4-4026-b38e-1a8492f4f419)

stleary:
@hexetia Thanks for the info and analysis. Will look into this.

stleary:
Closing due to fixed in #907. Will be merged after 3-day window.

PR #823, opened 2023-11-03
JSON parsing self reference object and array

sk02241994:
Fix for issue #743 
Self reference issue found in JSONObject and JSONArray

PR #907, opened 2024-10-05
Fix a bug when calling JSONArray.addAll() with Collection as Object

hexetia:
That fix the #901 that was introduced in #823

I commented the details of why we need that fix on https://github.com/stleary/JSON-java/issues/901#issuecomment-2395078117


stleary:
Fixes a bug when calling addAll with a collection (infinite recursion)

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No, a new unit test was added

No

**Review status**
**APPROVED**

Starting 3-day comment window

PR #913, opened 2024-11-18
update minimum supported version in readme to Java 8

Simulant87:
As of the FAQs and the build settings, Java 8 is the minimum required version. The readme is still stating that the minimum supported version is Java 1.6. Also added the link to the latest Java 6 supporting version copied from the FAQ.


Simulant87:
Java 6 should still be the minimum required/supported version for the main code, only the test code requires Java 8 as minimum version.

PR #914, opened 2024-11-18
use StandardCharsets.UTF_8 in JSONTokener, delete Java 1.6 build pipeline

Simulant87:
introduced with Java 7 StandardCharsets.UTF_8 does not need to lookup the Charset by name.


Simulant87:
Java 6 should still be the minimum required/supported version for the main code, only the test code requires Java 8 as minimum version.

PR #916, opened 2024-11-21
improve String parsing performance with CharSequenceReader

Simulant87:
I implemented a CharSequenceReader from scratch, which is noticable faster in parsing a String input than the current default implementation.

performance test implementation below.

Measurement on master branch:
I disabled my CPU turbo boost, to have a better comparison and less variance between the runs. Also with turbo boost enabled the results are comparable, just faster and with more variance.

435ms per parsing on average.

Measurement on improved branch:

354ms per parsing on average.

--> 18% performance improvement.

```
import java.io.*;
import java.nio.charset.StandardCharsets;

public class Main {

    public static void main(String[] args) throws IOException {
        // TODO: update path to 30mb input file train-v1.1.json from https://www.kaggle.com/datasets/stanfordu/stanford-question-answering-dataset
        InputStream inputStream = new FileInputStream("C:/JSON-java/train-v1.1.json");
        String content = inputStreamToString(inputStream);
        JSONObject object = new JSONObject(content);
        // JVM warm up: 10 iterations
        for (int i = 0; i < 10; i++) {
            object = new JSONObject(content);
            System.out.println("warmup: " + i);
        }

        // measured runs
        long start = System.currentTimeMillis();
        int iterations = 100;
        for (int i = 0; i < iterations; i++) {
            object = new JSONObject(content);
            System.out.println("test: " + i);
        }
        long end = System.currentTimeMillis();

        long totalTime = end - start;
        System.out.println(totalTime + "ms for " + iterations + " iterations of parsing JSONObject. " + totalTime / iterations + "ms per parsing on average.");
    }

    private static String inputStreamToString(InputStream inputStream) throws IOException {
        ByteArrayOutputStream result = new ByteArrayOutputStream();
        byte[] buffer = new byte[1024];
        int length;
        while ((length = inputStream.read(buffer)) != -1) {
            result.write(buffer, 0, length);
        }
        return result.toString(StandardCharsets.UTF_8.name());
    }

}
```

### DIFF
--- a/src/main/java/org/json/CharSequenceReader.java
+++ b/src/main/java/org/json/CharSequenceReader.java
@@ -1,0 +1,101 @@
+package org.json;
+
+import java.io.Reader;
+
+/**
+ * a Reader to read from a {@link CharSequence} e.g. a {@link String}.
+ */
+public class CharSequenceReader extends Reader {
+
+    private final CharSequence charSequence;
+    private int index;
+    private int markedIndex;
+
+    /**
+     * creates a new CharSequenceReader to read from the given CharSequence
+     * @param charSequence the input to read.
+     */
+    public CharSequenceReader(CharSequence charSequence) {
+        this.charSequence = charSequence;
+    }
+
+    /**
+     * reads the next character from the input.
+     * @return returns the read character or -1 if the end of the input is reached.
+     */
+    @Override
+    public int read() {
+        if (index < charSequence.length()) {
+            return charSequence.charAt(index++);
+        }
+        return -1;
+    }
+
+
+    /**
+     * reads a maximum number of characters from the input into the charBuffer starting at the offset
+     * and returns the number of actual read characters.
+     *
+     * @param charBuffer the target charBuffer to write the read characters into
+     * @param offset the offset index where to start writing into the charBuffer
+     * @param length the maximum number of characters to read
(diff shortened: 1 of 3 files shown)


stleary:
Appreciate the effort, but no, thanks. Improved performance is not itself a compelling reason to modify the code.

Simulant87:
I am sad to hear that and in the end I will accept your decision and not submit any further PRs for it, but please hear me out: I totally understand if there is currently no time to review it. But maybe you can give it a try later on.

I know that simplicity is one of the core values of this project, and therefore the default decision is the reject unnecessary changes. But also a performant implementation is one of the main goals, which this change continues to. There are many other projects depending on this library, which I assume would be happy to have this improvement integrated. As this part of the code is heavily called internally, it is not possible to use this change as an external extension. My implementation is straight forward to review and you accepted other performance improvements as well. I would be very happy if you rethink your initial decision maybe later on.

Simulant87:
For everyone interested: The same performance improvement can also be achieved by passing the improved `Reader` without synchronisation to an explicit` JSONTokener` in the constructor of the `JSONObject`. And there is also an open implementation with the same benefits available by [apache commons io](https://mvnrepository.com/artifact/commons-io/commons-io).

`new JSONObject(new JSONTokener(new org.apache.commons.io.input.CharSequenceReader(content)));`

PR #921, opened 2024-12-15
Restore strict mode text parsing

stleary:

The default JSON-Java library is quite permissive in allowing invalid JSON docs to be parsed successfully. The "strict mode"  feature is for users who need the library to enforce valid JSON text. 

A few months ago, strict mode was added but it was not fully completed. As a result, the code was backed out temporarily and is now restored, with some updates to complete the implementation. 

Strict mode is only available when parsing JSON from a text document.  
To enable strict mode, use code like the following:

```
        JSONParserConfiguration jsonParserConfiguration = 
                new JSONParserConfiguration()
                .withStrictMode(true);
        JSONObject jsonObject = new JSONObject(myString, jsonParserConfiguration));
```

What exactly is excluded in strict mode:
* Unparsed text at the end of JSON documents is not allowed
* JSONArrays must not have multiple consecutive commas
* JSONArray and JSONObject content must not terminate with a comma
* JSONObjects must not use a semicolon to separate a key from a value
* JSONObject keys and string values in JSONArrays or JSONObjects must be surrounded by double quotes

To test for regressions, update the JSONParserConfiguration class to have strict mode enabled by default. Then build the project and run the unit tests, which should all complete successfully.

Yes

**Risks**
Low

**Changes to the API?**
The original strict mode API has been restored.

No

No

No. Some tests that used invalid JSON text have been fixed. A few tests that specifically use invalid text are skipped when strict mode is defaulted. (see regression testing discussion above). New unit tests were added.

The internal implementation of strict mode has been updated.

**Review status**
**APPROVED** - by myself

Starting 3-day comment window

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -67,6 +67,10 @@ public class JSONArray implements Iterable<Object> {
      */
     private final ArrayList<Object> myArrayList;
 
+    private JSONTokener jsonTokener;
+
+    private JSONParserConfiguration jsonParserConfiguration;
+
     /**
      * Construct an empty JSONArray.
      */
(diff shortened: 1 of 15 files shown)


src/main/java/org/json/JSONArray.java
```diff
@@ -114,6 +138,9 @@ public JSONArray(JSONTokener x) throws JSONException {
                         throw x.syntaxError("Expected a ',' or ']'");
                     }
                     if (nextChar == ']') {
+                        if (jsonParserConfiguration.isStrictMode()) {
```
stleary, line 141: Errors should indicate 'strict mode' in the text.
Comments should be added as approriate.

PR #929, opened 2025-01-07
restore-moditect-pom.xml restore plugin

stleary:
Restores 20240303 behavior - module-info 

Yes

**Risks**
Low - this is a restore from a previous release

**Changes to the API?**
No

YES - 20241224 does not contain the module

No

No

No

**Review status**
**APPROVED** - by myself.
No comment window, for immediate release

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -194,6 +194,30 @@
                 </executions>
             </plugin>
             <plugin>
+                <groupId>org.moditect</groupId>
+                <artifactId>moditect-maven-plugin</artifactId>
+                <version>1.0.0.Final</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -193,6 +193,30 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
```
michael-ameri, line 196: @stleary now the plugin is duplicated

https://github.com/stleary/JSON-java/blob/master/pom.xml#L172-L219

stleary:
@michael-ameri Thanks for catching this.

PR #934, opened 2025-01-08
New unit tests for Strict mode using JSONTokener and JSONParserConfiguration

Simulant87:
Adding new test cases with invalid characters at th end of the input to validate that the strict mode is enforced by the constructors of JSONObject and JSONArray. Turns out, that the constructor with JSONTokener instead of String input does not enforce the strict mode at the end of the object creation.

demonstrating, but not fixing, the bug behavior described in https://github.com/stleary/JSON-java/issues/935

### DIFF
--- a/src/test/java/org/json/junit/JSONParserConfigurationTest.java
+++ b/src/test/java/org/json/junit/JSONParserConfigurationTest.java
@@ -1,6 +1,9 @@
 package org.json.junit;
 
-import org.json.*;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.json.JSONParserConfiguration;
 import org.junit.Test;
 
 import java.io.IOException;
--- a/src/test/java/org/json/junit/JSONParserConfigurationTest.java
+++ b/src/test/java/org/json/junit/JSONParserConfigurationTest.java
@@ -4,6 +4,7 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.json.JSONParserConfiguration;
+import org.json.JSONTokener;
 import org.junit.Test;
 
 import java.io.IOException;
--- a/src/test/java/org/json/junit/JSONParserConfigurationTest.java
+++ b/src/test/java/org/json/junit/JSONParserConfigurationTest.java
@@ -1,9 +1,6 @@
 package org.json.junit;
 
-import org.json.JSONArray;
-import org.json.JSONException;
-import org.json.JSONObject;
-import org.json.JSONParserConfiguration;
+import org.json.*;
 import org.junit.Test;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


src/test/java/org/json/junit/JSONParserConfigurationTest.java
```diff
@@ -1,9 +1,6 @@
 package org.json.junit;
 
-import org.json.JSONArray;
-import org.json.JSONException;
-import org.json.JSONObject;
-import org.json.JSONParserConfiguration;
+import org.json.*;
```
Simulant87, line 3:
```suggestion
import org.json.JSONArray;
import org.json.JSONException;
import org.json.JSONObject;
import org.json.JSONParserConfiguration;
```

stleary:
Unit tests are failing, was this tested locally before committing?

Simulant87:
The PR is not intended to be merged right away, while still having a complete green test result. So the currently failing behavior of 2 out of 4 provided tests is intended to demonstrate the current bug behavior and describe the expected behavior after a fix, which I haven't provided.

I provided 2 tests each for JSONObject and JSONArray. Each with the same String input to parse a valid Object/Array, followed by additional invalid parameters and having the strict mode enabled. So all tests are expected to throw an exception, because of the strict mode enabled and the invalid characters at end.
The plain String input works as expected, while the same String input wrapped by a JSONTokener doees not throw the expected exception. This is demonstrated by this PR.

I did not provide a fix for this bug. I tried adding t[his kind of check](https://github.com/stleary/JSON-java/blob/391c86931bb4367c4edf1b1115edc6df89cbb748/src/main/java/org/json/JSONObject.java#L460) [here](https://github.com/stleary/JSON-java/blob/391c86931bb4367c4edf1b1115edc6df89cbb748/src/main/java/org/json/JSONObject.java#L291). Although it makes my provided failing test green, it lead to other existing tests to fail, where nested Objects are parsed. As at the end of the nested object, the following characters for the outer object are matched incorrectly eagerly as invalid content following the nested object. To fix this I assume the JSONTokener should carry some kind of nesting/indent-level information, which has to be check as well and the check should only be applied at the outer most Object/Array level.

stleary:
I think I understand what you are saying. However, the unit tests failures are hidden behind unknown symbol errors, because you did not include this line in JSONParserConfigurationTest.java:
```
import org.json.JSONTokener;
```

Simulant87:
right, was forgot to push the missing import. should be fixed now.

stleary:
@Simulant87 These are good tests, but 2 of them will need to be updated to match actual exception messages, after #937 is merged

Simulant87:
Thank you, I could update them after https://github.com/stleary/JSON-java/pull/937 is merged, or as an alternative I integrated everything togher in https://github.com/stleary/JSON-java/pull/938

Simulant87:
This PR should be closed in favour of https://github.com/stleary/JSON-java/pull/938 (when that one is merged).

stleary:
Closed due to issue resolved in another commit

Issue #933, opened 2025-01-08
NullPointerExceptions parsing non-strict JSON using JSONTokener from version 20241224


bherw:
```Java
String s = "{\"a\": [],}";
var tokener = new JSONTokener(s);
var value = tokener.nextValue();
```

```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "org.json.JSONParserConfiguration.isStrictMode()" because "jsonParserConfiguration" is null
	at org.json.JSONObject.<init>(JSONObject.java:280)
	at org.json.JSONTokener.nextValue(JSONTokener.java:429)
	at Scratch.main(scratch_3.java:8)
```

```Java
String s = "[1,]";
var tokener = new JSONTokener(s);
var value = tokener.nextValue();
```

```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "org.json.JSONParserConfiguration.isStrictMode()" because "jsonParserConfiguration" is null
	at org.json.JSONArray.<init>(JSONArray.java:144)
	at org.json.JSONTokener.nextValue(JSONTokener.java:436)
	at Scratch.main(scratch_3.java:8)
```

It looks like the parsing implementations in the JSONObject and JSONArray constructors that take a JSONTokener don't check if the configuration is null when encountering non-standard JSON input.

Simulant87:
reproduced with test case in https://github.com/stleary/JSON-java/pull/936

stleary:
@bherw Thanks for catching this.

PR #936, opened 2025-01-08
Bugfix: NullPointerException thrown for invalid input on JSONTokener without strict mode

Simulant87:
expect input to be parsed as the invalid characters are at the end after a valid object, but got a NullPointerException instead, providing a fix with null checks.

fixing https://github.com/stleary/JSON-java/issues/933

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -141,14 +141,14 @@ public class JSONArray implements Iterable<Object> {
                     }
                     if (nextChar == ']') {
                         // trailing commas are not allowed in strict mode
-                        if (jsonParserConfiguration.isStrictMode()) {
+                        if (jsonParserConfiguration != null && jsonParserConfiguration.isStrictMode()) {
                             throw x.syntaxError("Strict mode error: Expected another array element");
                         }
                         return;
                     }
                     if (nextChar == ',') {
                         // consecutive commas are not allowed in strict mode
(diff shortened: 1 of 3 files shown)


stleary:
What if JSONTokener had a default JSONParserConfiguration?

Simulant87:
> What if JSONTokener had a default JSONParserConfiguration?

That would also fix the NullPointerExceptio, if the JSONTokener would be initiated with a default JSONParserConfiguration.

However, the JSONParserConfiguration is not passed in the constructor and is not final. it can be [changed by the setter](https://github.com/stleary/JSON-java/blob/391c86931bb4367c4edf1b1115edc6df89cbb748/src/main/java/org/json/JSONTokener.java#L87) after creation of the JSONTokener, which could also set it to null. In the case of passing null to the setter, it should then also create a new default JSONParserConfiguration.

stleary:
@Simulant87 The new unit tests can be kept after #937 is merged, but I believe the code in JSONObject and JSONArray will no longer be needed. But let me know if you disagree.

Simulant87:
I agree, the null checks would no be longer necessary. I integrated everything in https://github.com/stleary/JSON-java/pull/938 which validates it.

Simulant87:
this PR should be closed in favour of https://github.com/stleary/JSON-java/pull/938 (when that one is merged).

stleary:
Closing this PR as the work is being done in other tickets

PR #937, opened 2025-01-10
Remove field references to JSONTokener and JSONParserConfiguration in JSONArray and JSONObject

### LINKED ISSUE #927: JSONTokener reference is retained in JSONObject/JSONArray in 20241224
The JSONTokener object can have references to large objects like i/o buffers. And it's not really used after JSONObject/JSONArray constructor finishes. At least references to JSONTokener (and JSONParserConfiguration) can be nulled at the end of constructor.

Or you can introduce constructor like this:
`private JSONObject(JSONTokener x, JSONParserConfiguration jsonParserConfiguration, boolean requireInputEnd)`
that checks for the absense of trailing chars (by the way this check is duplicated in `JSONObject(String source)` constructor since it delegates to another constructor which checks for the same thing). This new constructor could use JSONTokener/JSONParserConfiguration without introducing fields.

### LINKED ISSUE #933: NullPointerExceptions parsing non-strict JSON using JSONTokener from version 20241224

### LINKED ISSUE #935: Bug: Strict mode is not checked on constructor using JSONTokener and JSONParserConfiguration
the constructors of JSONObject and JSONArray with JSONTokener as input instead of String do not enforce the strict mode at the end of the object creation, when invalid trailing characters are provided at the end of the input.

In demostrated the case in https://github.com/stleary/JSON-java/pull/934 , see the [failing test results](https://github.com/stleary/JSON-java/actions/runs/12679357291/job/35338888436?pr=934).

michael-ameri:
- fixes #935 
- fixes #933 
- resolves #927

TODOs, if general approach is accepted, before merging:
- add missing Javadocs
- add test cases from other PRs
-- https://github.com/stleary/JSON-java/pull/934/files#diff-136646e627f31ffe205a70d544d189378bec0a4948fd6b6fd50d541a6dd989a8R494-R527
-- https://github.com/stleary/JSON-java/pull/936/files#diff-4c5e9ffeff459dde83e61bd6361a19988fc84825ff3b62e40dc775711a5e12e0R328-R343

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -67,12 +67,6 @@ public class JSONArray implements Iterable<Object> {
      */
     private final ArrayList<Object> myArrayList;
 
-    // strict mode checks after constructor require access to this object
-    private JSONTokener jsonTokener;
-
-    // strict mode checks after constructor require access to this object
-    private JSONParserConfiguration jsonParserConfiguration;
-
     /**
      * Construct an empty JSONArray.
      */
@@ -102,14 +96,7 @@ public class JSONArray implements Iterable<Object> {
     public JSONArray(JSONTokener x, JSONParserConfiguration jsonParserConfiguration) throws JSONException {
         this();
 
-        if (this.jsonParserConfiguration == null) {
-            this.jsonParserConfiguration = jsonParserConfiguration;
-        }
-        if (this.jsonTokener == null) {
-            this.jsonTokener = x;
-            this.jsonTokener.setJsonParserConfiguration(this.jsonParserConfiguration);
-        }
-
+        boolean isInitial = x.getPrevious() == 0;
         if (x.nextClean() != '[') {
             throw x.syntaxError("A JSONArray text must start with '['");
         }
@@ -156,10 +143,18 @@ public class JSONArray implements Iterable<Object> {
                     x.back();
                     break;
                 case ']':
+                    if (isInitial && jsonParserConfiguration.isStrictMode() &&
+                            x.nextClean() != 0) {
+                        throw x.syntaxError("Strict mode error: Unparsed characters found at end of input text");
+                    }
                     return;
                 default:
                     throw x.syntaxError("Expected a ',' or ']'");
                 }
+            }
+        } else {
+            if (isInitial && jsonParserConfiguration.isStrictMode() && x.nextClean() != 0) {
+                throw x.syntaxError("Strict mode error: Unparsed characters found at end of input text");
             }
         }
     }
@@ -176,11 +171,6 @@ public class JSONArray implements Iterable<Object> {
      */
     public JSONArray(String source) throws JSONException {
         this(source, new JSONParserConfiguration());
-        // Strict mode does not allow trailing chars
-        if (this.jsonParserConfiguration.isStrictMode() &&
-                this.jsonTokener.nextClean() != 0) {
-                    throw jsonTokener.syntaxError("Strict mode error: Unparsed characters found at end of input text");
-        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


stleary:
@michael-ameri This looks like the right approach, although I am a little surprised it works. Will continue reviewing.

Simulant87:
very good contribution. I added the missing javadoc and adopted my tests in https://github.com/stleary/JSON-java/pull/938 while keeping your commit in the correct git history attribution making it possible to merge https://github.com/stleary/JSON-java/pull/938 instead of this one. If you prefer to have multiple PRs merge, that would also be fine to me, I can also adopt my tests afer this one is merged.

@stleary (Although it would be a backwards compatibility breaking change) I would strongly suggest to remove the getter and setter for JSONParserConfiguration from JSONTokener with the merge of this PR and quickly release a new version to minimise the (faulty) adoption of these methods. Within the library these methods are no longer required. An external caller should not call these methods, as they bring the potential to change the behaviour within a running parsing process, which should use the same configuration end to end.

michael-ameri:
@Simulant87 
> very good contribution. I added the missing javadoc and adopted my tests in https://github.com/stleary/JSON-java/pull/938 while keeping your commit in the correct git history attribution making it possible to merge https://github.com/stleary/JSON-java/pull/938 instead of this one. If you prefer to have multiple PRs merge, that would also be fine to me, I can also adopt my tests afer this one is merged.

thanks for adding the missing parts! I'm totally fine with closing this PR and merging that one.
> @stleary (Although it would be a backwards compatibility breaking change) I would strongly suggest to remove the getter and setter for JSONParserConfiguration from JSONTokener

I completely agree with removing the setter (and then making the field final).

Additionally, I think the `JSONParserConfiguration` shouldn't be passed as an argument in addition to `JsonTokener` [here](https://github.com/stleary/JSON-java/pull/937/files#diff-fec26d6d9d923315c17c0bef371ffe219fbcd3c025232be18aaaefec838a4d77R96) : 

`public JSONArray(JSONTokener x, JSONParserConfiguration jsonParserConfiguration)`

Instead, the `JSONParserConfiguration` of `x` should be used. (Same principle for `JSONObject` constructor)

 e.g., what would be the correct behavior, if `x.getJsonParserConfiguration().isStrictMode() != jsonParserConfiguration.isStrictMode()` ?

stleary:

* Fixes NullPointerException due to uninitialized JSONParserConfiguration
* Don't store config instance in JSONArray and JSONObject
* Ensures JSONTokener always has a JSONParserConfiguration instance

Yes

**Risks**
Low 

**Changes to the API?**
New API methods have been added

No, but the release cadence will be sped up for this commit.

No

No. The code was tested manually by defaulting strictmode and then running all unit tests. One test fails (the last case in testValid()), but this will be fixed in a follow up commit. Also manually tested with the unit tests in #934 and #936.

No

**Review status**
**APPROVED**

Starting 3-day comment window

stleary:
Notes: 

* The additional required unit tests in #934 and #936 will be added in a followup commit.
* testValid() last test case will be skipped on default strictmode test runs
* The JSONTokener config setter will be addressed in a followup commit
* Some aspects of this PR may still be under discussion. Fixes can be done in followup commits after consensus is reached.

stleary:
@michael-ameri Nice work, resolving multiple issues with minimal code changes.

stleary (approved):

PR #957, opened 2025-03-03
Add JSONArray.stream()

oswaldobapvicjr:
This push adds a method to obtain a `Stream` over the elements in a `JSONArray`.

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -2023,4 +2023,13 @@ public class JSONArray implements Iterable<Object> {
                 , cause);
     }
 
+    /**
+     * Returns a sequential {@code Stream} with this JSONArray as its source.
+     *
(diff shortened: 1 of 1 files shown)


src/main/java/org/json/JSONArray.java
```diff
@@ -2023,4 +2024,13 @@ private static JSONException wrongValueFormatException(
...
     }
 
+    /**
+     * Returns a sequential {@code Stream} with this JSONArray as its source.
+     *
+     * @return a sequential {@code Stream} over the elements in this JSONArray
+     */
+    public Stream<Object> stream() {
+        return myArrayList.stream();
+    }
```
javadev, line 2034: There is already a `toList()` method available. We can call `toList().stream()`.

stleary:
@oswaldobapvicjr Please check the [FAQ](https://github.com/stleary/JSON-java/wiki/FAQ) - lib code must compile with Java 6.
Closing due to not compatible with Java 6.

PR #961, opened 2025-03-19
Option to store null value in JSONObject when parsing a map

effad:
This PR provides a new JSONParserConfiguration option to allow a Java `Map.of("foo", null)` to be converted into `{"foo": null}` instead of `{}`.

See also Issue #296 and PR #338, which was rejected due to API concerns. Since now we have a proper configuration object these concerns are addressed.

### DIFF
--- a/src/main/java/org/json/JSONParserConfiguration.java
+++ b/src/main/java/org/json/JSONParserConfiguration.java
@@ -37,6 +37,7 @@ public class JSONParserConfiguration extends ParserConfiguration {
         clone.strictMode = strictMode;
         clone.maxNestingDepth = maxNestingDepth;
         clone.keepStrings = keepStrings;
+        clone.useNativeNulls = useNativeNulls;
         return clone;
     }
 
--- a/src/main/java/org/json/JSONParserConfiguration.java
+++ b/src/main/java/org/json/JSONParserConfiguration.java
@@ -8,6 +8,11 @@ public class JSONParserConfiguration extends ParserConfiguration {
      * Used to indicate whether to overwrite duplicate key or not.
      */
     private boolean overwriteDuplicateKey;
+    
+    /**
+     * Used to indicate whether ignore null values when converting java maps to JSONObject or not.
+     */
+    private boolean javaNullAsJsonNull;
 
     /**
      * Configuration with the default values.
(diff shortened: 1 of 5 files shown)


src/main/java/org/json/JSONParserConfiguration.java
```diff
@@ -67,6 +72,21 @@ public JSONParserConfiguration withOverwriteDuplicateKey(final boolean overwrite
 
         return clone;
     }
+    
+    /**
+     * Controls the parser's behavior when meeting duplicate keys.
```
stleary, line 77: This comment looks wrong

src/main/java/org/json/JSONParserConfiguration.java
```diff
@@ -8,6 +8,11 @@ public class JSONParserConfiguration extends ParserConfiguration {
      * Used to indicate whether to overwrite duplicate key or not.
      */
     private boolean overwriteDuplicateKey;
+    
+    /**
+     * Used to indicate whether ignore null values when converting java maps to JSONObject or not.
+     */
+    private boolean javaNullAsJsonNull;
```
stleary, line 15: Is there a better name for this property? E.G. claude.ai suggested `useNativeNulls`, to clearly describe the configured behavior. Or you can come up with your own choice.

stleary:
Adds the capability to parse null JSON to Java null instead of JSONObject.Null, when the value occurs in a map.

Yes

**Risks**
Low

**Changes to the API?**
Adds a new JSONParserConfiguration option


No

No

No. New unit tests were added

No

**Review status**
**APPROVED**

@effad Thanks for the PR. If you get a chance to make the code updates, they will be included in this PR.
Otherwise, someone will make the changes in a subsequent commit.

effad:
> @effad Thanks for the PR. If you get a chance to make the code updates, they will be included in this PR. Otherwise, someone will make the changes in a subsequent commit.
On it.

src/main/java/org/json/JSONParserConfiguration.java
effad: Fixed the comment.

src/main/java/org/json/JSONParserConfiguration.java
```diff
@@ -8,6 +8,11 @@ public class JSONParserConfiguration extends ParserConfiguration {
      * Used to indicate whether to overwrite duplicate key or not.
      */
     private boolean overwriteDuplicateKey;
+    
+    /**
+     * Used to indicate whether to convert java null values to JSONObject.NULL or ignoring the entry when converting java maps.
+     */
+    private boolean useNativeNulls;
```
michael-ameri, line 15: I think the ``clone()`` method needs to be extended to clone this field too

effad: You're right of course. I just added a test case to `JSONParserConfigurationTest` and fixed it.

stleary (approved):

Issue #967, opened 2025-04-04
Update keepStrings flag behavior


stleary:
`XMLParserConfiguration.withKeepStrings(true)` should set the `keepStrings` flag. It should be cleared if 
`withKeepNumberAsString(false)` or `withKeepBooleanAsString(false)` is called.
It can be reset to true if both `withKeepNumberAsString(true)` or `withKeepBooleanAsString(true)` are called.

This is just to keep isKeepStrings() consistent.

stleary:
Closed tue to fixed in #968

PR #968, opened 2025-04-06
Update keepStrings behavior to reflect changes in keepBooleanAsString, keepNumberAsString

marilynel:
#967


stleary:
Fixes an inconsistency in how keepStrings is handled. For example, if keepStrings was set, the keepStrings flag was not set.
The new behavior is:
* If keepStrings is set, the keepStrings, keepBooleanAsString, and keepNumberAsString flags will be set accordingly.
* If either keepBooleanAsString or keepNumberAsString is set, then keepStrings will be set accordingly.
Internally, there is no change to the existing behavior.

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No

**Review status**
**APPROVED**

stleary (approved):

PR #969, opened 2025-04-13
refactored large test for strict mode

### LINKED ISSUE #952: Refactor JSONObjectTest jsonObjectParsingErrors()
This unit test contains about 2 dozen separate tests. Each try block should be its own unit test.
Some of the tests will fail when default strictMode is true, and these should be addressed without skipping (see other tests that check isStrictMode before proceeding).

marilynel:
Fixes #952

### DIFF
--- a/src/test/java/org/json/junit/JSONObjectTest.java
+++ b/src/test/java/org/json/junit/JSONObjectTest.java
@@ -2316,332 +2316,417 @@ public class JSONObjectTest {
         }
     }
 
-    /**
-     * Explore how JSONObject handles parsing errors.
-     */
-    @SuppressWarnings({"boxing", "unused"})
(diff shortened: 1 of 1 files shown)


stleary:
A few dozen unit tests were skipped during strict mode, but only 2 were actually failing. This commit refactors the unit tests into separate methods. 

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No, unit tests were refactored

Yes

**Review status**
**APPROVED**

Starting 3-day comment window

**NOTE** 2 tests had the exception message assert commented out, since the purpose was to confirm the exception, not the message. However, this is still tech debt that can be fixed in a later commit.

stleary (approved):

PR #987, opened 2025-06-04
Refactor: Check equality with literal on lhs

AlexCai2019:
There are some equals() that are not `constant.equals(variable)`, but `variable.equals(constant)`.
I suggested to always use `constant.equals(variable)`, since it can avoid vary unintentionally bug, such as `NullPointerException`.

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -334,13 +334,11 @@ public class JSONArray implements Iterable<Object> {
      */
     public boolean getBoolean(int index) throws JSONException {
         Object object = this.get(index);
-        if (object.equals(Boolean.FALSE)
-                || (object instanceof String && ((String) object)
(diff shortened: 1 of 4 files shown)


stleary:
@AlexCai2019 refactoring is not encouraged in this project, but I think in this case it improves the code. 
Test coverage is good overall, except for JSONObject removeLeadingZerosOfNumber()
image

If you get a chance to add some test coverage in the next 3 days, it would be appreciated. Otherwise, I can add something later.

stleary:
Improves code safety by checking for string equality with literals on the lhs.

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No, but low coverage was found in one method.

This commit is a refactor.

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

PR #988, opened 2025-06-07
removed unused method from jsonobject

stleary:
A previous commit included a change to this code, but a later check showed the method is not used, so it is removed here.

Yes

**Risks**
Low

**Changes to the API?**
No, it was a private method

No

No

No

This commit was a refactor

**Review status**
**APPROVED** - by myself

Starting 3-day comment window


stleary:
Thanks @javadev, somehow I forgot about this PR due to family travel.

PR #989, opened 2025-06-22
Minor refactoring

AlexCai2019:
I found these 2 can be optimized during the last refactor.
- Remove `NULL.equals(object)` in `warp()` method on line 2756 of `JSONObject.java` since line 2752 has already tested it.
- Remove the empty string in `readByIndexToken()` method on line 249 of `JSONPointer.java`.

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -2752,8 +2752,7 @@ public class JSONObject {
             if (NULL.equals(object)) {
                 return NULL;
             }
-            if (object instanceof JSONObject || object instanceof JSONArray
-                    || NULL.equals(object) || object instanceof JSONString
+            if (object instanceof JSONObject || object instanceof JSONArray || object instanceof JSONString
                     || object instanceof Byte || object instanceof Character
                     || object instanceof Short || object instanceof Integer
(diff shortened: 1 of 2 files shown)


src/main/java/org/json/JSONObject.java
```diff
@@ -2752,8 +2752,7 @@ private static Object wrap(Object object, Set<Object> objectsRecord, int recursi
             if (NULL.equals(object)) {
                 return NULL;
             }
-            if (object instanceof JSONObject || object instanceof JSONArray
-                    || NULL.equals(object) || object instanceof JSONString
+            if (object instanceof JSONObject || object instanceof JSONArray || object instanceof JSONString
```
javadev, line 2755: formatting

javadev (approved):

stleary:
Removes an unneeded check for null in JSONObject.wrap()

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No

JSONPointer readByIndexToken() had a trivial change

**Review status**
**APPROVED**

Starting 3-day comment window

src/main/java/org/json/JSONObject.java
AlexCai2019: how should I do that?

Issue #822, opened 2023-10-27
Ordered output/file writing of JSONObjects (may be repeated, nevertheless please read carefully ...)


openworld42:
In one of my [repositories] (https://github.com/openworld42/Cellolution) I am going to save the simulation state (lots of data) in a JSON file, which is shorter can be easier changed/debugged than XML (which I used a lot in past projects). To do this in a realistic way, there is a need of ordering (similar to XML).

First of all: I had a deep dive into the code, made a fork (for my own); and yes, I read the FAQ and other files, especially "Will ordering ever be allowed ...", and I know the (IMHO unfortunate) spec "unordered"    ;-)

In any case, I will use JSON-java. It suits the needs close to perfect, pretty printing, easy usage, functionality, light weight, etc. 
Therefore, I will add ordering in the fork.

My suggestion: I can contribute the ordering to JSON-java - as far as I can understand not breaking the interfaces.
Due to the fact, that this a lot more work than doing it for my repo needs, I want to discuss the item and get possibly an ok for it (Assuming my code will please so ;-)

As far as I can see, there are several possibilities to add ordering not breaking the interfaces - if this is not rejected in general for some reasons:

1) Add a class member keyList (ArrayList) to  JSONObject - which is null by default, not a big memory consumption. In a Context class (or a static variable/setter within JSONObject: let say setOrdering(true) before the user takes action) the keyList gets a small ArrayList. If the keyList or flag exists/is set, all put(), accumulate(), populate()-dependent ops will add the key(s) to that keyList - this preserves ordering, remove ops have to take care of keyList. All get(), keys(), keySet(), entrySet(), ... ops will return ordered items, cause the keys will come in an ordered way or Set from keyList. All map-dependent ops will still work in an unordered way.
AFAIK, no interface is broken, all existing code works like in the past, old output/files/Strings are still the same - e.g. a previous JSONObject.setOrdered(true) makes things ordered. Some work for me ;-)

2) Extend JSONObject to let say JSONObjectOrdered,  add a member keyList (ArrayList) and override most of the above methods, using the keyList to retrieve any key, entrySet etc. Possibly some private methods have to be copied (e.g. private wrap() - not a smart solution).
In that case, other classes may undergo a lot of changes like: isInstanceOf JSONObjectOrdered. 
Any ordered attempt can be made by using JSONObjectOrdered instead of JSONObject to generate the JSON tree.
[...]

PR #998, opened 2025-07-13
Allow users to extend JSONObject with custom map

spaffrath:
For Issue #822. I know this has been a frequently discussed issue and I agree with the sentiment of keeping the repo as a reference implementation of the JSON spec. Allowing the map initialization to be overridden makes no functional changes to the repo at all but allows users the flexibility to extend the implementation to provide whatever ordering they may desire, be it FIFO, alphabetical, or whatever else may be needed.

Thank you for your consideration.


stleary:
Sorry, appreciate the effort, but this change should be done in a forked repository. It can't be accepted for this project.

Issue #980, opened 2025-05-16
Some strict mode issues


eleumik:
Hi,

while performing tests on read-only behavior of many JSON parsers I found some issues with this library, testing strict mode only.

The following are all "forgiven":

```
{"a":True}
{"a":TRUE}
{true : 3}
{TRUE : 3}
{1 : 3}
```

I noticed occurrences of [equalsIgnoreCase] around the code, which I suspect is the origin. You’ll know better than me, but probably you just need a helper method to choose (equals) based on context and value. The [true] or [1] key cases, I’m not sure about.

Also, this shouldn’t parse in strict mode, I believe:

```
{ "a" : 0. }
```

The good news: only 9 failures out of 124 tests for this lib (only Jackson at zero, after some config and a custom wrapper for duplicate keys, not little work). Fixing these would resolve 6 failures. Of the remaining 3: 2 are for multiple roots (`{} {}`), and 1 for (`{"a":[,4,null,8]}`) where the first item is treated as JSON-null; some libs skip it, some (correctly, I think) fail.

Thanks,  
Michele Vivoda

stleary:
@eleumik  Thank you, this is very helpful. Fixes will be provided soon.

stleary:
Closing due to fixed in #999

PR #999, opened 2025-07-13
fixed some strict mode issues

marilynel:
#980 

Addressed strict mode issues with boolean and null casing in values, and boolean, null, and number as keys.

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -213,6 +213,7 @@ public class JSONObject {
         this();
         char c;
         String key;
+        Object obj;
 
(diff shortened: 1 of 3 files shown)


stleary:
Fixes problems where Strict mode was not enforcing case for non-string values, and allowing non-string keys.

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No, new unit tests were added

No. Sonarqube found some minor issues. Submitter can fix them if time permits, otherwise I will fix them later.

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

PR #1000, opened 2025-07-20
fixing sonarcube issues

marilynel:


stleary:
Addresses SonarQube issues from recent commits. In the case of JSONObject populateMap(), there will be a marginal performance cost.
This is considered acceptable since bean parsing is not thought to be an especially common operation, and the cost is not high (2 method calls were added).

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No

Yes, this PR is just refactoring

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

PR #1005, opened 2025-09-07
fixing sonar cube issues

marilynel:


src/test/java/org/json/junit/JSONObjectTest.java
```diff
@@ -3896,8 +3896,8 @@ public void issue743SerializationMapWith512Objects() {
 
     @Test
     public void issue743SerializationMapWith1000Objects() {
-      HashMap<String, Object> map  = buildNestedMap(1000);
-      JSONParserConfiguration parserConfiguration = new JSONParserConfiguration().withMaxNestingDepth(1000);
+      HashMap<String, Object> map  = buildNestedMap(500);
```
stleary, line 3899: Please add a comment explaining why the limit was changed. Also, method signature should be updated to '500Objects'.

Not why this testcase started failing in laptop tests.

stleary:
Address sonarqube issues in JSONObject. More to come.

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No. There is one unit testcase that has been breaking in local builds, this was fixed. More investigation is needed, but not for this PR.

All of the code changes were for refactoring.

**Review status**
**APPROVED**

@marilynel Please fix the JavaDoc comments, if you get a chance before merge. Otherwise, they can be included  in the next batch of SonarQube fixes. Also, not sure why SonarQube is reporting a new issue for code that hasn't changed in more than a decade. Will address it with the rest of the issues.

Issue #1003, opened 2025-08-16
Implement JSONObject.fromJson() with unit tests


stleary:
Gson has an API method \<T> fromJson(String, Class\<T>), which converts a JSON doc to an instance of class T.
JSON-Java should have a similar capability. 
New unit tests to thoroughly exercise the functionality should be included.

stleary:
Feature is implemented, merge pending

stleary:
Closing this issue due to feature is completed in #1006. If there are any questions or issues with the fix, please open a new ticket.

PR #1006, opened 2025-09-09
1003: Implement JSONObject.fromJson() with unit tests

sk02241994:
#1003 

Add support for JSON to POJO deserialization via JSONBuilder

Introduced JSONBuilder class:
Provides customizable mappings for type conversions (Function<Object, ?>) and collection instantiation (Supplier<?>), enabling extensible JSON deserialization logic.

Enhanced JSONObject:
Added support for converting JSON into Java POJOs using a new method: fromJson(Class<T>).
Supports nested objects, collections (e.g., List<T>), and custom type mappings via JSONBuilder.
Optional JSONBuilder can be passed via constructor or setter.
Added comprehensive unit tests in JSONObjectTest:
Covers deserialization of primitive types, nested classes, collections, and parameterized generics (e.g., List<List<String>>).
Includes tests for custom type mapping (e.g., LocalDateTime).

### DIFF
--- a/src/main/java/org/json/JSONBuilder.java
+++ b/src/main/java/org/json/JSONBuilder.java
@@ -1,0 +1,122 @@
+package org.json;
+
+import java.util.Map;
+import java.util.HashMap;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.Set;
+import java.util.HashSet;
+import java.util.Collection;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * The {@code JSONBuilder} class provides a configurable mechanism for
+ * defining how different Java types are handled during JSON serialization
+ * or deserialization.
+ *
+ * <p>This class maintains two internal mappings:
+ * <ul>
+ *   <li>A {@code classMapping} which maps Java classes to functions that convert
+ *   an input {@code Object} into an appropriate instance of that class.</li>
+ *   <li>A {@code collectionMapping} which maps collection interfaces (like {@code List}, {@code Set}, {@code Map})
+ *   to supplier functions that create new instances of concrete implementations (e.g., {@code ArrayList} for {@code List}).</li>
+ * </ul>
+ *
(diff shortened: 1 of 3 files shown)


stleary:
@sk02241994 I won't have time to review in depth until later this week. My initial take is that fromJson() is long overdue, glad to see someone submit an implementation, thanks for that! For now, it will be better to get everything working and released with just the default options, so JSONBuilder can wait for a future PR. Also, all of the new SonarQube issues will need to be addressed, or marked false-positive if appropriate.

sk02241994:
@stleary  Sorry for the delay.
I have updated the to include docs and also solved the SonarQube issues (Some of those are false positives)
Also added methods to handle Enum and Map (They were missing in previous commit).

stleary:
@sk02241994 Can you do the implementation without adding any new classes? I understand that this will limit the types that fromJson() can support. The JavaDocs for the new methods should explain the limitations.

sk02241994:
@stleary  Sorry for the delay. I have updated the code to deserialize limited number of classes. Only those which are present in the JSONObject can be deserialized. Since we need to TypeConverter and InstanceCreatory interface, I am keeping those but changing the access type to be default, so no one can access those.

stleary:
@sk02241994 The code cannot be accepted in its current form. `fromJson()` can be implemented without the new classes `TypeConverter` and `InstanceCreator`, while retaining all of the required functionality. This will make the code cleaner and easier to maintain. Let me know if you need some suggestions on how to do this.

sk02241994:
@stleary I have updated the code there are some sonar issues which are irrelevant to the code changes I have made.

stleary:
Thanks @sk02241994, the changes look good, and I should be able to complete the review today.

stleary:
Implement fromJson() for JSONObject

Yes

**Risks**
Low

**Changes to the API?**
2 new API methods are added

No

No

No, new unit tests were added

No

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

PR #1019, opened 2025-10-24
Fix flaky

natnaelhaileselassie:
[INCOMPLETE PR]
### What is the purpose of this PR
This PR fixes the flaky test in JSONObjectTest.
There is nondeterminism which causes inconsistent test results. This happens because of ___ value. Using ____may fix the issue.

### Reproduce the test failure
Run the test using Non-Dex:

- `mvn clean install`
- `mvn clean test edu.illinois:nondex-maven-plugin:2.1.1:nondex`
or
- `mvn clean install`
- `mvn test edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=org.json.junit.JSONObjectTest.valueToString`

### Expected result
The `JSONObjectTest` passes without NonDex shuffling.
The `JSONObjectTest` passes with NonDex shuffling.
### Actual result
The `JSONObjectTest` passes without NonDex shuffling.
The `JSONObjectTest
` fails with NonDex shuffling.
### Why the test fails
The `insert issue` does `insert issue` incorrectly so it accesses data in a non-deterministic way.
.
### Fix
For the test, we changed _____  to fix the error regarding non-deterministism.

### DIFF
--- a/src/test/java/org/json/junit/JSONStringerTest.java
+++ b/src/test/java/org/json/junit/JSONStringerTest.java
@@ -2,7 +2,6 @@ package org.json.junit;
 
 /*
 Public Domain.
-
 */
 
 import static org.junit.Assert.*;


stleary:
Closing due to no files changed for this PR

PR #1027, opened 2025-12-20
Save/restore default locale in test

### LINKED ISSUE #1023: Test does not restore default locale
https://github.com/stleary/JSON-java/blob/master/src/test/java/org/json/junit/JSONObjectLocaleTest.java

This test changes the default JVM locale and doesn’t restore it.
Since Maven runs tests in the same JVM unless forking is enabled, this can influence other tests.

Yesterday i checked all the occurrences of Locale.setDefault in my projects and I found also this. It causes subtle problems when running / mvn packaging  many modules together.

It is of course very low priority and relevant only if one runs many builds in sequence, more for other projects than for this one

Thanks
Michele

PS. today is TimeZone.setDefault day, yes I am having fun...

Simulant87:
fixes #1023

### DIFF
--- a/src/test/java/org/json/junit/JSONObjectLocaleTest.java
+++ b/src/test/java/org/json/junit/JSONObjectLocaleTest.java
@@ -36,25 +36,31 @@ public class JSONObjectLocaleTest {
 
         MyLocaleBean myLocaleBean = new MyLocaleBean();
 
-        /**
-         * This is just the control case which happens when the locale.ROOT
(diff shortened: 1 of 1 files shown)


stleary:
Changing the default locale could affect other workflows. Save and restore the default for other maven tests.

Yes

**Risks**
Low

**Changes to the API?**
No

No

No

No, it fixes a test that leaks a locale change

No

**Review status**

**APPROVED**

Starting 3-day comment window

PR #1037, opened 2026-01-27
Fix ClassCastException in JSONML.toJSONArray and toJSONObject

### LINKED ISSUE #1034: ClassCastException in JSONML.toJSONArray due to Unsafe Type Cast
## Summary

`JSONML.toJSONArray()` performs an unsafe cast of the return value from the internal `parse()` method. When processing malformed XML input, `parse()` returns a `String` instead of `JSONArray`, causing `ClassCastException` and crashing applications that don't explicitly catch this unchecked exception.

## Root Cause

The `parse()` method returns `Object` type, but `toJSONArray()` casts directly without type checking:

### Vulnerable Code (JSONML.java:279)
```java
public static JSONArray toJSONArray(String string) throws JSONException {
    // parse() returns Object, but may actually return String for malformed input
    return (JSONArray)parse(new XMLTokener(string), true, null,
                           JSONMLParserConfiguration.ORIGINAL, 0);
    // No type check before cast!
}
```

### Trigger file

A crafted `crash_input.bin` file (13 bytes) containing malformed XML:
- Starts with `<\n/` (open tag with newline and slash)
- Contains invalid UTF-8 bytes `0xFF`
- Causes `parse()` to return String instead of JSONArray

### How to generate crash_input.bin

```bash
# Base64 decode
echo "PAov//////////8+Qg==" | base64 -d > crash_input.bin

# Verify (should be 13 bytes)
xxd crash_input.bin
# Output: 3c0a 2fff ffff ffff ffff ff3e 42
```

```python
# crash_gen.py
import base64

payload = base64.b64decode("PAov//////////8+Qg==")
with open("crash_input.bin", "wb") as f:
    f.write(payload)

print(f"Generated crash_input.bin ({len(payload)} bytes)")
```

---

## Trigger Method 1: Direct API Call (Library Usage)

Since json-java is a library, applications trigger this by calling the API directly:

```java
import org.json.JSONML;
import java.nio.file.Files;
import java.nio.file.Path;

public class ReproduceJSONML {
    public static void main(String[] args) throws Exception {
        // Read crash input
        byte[] data = Files.readAllBytes(Path.of("crash_input.bin"));
        String input = new String(data);

        // This throws ClassCastException
        JSONML.toJSONArray(input);
    }
}
```

**Build and run:**
```bash
# Compile json-java library
javac -d . src/main/java/org/json/*.java

# Compile and run reproduction
javac -cp . ReproduceJSONML.java
java -cp . ReproduceJSONML
```

**Output:**
```
java.lang.ClassCastException: class java.lang.String cannot be cast to class org.json.JSONArray
    at org.json.JSONML.toJSONArray(JSONML.java:279)
    at ReproduceJSONML.main(ReproduceJSONML.java:10)
```

---

## Trigger Method 2: Fuzzer (Jazzer)

```java
// Copyright 2025 O2Lab
// SPDX-License-Identifier: Apache-2.0

import org.json.JSONML;
import java.nio.charset.StandardCharsets;

public class JsonMLFuzzer {
    public static void fuzzerTestOneInput(byte[] data) {
        String input = new String(data, StandardCharsets.UTF_8);
        try {
            JSONML.toJSONArray(input);
        } catch (JSONException e) {
            // Expected parsing errors - ignore
        }
        // ClassCastException is NOT caught - will crash fuzzer
    }
}
```

**Build and run with Jazzer:**
```bash
# Build with oss-fuzz / Jazzer
python3 infra/helper.py build_fuzzers --clean json-java

# Run fuzzer
./JsonMLFuzzer -rss_limit_mb=2048 corpus/
```

---

## Impact

| Aspect | Details |
|--------|---------|
| **Type** | Denial of Service (DoS) |
| **Severity** | Medium |
| **Attack Vector** | Malformed XML input to `JSONML.toJSONArray()` |
| **Affected Components** | `JSONML.toJSONArray()`, `JSONML.toJSONArray(XMLTokener)` |
| **Affected Versions** | All versions up to 20251224 |
| **CWE** | CWE-704 (Incorrect Type Conversion or Cast) |

---

## Suggested Fix

Add type checking before the cast:

```java
public static JSONArray toJSONArray(String string) throws JSONException {
    Object result = parse(new XMLTokener(string), true, null,
                         JSONMLParserConfiguration.ORIGINAL, 0);
    if (result instanceof JSONArray) {
        return (JSONArray) result;
    }
    throw new JSONException("Expected JSONArray but got " +
                           result.getClass().getSimpleName());
}
```

---

Contribution: This potential vulnerability was found by AI-Based Vuln Detection system FuzzingBrain https://github.com/o2lab/afc-crs-all-you-need-is-a-fuzzing-brain.

OwenSanzas:
Add type checking before casting parse() results to JSONArray/JSONObject. When parse() returns an unexpected type (e.g., String for malformed input), the code now throws a descriptive JSONException instead of ClassCastException.

This prevents unchecked exceptions from propagating to callers who only expect JSONException from these methods.

Fixes #1034

### DIFF
--- a/src/test/java/org/json/junit/JSONMLTest.java
+++ b/src/test/java/org/json/junit/JSONMLTest.java
@@ -986,4 +986,70 @@ public class JSONMLTest {
         }
     }
 
+    /**
+     * Tests that malformed XML causing type mismatch throws JSONException.
+     * Previously threw ClassCastException when parse() returned String instead of JSONArray.
+     * Related to issue #1034
+     */
+    @Test(expected = JSONException.class)
+    public void testMalformedXMLThrowsJSONExceptionNotClassCast() {
+        // This malformed XML causes parse() to return wrong type
+        byte[] data = {0x3c, 0x0a, 0x2f, (byte)0xff, (byte)0xff, (byte)0xff,
+                      (byte)0xff, (byte)0xff, (byte)0xff, (byte)0xff,
+                      (byte)0xff, 0x3e, 0x42};
+        String xmlStr = new String(data);
+        JSONML.toJSONArray(xmlStr);
+    }
+
+    /**
+     * Tests that type mismatch in toJSONObject throws JSONException.
+     * Validates safe type casting in toJSONObject methods.
+     */
+    @Test
+    public void testToJSONObjectTypeMismatch() {
+        // Create XML that would cause parse() to return wrong type
+        String xmlStr = "<\n/\u00ff\u00ff\u00ff\u00ff\u00ff\u00ff\u00ff\u00ff>B";
+        try {
+            JSONML.toJSONObject(xmlStr);
+            fail("Expected JSONException for type mismatch");
+        } catch (ClassCastException e) {
+            fail("Should throw JSONException, not ClassCastException");
+        } catch (JSONException e) {
+            // Expected - verify it's about type mismatch
+            assertTrue("Exception message should mention type error",
+                e.getMessage().contains("Expected") || e.getMessage().contains("got"));
+        }
+    }
+
+    /**
+     * Tests that valid XML still works correctly after the fix.
+     * Ensures the type checking doesn't break normal operation.
+     */
+    @Test
+    public void testValidXMLStillWorks() {
+        String xmlStr = "<root><item>value</item></root>";
+        try {
+            JSONArray jsonArray = JSONML.toJSONArray(xmlStr);
+            assertNotNull("JSONArray should not be null", jsonArray);
+            assertEquals("root", jsonArray.getString(0));
+        } catch (Exception e) {
+            fail("Valid XML should not throw exception: " + e.getMessage());
+        }
+    }
+
+    /**
+     * Tests that valid XML to JSONObject still works correctly.
+     */
+    @Test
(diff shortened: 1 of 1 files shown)


stleary:
Changes look OK, but where are the unit tests to confirm the new behavior?

[stleary 1/28/2026] Issue fixed

stleary:
Replace ClassCastException with a more useful JSONException in JSONML toJSONArray() and toJSONObject()

Yes

**Risks**
Low

**Changes to the API?**
Yes, a different exception is thrown. API changes are not allowed, but in this case the ClassCastException is arguably a flaw in the original design.

No

No

No, and new unit tests were added ~but no new unit tests were added. This will need to be corrected in order to complete this PR.~

No

**Review status**
**APPROVED**

Starting 3-day comment window

OwenSanzas:
Hey @stleary, all requested changes are complete. Could you please take a look?                                                                                                                                    
                                                                                                                                                                                                                     
After the fix:                                                                                                                                                                                                     
  1. PoC for #1034 is mitigated                                                                                                                                                                                      
  2. All newly-added unit tests pass                                                                                                                                                                                 
  3. All existing unit tests pass

PR #1044, opened 2026-03-12
Ignore static fields in JSONObject.fromJson()

### LINKED ISSUE #1043: Potentially unsafe deserialization in JSONObject#fromJson() static fields are not excluded
This behavior may be security-relevant when `JSONObject#fromJson(Class<T>)` is used on untrusted input, so I am reporting it here as a public issue in line with the repository's published vulnerability-reporting guidance.

In release `20251224`, `JSONObject#fromJson(Class<T>)` iterates `clazz.getDeclaredFields()`, calls `Field#setAccessible(true)`, and assigns matching JSON values without excluding `static` fields:

https://github.com/stleary/JSON-java/blob/cf653682be81e5ff630c8ac5527a5060b9a813e8/src/main/java/org/json/JSONObject.java#L3365-L3378

As a result, deserialization can write not only per-instance state but also mutable declared class-level state on the caller-selected target type, subject to normal runtime access restrictions. That expands the binding surface beyond ordinary object population and may have integrity implications in applications that use `fromJson()` with untrusted JSON.

If helpful, I can provide a minimal reproducer and a regression test. A straightforward mitigation appears to be skipping static fields during binding.

yuki-matsuhashi:
Fixes #1043

This PR updates `JSONObject#fromJson(Class<T>)` to bind only instance fields.

Changes in this PR:
- skip static fields
- add a regression test
- update the Javadoc

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -3349,7 +3349,7 @@ public class JSONObject {
      * of the given class. It supports basic data types including {@code int}, {@code double},
      * {@code float}, {@code long}, and {@code boolean}, as well as their boxed counterparts.
      * The target class must have a no-argument constructor, and its field names must match
-     * the keys in the JSON string.
+     * the keys in the JSON string. Static fields are ignored.
      *
      * <p><strong>Note:</strong> Only classes that are explicitly supported and registered within
      * the {@code JSONObject} context can be deserialized. If the provided class is not among those,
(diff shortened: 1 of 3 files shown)


stleary:
fromJson() should not populate static class fields

Yes

**Risks**
Low

**Changes to the API?**
No

No

No, JavaDoc was updated

No, and a new unit test was added

No

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

PR #1046, opened 2026-03-23
Validate XML numeric character references before string construction

### LINKED ISSUE #1045: XML.toJSONObject() throws uncaught IllegalArgumentException for out-of-range numeric character references
In line with the repository's published vulnerability-reporting guidance, I am reporting this issue here as a public issue.

### Summary

When parsing crafted XML containing an out-of-range numeric character reference such as `&#x110000;`, `XML#toJSONObject()` throws an uncaught `IllegalArgumentException` instead of a controlled parsing exception such as `JSONException`.

As a result, applications that parse attacker-controlled XML may encounter an uncaught runtime exception. Depending on the integration, this may result in request failure or denial of service.

I reproduced this in release `20251224`.

### Details

The apparent root cause is in `XMLTokener#unescapeEntity()`, where a decoded numeric character reference is passed to string construction without first validating that it is a valid Unicode code point:

https://github.com/stleary/JSON-java/blob/cf653682be81e5ff630c8ac5527a5060b9a813e8/src/main/java/org/json/XMLTokener.java#L162-L171

#### Minimal PoC
```java
XML.toJSONObject("<a>&#x110000;</a>");
```

I also checked a few closely related inputs while narrowing this down:
- `&#1114112;` reproduces the same behavior.
- The same behavior is also reachable from an attribute value, e.g. `<a b="&#x110000;"/>`.
- `&#xD800;` did not reproduce the same uncaught exception in my testing.

This suggests that the immediate issue here is specifically the handling of out-of-range Unicode code points during string construction, rather than XML-invalid numeric character references in general.

#### Observed Result
```text
java.lang.IllegalArgumentException: 1114112
        at java.base/java.lang.StringUTF16.toBytes(Unknown Source)
        at java.base/java.lang.String.<init>(Unknown Source)
        at org.json.XMLTokener.unescapeEntity(XMLTokener.java:171)
        at org.json.XMLTokener.nextEntity(XMLTokener.java:148)
        at org.json.XMLTokener.nextContent(XMLTokener.java:117)
        at org.json.XML.parse(XML.java:407)
        at org.json.XML.toJSONObject(XML.java:780)
        at org.json.XML.toJSONObject(XML.java:866)
        at org.json.XML.toJSONObject(XML.java:665)
        at PoC.main(PoC.java:7)
```

yuki-matsuhashi:
Fixes #1045

This PR validates XML numeric character references before string construction.

Changes in this PR:
- reject XML-invalid numeric character references with `JSONException`
- add regression tests for out-of-range hex and decimal values
- add regression coverage for surrogate values and attribute values

#### Note

#1045 focused on out-of-range values such as `&#x110000;`, which raised an uncaught runtime exception before this fix. Surrogate values such as `&#xD800;` did not reproduce the same exception in my testing, but they are still invalid XML character references and are rejected by the same validation.

### DIFF
--- a/src/main/java/org/json/XML.java
+++ b/src/main/java/org/json/XML.java
@@ -158,7 +158,7 @@ public class XML {
      * @param cp code point to test
      * @return true if the code point is not valid for an XML
      */
-    private static boolean mustEscape(int cp) {
+    static boolean mustEscape(int cp) {
         /* Valid range from https://www.w3.org/TR/REC-xml/#charsets
          *
          * #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]
--- a/src/main/java/org/json/XMLTokener.java
+++ b/src/main/java/org/json/XMLTokener.java
@@ -167,6 +167,9 @@ public class XMLTokener extends JSONTokener {
             int cp = (e.charAt(1) == 'x' || e.charAt(1) == 'X')
                 ? parseHexEntity(e)
                 : parseDecimalEntity(e);
+            if (XML.mustEscape(cp)) {
+                throw new JSONException("Invalid numeric character reference: &#" + e.substring(1) + ";");
+            }
             return new String(new int[] {cp}, 0, 1);
         }
         Character knownEntity = entity.get(e);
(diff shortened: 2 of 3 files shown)


stleary:
XML code throws an uncontrolled exception when parsing invalid codepoints to JSON.  This fix replaced the exception with a JSONException

Yes

**Risks**
Low

**Changes to the API?**
Yes, but just in the type of exception thrown

No

No

No

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

Issue #1049, opened 2026-04-04
Inconsistent null Handling Across JSONObject Construction Paths


venkat-cheenu:
**Jar version :** json-20251224.jar

**Reproduction :**
```
// String constructor — null preserved
new JSONObject("{\"name\":null}");         // {"name":null}

// Map constructor — null dropped
Map<Object, Object> map = new HashMap<>();
map.put("name", null);
new JSONObject(map);                       // {}

// put() — null dropped
JSONObject json = new JSONObject();
json.put("name", (String) null);           // {}
```

**Expected :**
All construction paths should treat null consistently. Per the JSON spec (RFC 8259), null is a valid value and should be retained.

stleary:
@venkat-cheenu Thanks for bringing this up. This is a known problem for map parsing. For example, see #296.
For now, you can use these workarounds:
```
    @Test
    public void foo() {
        JSONObject jsonObject = new JSONObject();
        Map<String, Object> map = new HashMap<>();
        map.put("mapKeyForNull", null);
        jsonObject = new JSONObject(map, new JSONParserConfiguration().withUseNativeNulls(true));
        jsonObject.put("putKeyForNull", JSONObject.NULL);
        System.out.println(jsonObject.toString(2));
    }
```
Output:
```
{
  "putKeyForNull": null,
  "mapKeyForNull": null
}
```

However, it might be reasonable to update the code to accept nulls in maps by default when running in strict mode. Would this work for you?

venkat-cheenu:
Thanks for the detailed comment. @stleary 

The codebase I'm working on uses this library extensively—across about 22,000 files. Is there a way to enable a system-wide configuration to allow NativeNull and overrideDuplicateKey?

stleary:
@venkat-cheenu Do you have any suggestions on how to do this?

venkat-cheenu:
Hi @stleary,

To solve this, we could introduce a global, system-wide configuration state while keeping the individual JSONParserConfiguration objects immutable.

We could add a thread-safe static reference (like an AtomicReference) to hold a default global configuration. It would look something like this for the user:

Java
```
// 1. Define your custom configuration
JSONParserConfiguration myConfig = new JSONParserConfiguration()
    .withUseNativeNulls(true)
    .withOverwriteDuplicateKey(true);

// 2. Set it globally (do this once during application startup)
JSONParserConfiguration.setGlobalConfiguration(myConfig);
```

[...]

PR #1050, opened 2026-04-05
[FEATURE] Global JSONParserConfiguration support

venkat-cheenu:


stleary:
The Java6 compiler error should be easy to fix, I think you just need to use this statement: 
```
  private static final AtomicReference<JSONParserConfiguration> GLOBAL_CONFIG  = 
        new AtomicReference<JSONParserConfiguration>(null);
```

AtomicReference may not be a good idea. This lib is not thread-safe, per design. Adding it here may give the impression the project is moving in that direction, which it is not.

There are no unit tests for the new/changed code.

If some maps need to be parsed with keep-null, and some maps without it, how would you get that behavior?

I'm not convinced yet that this is the best approach. It may be that this problem should be fixed in user code.

venkat-cheenu:
> AtomicReference may not be a good idea. This lib is not thread-safe, per design. Adding it here may give the impression the project is moving in that direction, which it is not.

I have removed AtomicReference and am now directly using the class.

> There are no unit tests for the new/changed code.

I have added unit tests to cover the new and modified code.

> If some maps need to be parsed with keep-null, and some maps without it, how would you get that behavior?


This Global JSONParserConfiguration support enables user to have SystemWide setting for parsing Json. User can override it with new JSONParserConfiguration support when they dealing with to be parsed differently

stleary:
@venkat-cheenu, Sorry for taking so long to respond, and thanks for your efforts on this PR. It is a reasonable implementation for a global config object, but I've been advised that a simpler and more targeted result can be achieved by updating `JSONObject.put()` to allow an optional `JSONParserConfiguration` param. #1049 will be kept open until a fix is ready.

Issue #1063, opened 2026-06-18
[security disclosure] Unbounded numeric literal parsing causes resource exhaustion (DoS) — missing number length limit


waydeshi:
Hi, I'd like to report a security issue in org.json that causes availability impact in applications that parse user-supplied JSON or XML.

## Issue

When parsing a JSON or XML document containing a very large numeric literal (e.g. a number with tens of thousands of digits), the parser constructs a BigInteger or BigDecimal object from the raw string with no length limit. Because BigInteger construction has super-linear time complexity, a moderately sized request body can cause a server thread to block for several seconds.

This issue is present in JSONTokener.java and JSONObject.java. It affects all three primary parsing entry points: new JSONObject(String), new JSONArray(String), and XML.toJSONObject(String). No existing JSONParserConfiguration option (including withStrictMode(true)) prevents the behavior.                                           

## Affected versions: 20150729 through 20260522 (all releases since BigInteger/BigDecimal support was added).

CVSS 3.1: 5.9 Medium — AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:H
CWE: CWE-400, CWE-770

## Note on prior CVEs: 
This is distinct from CVE-2022-45688 (XML nesting depth → StackOverflow) and CVE-2023-5072 (JSON structural nesting → StackOverflow). Both of those issues involve recursive call-stack depth and were fixed by catching StackOverflowError or adding a maxNestingDepth guard. The numeric string length issue is in a different code path and has not been addressed by either fix.

## Suggested fix

Add a configurable maximum numeric literal length to JSONParserConfiguration, similar to Jackson's StreamReadConstraints.withMaxNumberLength() introduced in Jackson 2.15 (April 2023) with a default of 1,000 characters. A guard inside the accumulation loop in nextSimpleValue() would prevent construction of BigInteger/BigDecimal from excessively long strings.

Full security advisory available upon request via private channel.

stleary:
@waydeshi Thanks for submitting this issue, looks like a valid concern. If you have code to recreate the bug, please post it here.

waydeshi:
@stleary below is the PoC code.

### POC Code

```java
import org.json.*;

public class Poc2NumericDoSv2 {
    public static void main(String[] args) {
        System.out.println("=== CVE PoC: JSONTokener Unbounded Numeric String DoS ===");

        // Scaling: timing at different digit counts
        System.out.println("\n-- Unquoted numeric literal (JSONObject) --");
[...]

PR #1065, opened 2026-07-03
Fixes CVE-2026-59171

Damien-Warren-206:
Description: 
See #1063
Fixed JSONObject and XML to check length before calling stringToNumber()

Note: 

JSONArray and JSONTokener did not need to be fixed because all code goes through JSONObject. Added multiple unit tests to test the new functionality.

### DIFF
--- a/src/main/java/org/json/JSONObject.java
+++ b/src/main/java/org/json/JSONObject.java
@@ -2700,7 +2700,9 @@ public class JSONObject {
         char initial = string.charAt(0);
         if ((initial >= '0' && initial <= '9') || initial == '-') {
             try {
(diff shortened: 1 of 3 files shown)


stleary:
Don't try to parse JSON strings consisting of numbers with more than 1000 chars. This is a new hard-coded limit. By default, these values will be parsed as strings. A future fix may make the limit configurable.

Yes

**Risks**
Low.
This is a behavioral change in response to a CVE

**Changes to the API?**
No

Yes

No

No

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

PR #1068, opened 2026-07-16
Max number length config for BigInteger and BigDecimal

### LINKED ISSUE #1066: ParserConfiguration should support max number length

stleary:
**Description**
Additional work for CVE-2026-59171 (See #1065):
* Max number length should be configurable in ParserConfiguration
* Max number length for BigInteger and BigDecimal should be enforced for getBigInteger() and optBigInteger().
This fixes #1066 

**Refactoring**
N/A

**Testing done**
New unit tests were added.

### DIFF
--- a/src/main/java/org/json/JSONArray.java
+++ b/src/main/java/org/json/JSONArray.java
@@ -483,8 +483,29 @@ public class JSONArray implements Iterable<Object> {
      *             to a BigInteger.
      */
     public BigInteger getBigInteger (int index) throws JSONException {
+        return this.getBigInteger(index, new JSONParserConfiguration());
+    }
+
+    /**
+     * Get the BigInteger value associated with an index.
+     *
+     * @param index
+     *            The index must be between 0 and length() - 1.
+     * @param jsonParserConfiguration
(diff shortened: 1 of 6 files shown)


stleary:
Completes the fix for CVE-2026-59171 Unbounded numeric string parsing

Yes

**Risks**
Low. The behavior for very large BigInteger and BigDecimal has changed, but it is unlikely that anyone will encounter the change with valid JSON. If they do, the capability to disable this fix is included. 

**Changes to the API?**
The lib will not attempt to parse strings containing very large numbers. When this happens, a string will be emitted instead of a BigInteger or BigDecimal.

Yes

Yes

No, unit tests were added

Minimal refactoring due to SonarQube results

**Review status**
**APPROVED** - by myself

Thanks to @waydeshi, @Damien-Warren-206, and @mechko for contributions that resulted in this fix


Starting 3 day comment window

Issue #308, opened 2017-01-10
XML.toString() returns invalid xml


MrFaridMam:
I created JsonObject from String, then used XML.toString(JsonObject o) to create XML String to send response in my Servlet. But browser says it's invalid.

johnjaylward:
Can you give an example to reproduce this?

MrFaridMam:
try this json:

```
{  
   "someObject":{  

   },
   "someObject2":{  

   },
   "someArray":{  
      "1234":[  
         {  
            "object":"data"
         }
      ]
   }
}
```

there is 2 error: 1st - there is no global <root> tag after conversion. 2nd - <1234> tag is incorrect, so I added to XML.java this code(416th line):

 ```
if (key.matches("\\d+")) {
                    key = "id_" + key;
                }
```

stleary:
Given the JSON text input, the transformed XML looks reasonable to me. Please post what you think it should have been.

<someObject2></someObject2>
<someObject></someObject>
<someArray>
 <1234>
  <object>data</object>
 </1234>
</someArray>

johnjaylward:
this is a known issue. Our policy to date has been that the XML output does not guarantee valid XML. This is especially true with numeric element IDs. <1234> is not valid XML. Our suggestion to other posters has been to pre-process the JSONObject before calling `XML.toString(JSONObject o)`

See #294 for the most recent discussion on it.

stleary:
Post if you want to reopen this issue for further discussion.

Issue #294, opened 2016-10-14
XML.toString does not handle the case where JSON keys are not valid XML element names.


imhogan:
JSON key names are not as tightly constrained as XML element names - eg:
 {"6":"x"," this has spaces and quotes\"!":"Ouch"}
 is a valid JSON object, but XML.toString would convert this to 
<root><6>x</6>< this has spaces and quotes"!>Ouch"</ this has spaces and quotes"!></root>
Which is not valid XML. One solution is to prefix element names which start with a digit with _ and replace other invalid characters with an _ too. Whilst this makes the resultant XML valid, meaning can be lost, so to retain the meaning an attribute could be added to provide the original JSON key name. This approach would result in the XML:
<root><_6 JSON_Key="6">x</6><_this_has_spaces_and_quotes__ JSONKey=" this has spaces and quotes&quot;!">Ouch"</_this_has_spaces_and_quotes__></root>

See https://www.w3.org/TR/REC-xml/#NT-Name for constraints on XML element names. Note, ":" should be replaced with another character as otherwise it could result in an element name which appears to contain a namespace prefix.

johnjaylward:
This has been brought up before in #166, #138, and #123. I don't see this getting corrected anytime soon.

It is currently expected that the developer calling this library pre-process the JSON to make any correction to key names so that it fits their application needs.

I doubt any change here would break an existing application since invalid node name could not be processed. I personally believe the best solution would be to throw an exception during XML generation if the node name would be invalid.

stleary:
Changing users' content is not an option. In #123, Douglas offered throwing an exception on invalid input. No objection to someone doing this now.

Issue #166, opened 2015-10-16
XML.toString() need to handle invalid characters in nodeName


anilverma070:
When we are using, it happened that keyNames in out JSON contained $ character. When this JSON was converted to XML, it puts those characters unmodified as xml node name, which makes xml invalid. Please remove invalid characters from xml node names when converting from JSON to xml.

stleary:
Unless the JSON started out as valid XML and then became invalid after processing by JSON-java, this sounds like a problem with your input data. Can you provide a snippet of the data? Please also include the API methods used to pass the data to JSON-java, and how it is retrieved.

johnjaylward:
also see #138 and #123 for similar issues as well.

stleary:
As previously noted, our choices are to throw an exception or leaving the code in place. No changes are planned for this issue at the present time.

PR #1072, opened 2026-07-23
Reject invalid XML element names

### LINKED ISSUE #1071: XML.toString: unescaped keys allow XML element injection (CWE-91) — proposal to throw per #294
### Summary

`XML.toString(JSONObject)` emits JSONObject keys verbatim as XML element
names. A key containing `<`, `>` or `/` breaks out of its own tag and
injects arbitrary sibling elements into the output. When the JSONObject
is built from untrusted JSON (`new JSONObject(String)` — the documented
JSON→XML flow), that is attacker-controlled structure in the emitted XML.

### Prior art

This has come up before as a *well-formedness* question — #123, #138,
#166, #294, #308 — with the standing guidance being "caller must
pre-process keys." I'm re-raising because:

1. **Impact is injection, not just malformed output** (PoC below).
2. In #294 you [wrote](https://github.com/stleary/JSON-java/issues/294#issuecomment-254003963)
   *"No objection to someone doing this now"* re: throwing on invalid
   input — this issue is that "someone."

### Reproduction (master @ 1795e8c / release 20260719)

```java
import org.json.*;
public class Repro {
    public static void main(String[] a) {
        JSONObject jo = new JSONObject(
            "{\"a/><injected>evil</injected><a\":\"\"}");
        String xml = XML.toString(jo, "root");
        System.out.println("XML output: " + xml);
        System.out.println("Re-parsed:  " + XML.toJSONObject(xml).toString(2));
    }
}
```
```
XML output: <root><a/><injected>evil</injected><a/></root>
Re-parsed:  {"root": {
  "a": [
    "",
    ""
  ],
  "injected": "evil"
}}
```

A single JSON key produces an attacker-chosen `<injected>` element that
survives a round-trip through `XML.toJSONObject`. Downstream impact
depends on the XML consumer (config/SOAP smuggling, stored XSS if later
rendered, XXE pivot if the consumer resolves an injected `<!DOCTYPE>`).

### Affected code

`XML.toString(Object, String, XMLParserConfiguration, int, int)` in
`src/main/java/org/json/XML.java`: `tagName` at lines 977 / 1066 /
1100 / 1102–1103 and `key` at lines 1023 / 1027 / 1037 / 1040 / 1048
are appended between `<` / `>` without validation. Values already go
through `escape()`; keys do not.

### Proposed fix — throw on invalid XML Name (per #294 / #123)

Reject any key or `tagName` that is not a valid XML `Name`:

```java
private static void mustBeXmlName(String name) throws JSONException {
    if (name == null || name.isEmpty()
            || !isXmlNameStart(name.charAt(0))) {
        throw new JSONException("'" + name + "' is not a valid XML element name");
    }
    for (int i = 1; i < name.length(); i++) {
        if (!isXmlNameChar(name.charAt(i))) {
            throw new JSONException("'" + name + "' is not a valid XML element name");
        }
    }
}
```

called once for `tagName` at method entry and once per `key` at the top
of the key loop. `<`, `>`, `/`, space, `"`, `&` are all outside the XML
`Name` production, so the injection vector is closed. Keys that are
already valid element names are unaffected.

This is the option you indicated was acceptable in #294 (*"changing
users' content is not an option … no objection to [throwing]"*). If
you'd instead prefer the `JSONML` approach (`XML.escape(tagName)`,
`JSONML.java:557–558`), happy to do that — but going with your stated
preference by default.

I have a patch + unit test ready and can open a PR against `master`.

CWE-91. Affects all releases ≤ 20260719.

mechko:
Fixes #1071. Implements the throw-on-invalid-name approach discussed in #294 / #123, which also resolves the long-standing well-formedness question in #166 / #308.

### Problem

`XML.toString` emits JSONObject keys verbatim as XML tag names. A key containing `<`, `>` or `/` breaks out of its element and injects arbitrary sibling structure:

```
{"a/><injected>evil</injected><a": ""}
    → <root><a/><injected>evil</injected><a/></root>
```

Values already go through `escape()`; keys/`tagName` do not.

### Fix

Validate every key and `tagName` against the XML 1.0 (5th ed.) `Name` production before emitting; throw `JSONException` on mismatch.

- `mustBeXmlName(String)` — throws if the string is null/empty or contains a code point outside the `Name` production
- `isXmlNameStart(int)` / `isXmlNameChar(int)` — the two productions, code-point aware (handles supplementary planes)
- Called once for `tagName` at the top of the private `toString` overload, and once per `key` at the top of the key loop (the `cDataTagName` sentinel is skipped since it never becomes a tag)

`<`, `>`, `/`, whitespace, `"`, `&`, `@` are all outside `NameChar`, so the injection vector is closed. Keys that are already valid XML Names are unaffected — `mustBeXmlName` is a no-op for them.

### Behaviour change

`XML.toString` now throws for keys that are not valid XML Names, where previously it emitted malformed XML:

| Key | Before | After |
|---|---|---|
| `"a/><injected>…"` | element injection | `JSONException` |
| `"123foo"` | `<123foo>` (invalid XML) | `JSONException` |
| `"has space"` | `<has space>` (invalid XML) | `JSONException` |
| `"foo@bar"` | `<foo@bar>` (invalid XML) | `JSONException` |
| `"ns:name"`, `"a-b.c_d"`, `"élément"` | unchanged | unchanged |

The two existing `shouldHandleIllegalJSONNodeNames` tests already documented the old output as "invalid XML" / "possible bug"; they've been rewritten to assert the throw.

### Verification

```
$ java -cp target/classes:. XmlRepro
Exception in thread "main" org.json.JSONException: 'a/><injected>evil</injected><a' is not a valid XML element name.
        at org.json.XML.mustBeXmlName(XML.java:259)
        at org.json.XML.toString(XML.java:1063)
```

`mvn test`: 791 run, 0 fail, 0 error, 6 skipped (pre-existing).

### Tests

- `XMLTest.shouldHandleIllegalJSONNodeNames` / `XMLConfigurationTest.shouldHandleIllegalJSONNodeNames` — rewritten to assert `JSONException` for `123IllegalNode` and `Illegal@node`
- `XMLTest.toStringRejectsElementInjectionInKey` — the #1071 payload throws; invalid caller-supplied `tagName` throws
- `XMLTest.toStringAcceptsValidXmlNames` — hyphen/dot/underscore/colon, Latin-1 letters, and `cDataTagName` sentinel still serialise

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/java/org/json/XML.java
+++ b/src/main/java/org/json/XML.java
@@ -262,6 +262,10 @@ public class XML {
         }
     }
 
+    private static boolean inRange(int cp, int lo, int hi) {
+        return cp >= lo && cp <= hi;
+    }
+
     /**
      * XML 1.0 (5th ed.) {@code NameStartChar} production.
      *
@@ -269,21 +273,22 @@ public class XML {
      * @return true if {@code cp} may start an XML Name
      */
     private static boolean isXmlNameStart(int cp) {
-        return cp == ':' || cp == '_'
-                || (cp >= 'A' && cp <= 'Z')
-                || (cp >= 'a' && cp <= 'z')
-                || (cp >= 0xC0   && cp <= 0xD6)
-                || (cp >= 0xD8   && cp <= 0xF6)
-                || (cp >= 0xF8   && cp <= 0x2FF)
-                || (cp >= 0x370  && cp <= 0x37D)
-                || (cp >= 0x37F  && cp <= 0x1FFF)
-                || (cp >= 0x200C && cp <= 0x200D)
-                || (cp >= 0x2070 && cp <= 0x218F)
-                || (cp >= 0x2C00 && cp <= 0x2FEF)
-                || (cp >= 0x3001 && cp <= 0xD7FF)
-                || (cp >= 0xF900 && cp <= 0xFDCF)
-                || (cp >= 0xFDF0 && cp <= 0xFFFD)
-                || (cp >= 0x10000 && cp <= 0xEFFFF);
+        return cp == ':'
+                || cp == '_'
+                || inRange(cp, 'A', 'Z')
+                || inRange(cp, 'a', 'z')
+                || inRange(cp, 0xC0,    0xD6)
+                || inRange(cp, 0xD8,    0xF6)
+                || inRange(cp, 0xF8,    0x2FF)
+                || inRange(cp, 0x370,   0x37D)
+                || inRange(cp, 0x37F,   0x1FFF)
+                || inRange(cp, 0x200C,  0x200D)
+                || inRange(cp, 0x2070,  0x218F)
+                || inRange(cp, 0x2C00,  0x2FEF)
+                || inRange(cp, 0x3001,  0xD7FF)
+                || inRange(cp, 0xF900,  0xFDCF)
+                || inRange(cp, 0xFDF0,  0xFFFD)
+                || inRange(cp, 0x10000, 0xEFFFF);
     }
 
     /**
@@ -294,11 +299,12 @@ public class XML {
      */
     private static boolean isXmlNameChar(int cp) {
         return isXmlNameStart(cp)
-                || cp == '-' || cp == '.'
-                || (cp >= '0' && cp <= '9')
+                || cp == '-'
+                || cp == '.'
                 || cp == 0xB7
-                || (cp >= 0x0300 && cp <= 0x036F)
-                || (cp >= 0x203F && cp <= 0x2040);
+                || inRange(cp, '0', '9')
+                || inRange(cp, 0x0300, 0x036F)
+                || inRange(cp, 0x203F, 0x2040);
     }
 
     /**
--- a/src/main/java/org/json/XML.java
+++ b/src/main/java/org/json/XML.java
@@ -233,6 +233,75 @@ public class XML {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


stleary:
@mechko Sorry for the long wait, will get to this before next week. While reviewing the PR, I found that strict mode is not filtering incoming text for valid JSON chars. I will fix this for strict mode, but keep the current behavior for non-strict mode.

stleary:
@mechko Thanks for the PR. Can you please restrict this fix to the security issues mentioned in #1071. Enforcing well-formed tags could break backwards compatibility. Fixing the security issues also does this, but is justified because the fix prevents an injection attack.
* Only the chars <  >  &  "  '  /  in tag names need to be caught
* Please remove the description text referring to well-formedness and #166 / #308

mechko:
@stleary I adjusted the fix according to your comment, please let me know if you still want anything to be changed.

stleary:
Don't allow invalid chars when converting JSON to XML.

**Risks**
Low. 

This is a behavior change, but it is expected that it rarely occurs. 
If there are issues, it can be triggered by an opt-in parser config flag instead.

**Changes to the API**
No

No

Yes

**Unit Tests**
New unit tests were added

**Refactoring**
None

**Review status**
**APPROVED** 

Starting 3-day comment window

stleary (approved):

PR #1077, opened 2026-09-18
JSONObject/JSONArray ctors override JSONParserConfiguration for strict mode

### LINKED ISSUE #945: Which JSONParserConfiguration to use in JSONObject/JSONArray constructors
From #927:

What is the purpose of JSONParserConfiguration parameter in JSONArray/JSONObject constructors? Is it not enough to use JSONTokener's JSONParserConfiguration? For example, when calling "JSONArray(JSONTokener x, JSONParserConfiguration jsonParserConfiguration)" constructor the array syntax is checked according to jsonParserConfiguration parameter, but nested objects/arrays syntax is checked according to JSONTokener's configuration. It seems a bit inconsistent.

Both the JSONArray/JSONObject instance and the JSONTokener parameter already have a JSONParserConfiguration. Only one of these should be used. I am leaning towards replacing the JSONTokener config instance with the config param. This should also be called out in the JavaDocs for the corresponding constructors.

Mahmoodifar:
When a JSONObject or JSONArray constructor receives both a JSONTokener and an explicit parser configuration, nested values still use the tokener's previous configuration. For example, strict parsing can accept single-quoted values.

Apply the supplied configuration throughout construction, including nested values, using a package-private setter. Restore the tokener's original configuration in a finally block on both success and failure, and document this behavior.

Fixes #945.

Regression coverage includes strict and lenient overrides, nested duplicate keys, sibling values, empty containers, malformed input, trailing content, and reuse of the tokener after construction.

Validation on Java 17:
- `mvn -B -ntp clean test`
- `mvn -B -ntp test -Ptest-strict-mode javadoc:jar@attach-javadocs`

Both suites: 804 tests, 6 skipped, no failures or errors. Javadocs built successfully. Five configuration-restoration tests failed before the follow-up fix.


src/main/java/org/json/JSONArray.java
```diff
@@ -108,6 +110,7 @@ public JSONArray(JSONTokener x, JSONParserConfiguration jsonParserConfiguration)
      */
     JSONArray(JSONTokener x, JSONParserConfiguration jsonParserConfiguration, boolean isInitial) throws JSONException {
         this();
+        x.setJsonParserConfiguration(jsonParserConfiguration);
```
stleary: Please restore the original JSONParserConfiguration to JSONTokener before returning.

src/main/java/org/json/JSONObject.java
```diff
@@ -233,6 +235,7 @@ public JSONObject(JSONTokener x, JSONParserConfiguration jsonParserConfiguration
      */
     JSONObject(JSONTokener x, JSONParserConfiguration jsonParserConfiguration, boolean isInitial) throws JSONException {
         this();
+        x.setJsonParserConfiguration(jsonParserConfiguration);
```
stleary: Please restore the original JSONParserConfiguration to JSONTokener before returning.

src/main/java/org/json/JSONArray.java
stleary: JSONTokener.setParserConfiguration() is deprecated. Please add and use a new method that does the same thing, but make it package-private.

stleary:
@Mahmoodifar Thanks for the PR. Looks good, a couple of minor items to address.

stleary:
Fixes issue where JSONObject/JSONArray ctor and JSONTokener have different parser config values for strict mode. Now the ctor takes precedence. The original JSONTokener strict mode setting is restored before returning.

**Risks**
Low

Yes. This will change the existing behavior when parsing with a JSONTokener which does not have the same strict mode setting as the JSONObject or JSONArray. This is an acceptable corner-case behavior change.

**Changes to the API**
No

No

No, but a future readme update should document strict mode.

**Unit Tests**
New unit tests were added

**Refactoring**
No

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

PR #1078, opened 2026-09-19
Decode JSONPointer tokens once only

jakezwang:
`JSONPointer` decodes string tokens in its constructor, then decodes them again during object lookup. For example, `/~01` incorrectly selects the `/` member instead of the literal `~1` member. Builder and token-list pointers also misinterpret literal tokens containing `~0` or `~1`.

Use the stored token directly in `queryFrom`. Add regressions for string and URI-fragment pointers, builders, and token lists with distinct colliding keys. This follows the decoding order in [RFC 6901 section 4](https://www.rfc-editor.org/rfc/rfc6901#section-4) and addresses the double-unescaping behavior mentioned in #561.

Validation: the three new tests fail before the fix. The JDK 8 and 21 suites and the JDK 21 strict-mode profile each report 801 tests, zero failures/errors and six existing skips. Package and Javadoc generation pass, as does Java 6 source/target compilation.


stleary:
@jakezwang Thanks for the PR, good catch and a good fix. Apparently, fixing this issue exposes a pre-existing bug in `toURIFragment()`, for which I will add a separate issue.

stleary:
Fixes an issue in JSONPointer where double-unescaping is performed incorrectly. This bug has been in the code since the initial commit. 

**Risks**
Moderately low, for users who have been working around this issue.

This fix includes a behavior change: users who have been aware of the problem and worked around it may find that their code no longer works. This is considered an acceptable change. Implementing the fix also exposes another pre-existing bug, which will be fixed elsewhere.
 
**Changes to the API**
No

No

No

**Unit Tests**
New unit tests were added

**Refactoring**
None

**Review status**
**APPROVED**

Starting 3-day comment window

stleary (approved):

